PR #52, opened 2024-02-21
fix: update the address book and perform helm upgrade to fix mirror node importer

jeromy-cannon:
## Description

* update the address book secret for mirror node importer
* enhance `solo node start` to trigger a `helm upgrade` supplying the address book as a value
* some changes from `full-stack-testing` repo that got missed during the migration
* update workflow to upload solo.log to GitHub

### Related Issues

* Closes #https://github.com/hashgraph/full-stack-testing/issues/761
* depends on https://github.com/hashgraph/full-stack-testing/pull/777
* migrated from https://github.com/hashgraph/full-stack-testing/pull/776


jest.config.mjs
```diff
@@ -15,9 +15,21 @@
  *
  */
 const config = {
-  testRegex: '(/__tests__/.*|(\\.|/)(test|spec))\\.(mjs?)$',
   moduleFileExtensions: ['js', 'mjs'],
   verbose: true,
+  testSequencer: './test/e2e/jestCustomSequencer.cjs',
```
leninmehedy, line 20: We need to see if we can avoid the need to run tests in sequence. Each test should "in theory" independent of each other as much as possible (each test should set its own environment and tear down afterwards).

Please create a ticket so that we can remove dependencies between tests.

src/core/account_manager.mjs
```diff
@@ -271,7 +296,7 @@ export class AccountManager {
     let keys
     try {
       keys = await this.getAccountKeys(accountId, nodeClient)
-      this.logger.debug(`retrieved keys for account ${accountId.toString()}`)
+      this.logger.silly(`retrieved keys for account ${accountId.toString()}`)
```
leninmehedy, line 299: Not sure if need this unconventional `silly` log type?

src/core/account_manager.mjs
```diff
@@ -515,4 +540,40 @@ export class AccountManager {
...
+        modified = true
+      }
+    }
+
+    if (modified) {
+      addressBookBytes = HashgraphProto.proto.NodeAddressBook.encode(addressBook).finish()
+    }
+
+    // convert addressBook into base64
+    return Base64.encode(addressBookBytes)
```
leninmehedy, line 577: I remember writing this function, but I did not test whether `Base64.encode` works when `addressBookBytes` is not modified. 

`addressBookBytes` is Uint8 array when fetched from the network, but then `HashgraphProto.proto.NodeAddressBook.encode` generate a buffer.

leninmehedy (approved):

src/commands/node.mjs
```diff
@@ -483,6 +496,36 @@ export class NodeCommand extends BaseCommand {
           })
         }
       },
+      {
+        title: 'Upgrade the network deployment to enable mirror node with the address book',
```
leninmehedy, line 500: 'Upgrade the network deployment to enable mirror node with the address book' -> `Enable mirror node`

jest.config.mjs
jeromy-cannon: true, it was bugging me, too.

Created: https://github.com/hashgraph/solo/issues/62

src/core/account_manager.mjs
jeromy-cannon: yes, I didn' t like the crazy amount of logs these were creating.  I wanted to make them trace, but the npm that winston logger was defaulting to only had silly.

I'll just go ahead and remove them.  They were useful when I was getting it all working, but now that it is pretty solid it can be removed.

src/core/account_manager.mjs
jeromy-cannon: sure, I'll add a test

src/core/account_manager.mjs
jeromy-cannon: removed

src/commands/node.mjs
jeromy-cannon: updated

src/core/account_manager.mjs
jeromy-cannon: I want to do a mock test, but I want this to merge before I run my release.  I did this test to verify your code:
image

Here is a new Issue to track writing the test: https://github.com/hashgraph/solo/issues/63

matteriben (approved):

nathanklick (approved):

PR #84, opened 2024-02-27
feat(cli): allow user to set jvm options for Hedera nodes

### LINKED ISSUE #68: Solo should allow overriding the JVM settings

leninmehedy:
## Description

* fix: fetch platform code directly inside the container
  * This is because fetching the platform zip file locally and then uploading to the container is time consuming and brittle.
* feat: allow user to set jvm options for Hedera nodes

When node starts, JVM options are loaded from `/opt/hgcapp/services-Hedera/HapiApp2.0/application.env`

Now user is able to set JVM options with flag `--application-env` flag.  
`node start --application-env application.env`

### Related Issues

* Closes #68

### DIFF
--- a/src/commands/node.mjs
+++ b/src/commands/node.mjs
@@ -228,15 +228,6 @@ export class NodeCommand extends BaseCommand {
         task: (ctx, task) => self.taskCheckNetworkNodePods(ctx, task)
       },
       {
-        title: 'Fetch platform software',
-        task: async (ctx, _) => {
-          const config = ctx.config
-          if (config.force || !fs.existsSync(config.buildZipFile)) {
-            ctx.config.buildZipFile = await self.downloader.fetchPlatform(ctx.config.releaseTag, config.cacheDir)
-          }
-        }
-      },
-      {
         title: 'Generate Gossip keys',
         task: async (ctx, _) => {
           const config = ctx.config
@@ -349,7 +340,7 @@ export class NodeCommand extends BaseCommand {
         }
       },
       {
-        title: 'Upload platform software into network nodes',
+        title: 'Fetch platform software into network nodes',
         task:
           async (ctx, task) => {
             const config = ctx.config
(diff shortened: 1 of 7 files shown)


jeromy-cannon:
@leninmehedy , just a thought, but if we created a persistent volume mount, we could just download one time and reduce the traffic

jeromy-cannon (approved):

leninmehedy:
> @leninmehedy , just a thought, but if we created a persistent volume mount, we could just download one time and reduce the traffic

Yes, hold on that thought. Please create a ticket to create persistent-volume and add your notes. PVC should be a high priority (maybe P1)

src/commands/node.mjs
```diff
@@ -460,6 +455,15 @@ export class NodeCommand extends BaseCommand {
               title: `Start node: ${chalk.yellow(nodeId)}`,
               task: async () => {
                 await self.k8.execContainer(podName, constants.ROOT_CONTAINER, ['rm', '-rf', `${constants.HEDERA_HAPI_PATH}/data/logs`])
+
+                // copy application.env file if required
+                if (ctx.config.applicationEnv) {
+                  const stagingDir = Templates.renderStagingDir(self.configManager, flags)
+                  const applicationEnvFile = path.join(stagingDir, 'application.env')
+                  fs.cpSync(ctx.config.applicationEnv, applicationEnvFile)
```
jeromy-cannon, line 463: Talking with Nathan it sounds like all of these config files need to be a ConfigMap that is mounted in.

nathanklick (approved):

src/commands/node.mjs
leninmehedy: As discussed, we shall need to work on a separate PR where `application.env` will be mounted from a config-map to `/etc/network-node/` directory. 

This will also require updating the docker images in `full-stack-testing/docker` to ensure `application.env` is loaded from `/etc/network-node/`:
`docker/ubi8-init-java21/network-node.service`
`docker/ubi8-init-java21/Dockerfile`

leninmehedy: See https://github.com/hashgraph/solo/issues/110

leninmehedy:
Merging this as this is a working copy that enabled us to provision a 10K TPS network. Further enhancements will be done as separate PRs.

PR #139, opened 2024-03-06
fix: solo should use a fixed FST chart version

### LINKED ISSUE #109: Solo should use a default FST chart version
Create a `version.mjs` with the required versions such that we can update it in CI during release instead of modifying the code.

leninmehedy:
## Description

* fix: solo should use a fixed FST chart version

### Related Issues

* Closes #109


nathanklick (approved):

version.mjs
```diff
@@ -14,4 +14,10 @@
  * limitations under the License.
  *
  */
+
+/**
+ * This file should only contain versions for dependencies
+ */
+
 export const HELM_VERSION = 'v3.14.2'
+export const FST_CHART_VERSION = 'v0.22.0'
```
jeromy-cannon, line 23: I think this needs to move into a json or yaml file, and then parsed in.

nathanklick (approved):

PR #159, opened 2024-03-13
fix: refactor account initialization and mirror node setup from node command

### LINKED ISSUE #116: Implement mirror-node command
Currently mirror-node is setup during `node start`. We should separate it out in a separate command so that we have more control:

- `solo mirror-node start`
- `solo mirror-node stop`

`node start` does not need to deploy `mirror-node` anymore which would reduce delays on various account related tasks/tests.

### LINKED ISSUE #143: Refactor account keys update into Account Command

### LINKED ISSUE #151: Fix test error
https://swirldslabs.slack.com/archives/C05CU3X8T4P/p1709937005710819

### LINKED ISSUE #153: solo node start does not terminate
### Bill of Materials or CLI Version

0.22.0

### Execution Environment

Local Node CLI

### Slf4j Logging Implementation

None

### To Reproduce

```markdown
1. Run `solo node start`
```

### Describe the bug

`solo node start` does not terminate:

```
Version                 : 0.22.0
Kubernetes Context      : kind-solo
Kubernetes Cluster      : kind-solo
Kubernetes Namespace    : solo
(node:16442) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
✔ Initialize [0.8s]
✔ Identify network pods
  ✔ Check network pod: node0
  ✔ Check network pod: node1
  ✔ Check network pod: node2
✔ Starting nodes [0.1s]
  ✔ Start node: node0 [0.1s]
  ✔ Start node: node1 [0.1s]
  ✔ Start node: node2 [0.1s]
✔ Check nodes are ACTIVE [24s]
  ✔ Check node: node0 [24s]
  ✔ Check node: node1 [0.1s]
  ✔ Check node: node2 [0.1s]
✔ Enable mirror node [33s]
  ✔ Check node proxies are ACTIVE [7s]
    ✔ Check proxy for node: node0 [2s]
    ✔ Check proxy for node: node1 [1s]
    ✔ Check proxy for node: node2 [4s]
  ✔ Prepare address book [0.2s]
  ✔ Deploy mirror node [3s]
  ✔ Waiting for Hedera Explorer to be ready [22s]
✔ Update special account keys [54s]
  ✔ Prepare for account key updates
  ✔ Update special account key sets [54s]
    ✔ Updating set 1 of 30 [0.7s]
    ✔ Updating set 2 of 30 [2s]
    ✔ Updating set 3 of 30 [1s]
    ✔ Updating set 4 of 30 [2s]
    ✔ Updating set 5 of 30 [1s]
    ✔ Updating set 6 of 30 [1s]
    ✔ Updating set 7 of 30 [2s]
    ✔ Updating set 8 of 30 [1s]
    ✔ Updating set 9 of 30 [1s]
    ✔ Updating set 10 of 30 [1s]
    ✔ Updating set 11 of 30 [1s]
    ✔ Updating set 12 of 30 [1s]
    ✔ Updating set 13 of 30 [1s]
    ✔ Updating set 14 of 30 [2s]
    ✔ Updating set 15 of 30 [2s]
    ✔ Updating set 16 of 30 [1s]
    ✔ Updating set 17 of 30 [1s]
    ✔ Updating set 18 of 30 [1s]
    ✔ Updating set 19 of 30 [4s]
    ✔ Updating set 20 of 30 [1s]
    ✔ Updating set 21 of 30 [1s]
    ✔ Updating set 22 of 30 [1s]
    ✔ Updating set 23 of 30 [1s]
    ✔ Updating set 24 of 30 [1s]
    ✔ Updating set 25 of 30 [1s]
    ✔ Updating set 26 of 30 [1s]
    ✔ Updating set 27 of 30 [2s]
    ✔ Updating set 28 of 30 [1s]
    ✔ Updating set 29 of 30 [1s]
    ✔ Updating set 30 of 30 [0.5s]
  ✔ Display results

> upgrading chart: fullstack-deployment
OK chart 'fullstack-deployment' is upgraded
Account keys updated SUCCESSFULLY: 701
```

After execution, the program should exit and user should be able to run a new command.

### Whole JUnit/CLI Logs

```shell
NA
```

### Additional Context

_No response_

### LINKED ISSUE #165: Upgrade Hedera JS SDK
We need to upgrade it to `2.43.0-beta.1` because of our recent fix: https://github.com/hashgraph/hedera-sdk-js/pull/2195

We need to update it here: https://github.com/hashgraph/solo/blob/main/package.json#L29

leninmehedy:
## Description

* Isolate account initialization and mirror node setup from node command
* Remove unnecessary socket cleanup code (`_stopPortForwards`) as the actual issue was with hedera-sdk-js:
     * See: https://github.com/hashgraph/hedera-sdk-js/issues/2194
* Make node-command and account-command e2e tests independent from each other (i.e. use separate namespaces)
* Remove unused function `resetHapiDirectories` to avoid test errors: https://github.com/hashgraph/solo/issues/151

### Related Issues
* Closes #143 
* Closes #116 
* Closes #153 
* Closes #165 
* Closes #151 

### Todo:
* https://github.com/hashgraph/solo/issues/168
* https://github.com/hashgraph/solo/issues/169

### DIFF
--- a/src/core/account_manager.mjs
+++ b/src/core/account_manager.mjs
@@ -469,18 +469,15 @@ export class AccountManager {
     const self = this
 
     return new Promise((resolve, reject) => {
-      const socket = net.createConnection({
-        host,
-        port
-      })
-      socket.on('error', (e) => {
-        socket.destroy()
+      const s = new net.Socket()
+      s.on('error', (e) => {
+        s.destroy()
         reject(new FullstackTestingError(`failed to connect to '${host}:${port}': ${e.message}`, e))
       })
 
-      socket.on('connect', () => {
-        self.logger.debug(`Connected to port '${port}' of pod ${podName} at IP address ${host}`)
-        socket.destroy()
+      s.connect(port, host, () => {
+        self.logger.debug(`Connection test successful: ${host}:${port}`)
+        s.destroy()
         resolve(true)
       })
     })
--- a/src/core/account_manager.mjs
+++ b/src/core/account_manager.mjs
@@ -122,12 +122,12 @@ export class AccountManager {
    * batch up the accounts into sets to be processed
    * @returns {*[]} an array of arrays of numbers representing the accounts to update
    */
-  batchAccounts () {
+  batchAccounts (accountRange = constants.SYSTEM_ACCOUNTS) {
     const batchSize = constants.ACCOUNT_CREATE_BATCH_SIZE
     const batchSets = []
 
     let currentBatch = []
-    for (const [start, end] of constants.SYSTEM_ACCOUNTS) {
+    for (const [start, end] of accountRange) {
       let batchCounter = start
       for (let i = start; i <= end; i++) {
         currentBatch.push(i)
--- a/src/core/account_manager.mjs
+++ b/src/core/account_manager.mjs
@@ -30,7 +30,7 @@ import {
   Status,
   TransferTransaction
 } from '@hashgraph/sdk'
-import { FullstackTestingError } from './errors.mjs'
+import { FullstackTestingError, MissingArgumentError } from './errors.mjs'
 import net from 'net'
 import { Templates } from './templates.mjs'
 
@@ -395,6 +395,10 @@ export class AccountManager {
    * @returns {AccountInfo} the private key of the account
    */
   async accountInfoQuery (accountId) {
+    if (!this._nodeClient) {
+      throw new MissingArgumentError('node client is not initialized')
+    }
+
     return await new AccountInfoQuery()
       .setAccountId(accountId)
       .execute(this._nodeClient)
--- a/src/core/account_manager.mjs
+++ b/src/core/account_manager.mjs
@@ -31,10 +31,8 @@ import {
   TransferTransaction
 } from '@hashgraph/sdk'
 import { FullstackTestingError } from './errors.mjs'
-import { sleep } from './helpers.mjs'
 import net from 'net'
 import { Templates } from './templates.mjs'
-import { K8 } from './k8.mjs'
 
 const REASON_FAILED_TO_GET_KEYS = 'failed to get keys for accountId'
 const REASON_SKIPPED = 'skipped since it does not have a genesis key'
@@ -42,8 +40,6 @@ const REASON_FAILED_TO_UPDATE_ACCOUNT = 'failed to update account keys'
 const REASON_FAILED_TO_CREATE_K8S_S_KEY = 'failed to create k8s scrt key'
 const FULFILLED = 'fulfilled'
 const REJECTED = 'rejected'
-const MAX_PORT_FORWARD_SLEEP_ITERATIONS = 500
-const PORT_FORWARD_CLOSE_SLEEP = 500
 
 /**
  * Copyright (C) 2024 Hedera Hashgraph, LLC
@@ -130,16 +126,14 @@ export class AccountManager {
     const batchSize = constants.ACCOUNT_CREATE_BATCH_SIZE
     const batchSets = []
 
-    let batchCounter = 0
     let currentBatch = []
     for (const [start, end] of constants.SYSTEM_ACCOUNTS) {
+      let batchCounter = start
       for (let i = start; i <= end; i++) {
         currentBatch.push(i)
         batchCounter++
 
-        // push only a single account for the first to give time for the
-        // network to create all the system accounts and become stable
-        if (batchCounter >= batchSize || batchSets.length < 1) {
+        if (batchCounter % batchSize === 0) {
           batchSets.push(currentBatch)
           currentBatch = []
           batchCounter = 0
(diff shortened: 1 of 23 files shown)


leninmehedy:
> 0.00% <0.00%> (ø)



> [src/commands/mirror_node.mjs](https://app.codecov.io/gh/hashgraph/solo/pull/159?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=hashgraph#diff-c3JjL2NvbW1hbmRzL21pcnJvcl9ub2RlLm1qcw==)	`0.00% <0.00%> (ø)`

Need to write `test/e2e/mirror_node_e2e.test.mjs`

src/core/account_manager.mjs
```diff
@@ -126,20 +122,18 @@ export class AccountManager {
...
-    for (const [start, end] of constants.SYSTEM_ACCOUNTS) {
+    for (const [start, end] of accountRange) {
+      let batchCounter = start
       for (let i = start; i <= end; i++) {
         currentBatch.push(i)
         batchCounter++
 
-        // push only a single account for the first to give time for the
-        // network to create all the system accounts and become stable
-        if (batchCounter >= batchSize || batchSets.length < 1) {
```
jeromy-cannon, line 142: I added `batchSets.length < 1` because I noticed that the Hedera Services code is lazy initializing all of the system accounts when the first account update or create comes through.  If a batch of transactions hit it at the same time, it occasionally locks up the nodeClient and cause timeouts.  Has this been running successfully both locally and in the pipeline consistently?

jeromy-cannon: I see you made it contiguous, which would add a lot of stability at the cost of speed.  I'm guessing we won't need the account key update very often, so it is worth it?

src/core/k8.mjs
```diff
@@ -700,13 +665,24 @@ export class K8 {
...
+  }
 
-    return server.listen(localPort, '127.0.0.1')
+  /**
+   * Stop the port forwarder server
+   *
+   * @param server an instance of server returned by portForward method
+   * @return {Promise<void>}
+   */
+  async stopPortForward (server) {
```
jeromy-cannon, line 683: I don't think this needs to be async with just calling server.close().  Server.close() is some nasty old cold that uses event management, so when it returns back, it isn't closed.   It will close once it has time to process in the callback chain.

src/core/account_manager.mjs
jeromy-cannon: If you fixed the root cause with the logger in the hedera-sdk library, then why wouldn't we want to keep the asynchronous code?  It would run a lot faster.  Is it still having issues, or did you make it synchronous before figuring out the root cause?

src/core/constants.mjs
```diff
@@ -81,7 +82,7 @@ export const GENESIS_KEY = process.env.GENESIS_KEY || '302e020100300506032b65700
 export const SYSTEM_ACCOUNTS = [[3, 100], [200, 349], [400, 750], [900, 1000]] // do account 0.0.2 last and outside the loop
 export const TREASURY_ACCOUNT = 2
 export const LOCAL_NODE_START_PORT = process.env.LOCAL_NODE_START_PORT || 30212
-export const ACCOUNT_CREATE_BATCH_SIZE = process.env.ACCOUNT_CREATE_BATCH_SIZE || 25
+export const ACCOUNT_CREATE_BATCH_SIZE = process.env.ACCOUNT_CREATE_BATCH_SIZE || 50
```
jeromy-cannon, line 85: the code is no longer using this as a 'batch size', so I don't think the name is appropriate anymore.

src/core/account_manager.mjs
leninmehedy: > did you make it synchronous before figuring out the root cause?

Good question, I don't remember exactly. My original suspicion was that same _nodeClient was being used to invoke transactions in "parallel" and probably some internal state was causing trouble. 

Sequential operation made it a lot simpler to pinpoint the hedera-sdk bug. So far, it is very stable.

We can reintroduce the batch and see how stable that is.

src/core/account_manager.mjs
```diff
@@ -126,20 +122,18 @@ export class AccountManager {
...
-  batchAccounts () {
+  batchAccounts (accountRange = constants.SYSTEM_ACCOUNTS) {
     const batchSize = constants.ACCOUNT_CREATE_BATCH_SIZE
     const batchSets = []
 
-    let batchCounter = 0
     let currentBatch = []
-    for (const [start, end] of constants.SYSTEM_ACCOUNTS) {
+    for (const [start, end] of accountRange) {
+      let batchCounter = start
```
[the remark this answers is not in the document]
leninmehedy: It helps with unit testing as we can pass a smaller range.

src/core/constants.mjs
leninmehedy: Well, it still tries to process in a sequential 'group' or batch. So it still makes sense. We can reintroduce the 'parallel' update and see if that's stable.

src/core/k8.mjs
leninmehedy: Ok, thanks. I'll remove and do some testing.

src/core/account_manager.mjs
leninmehedy: I've reintroduced the parallel account update. It reduced the execution time significantly (32 seconds compared to >8minutes). There was a spike of open sockets (from 120 to 260) as expected, but such as spike of few hundred sockets is ok I think.

Screenshot 2024-03-19 at 2 30 03 PM

src/core/k8.mjs
leninmehedy: I've removed `async` and made the function return a proper promise that handles the callback of that close function.

src/core/account_manager.mjs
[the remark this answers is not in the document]
leninmehedy: Other option is to pass it as constructor arg and cache it in the `AccountManager`.

jeromy-cannon (approved):

src/core/k8.mjs
jeromy-cannon: yes, I dug really deep into that code, because I thought it was the issue causing the node start command to not exit.  it is really meant to be run in a browser, not NodeJS.  I think it used because it works in both and creates code reuse.

PR #173, opened 2024-03-20
fix: update HAProxy UP check to use HAProxy Data Plane API

### LINKED ISSUE #169: HAProxy UP check is unreliable
Currently HAProxy status is checked using the log. However, it often fails to log the message because probably it gives up retrying....Also if HAProxy is restarted, it looses the logs and does not output the line `Server be_servers/server1 is UP` consistently (See https://github.com/hashgraph/solo/blob/c6474de87a2f1a14fd393a9cbe614f606ebbf3fc/src/commands/node.mjs#L557). 

As a result, `solo` fails to report if node start was successful.

My initial solution was to restart HAProxy, however even after manually restarting, it didn't solve the issue as HAProxy didn't output the log message `Server be_servers/server1 is UP` although sdk was able to invoke transactions. When I viewed `/stats`, it did report that node was up.

We need to figure out a reliable way to check if HAProxy is UP or not. Not sure if we can use the HAProxy `/stats` endpoint?, or use hedera-sdk to poll? 

Notes:
- If you would like to view the `/stats` endpoint, it is a bit complicated as below
  - Expose port 9090 of the HAProxy service (e.g. `haproxy-node0-svc`)
  - Open the corresponding configmap and at the bottom you will find the `auth` block separated by `:`. First part is username and the second part is the password.
  - Go to `localhost:9090/stats` and when prompted enter the username and password.

jeromy-cannon:
## Description

* update HAProxy UP check to use HAProxy Data Plane API

### Related Issues

* Closes #169
* Relates to https://github.com/hashgraph/full-stack-testing/pull/809

### DIFF
--- a/src/commands/node.mjs
+++ b/src/commands/node.mjs
@@ -656,7 +656,7 @@ export class NodeCommand extends BaseCommand {
           for (const nodeId of ctx.config.nodeIds) {
             subTasks.push({
               title: `Check proxy for node: ${chalk.yellow(nodeId)}`,
-              task: async () => await self.checkNetworkNodeProxyUp(ctx.config.namespace, nodeId, localPort++)
+              task: async () => await self.checkNetworkNodeProxyUp(nodeId, localPort++)
             })
           }
 
@@ -688,14 +688,13 @@ export class NodeCommand extends BaseCommand {
 
   /**
    * Check if the network node proxy is up, requires close() to be called after
-   * @param namespace the namespace
    * @param nodeId the node id
    * @param localPort the local port to forward to
    * @param maxAttempts the maximum number of attempts
    * @param delay the delay between attempts
    * @returns {Promise<boolean>} true if the proxy is up
    */
-  async checkNetworkNodeProxyUp (namespace, nodeId, localPort, maxAttempts = 10, delay = 5000) {
+  async checkNetworkNodeProxyUp (nodeId, localPort, maxAttempts = 10, delay = 5000) {
     const podArray = await this.k8.getPodsByLabel([`app=haproxy-${nodeId}`, 'fullstack.hedera.com/type=haproxy'])
 
     let attempts = 0
--- a/src/commands/node.mjs
+++ b/src/commands/node.mjs
@@ -696,12 +696,12 @@ export class NodeCommand extends BaseCommand {
    * @returns {Promise<boolean>} true if the proxy is up
    */
   async checkNetworkNodeProxyUp (namespace, nodeId, localPort, maxAttempts = 10, delay = 5000) {
-    const podArray = await this.k8.getPodsByLabel(namespace, [`app=haproxy-${nodeId}`, 'fullstack.hedera.com/type=haproxy'])
+    const podArray = await this.k8.getPodsByLabel([`app=haproxy-${nodeId}`, 'fullstack.hedera.com/type=haproxy'])
 
     let attempts = 0
     if (podArray.length > 0) {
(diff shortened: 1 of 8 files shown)


src/core/k8.mjs
```diff
@@ -197,14 +197,14 @@ export class K8 {
 
   /**
    * Get pods by labels
+   * @param namespace the namespace of the pods
    * @param labels list of labels
    * @return {Promise<Array<V1Pod>>}
    */
-  async getPodsByLabel (labels = []) {
-    const ns = this._getNamespace()
+  async getPodsByLabel (namespace, labels = []) {
```
leninmehedy, line 204: Not sure we need to pass namespace in the k8 function since k8 instances picks it from configManager. So as long as configManager is up-to-date, it should work.

src/core/k8.mjs
```diff
@@ -661,19 +661,45 @@ export class K8 {
    * @param podName pod name
    * @param localPort local port
    * @param podPort port of the pod
+   * @param namespace namespace of the pod (optional)
    */
-  async portForward (podName, localPort, podPort) {
-    const ns = this._getNamespace()
+  async portForward (podName, localPort, podPort, namespace = null) {
+    const ns = namespace || this._getNamespace()
```
leninmehedy, line 667: same comment on passing the namespace as above.

src/commands/node.mjs
```diff
@@ -664,45 +680,48 @@ export class NodeCommand extends BaseCommand {
...
+  /**
+   * Check if the network node proxy is up, requires close() to be called after
+   * @param namespace the namespace
+   * @param nodeId the node id
+   * @param localPort the local port to forward to
+   * @param maxAttempts the maximum number of attempts
+   * @param delay the delay between attempts
+   * @returns {Promise<boolean>} true if the proxy is up
+   */
+  async checkNetworkNodeProxyUp (namespace, nodeId, localPort, maxAttempts = 10, delay = 5000) {
```
leninmehedy, line 698: `namespace` is already in configManager and k8 has access to it. so no need to really pass it to k8.

jeromy-cannon: I needed it for overriding in the test case when I was testing locally and had it pointed to a different cluster.

src/core/k8.mjs
jeromy-cannon: same response

src/core/k8.mjs
jeromy-cannon: same response

src/commands/node.mjs
jeromy-cannon: Also, stateless/functional programming is more scalable in the long run.  The current way provides more flexibility when testing and future usage.

leninmehedy: In the test, we could just call `configManager.setFlag(flags.namespace, <namespace>)` to set new namespace. This is how other tests are testing different namespace scenario: see an example [here](https://github.com/hashgraph/solo/pull/127/files#diff-f5f3a03bb79f5baee73a5e182b7a8bd32b8c9bf587d5f52f9fc294da986e73dcR79). Let me know if that does not solve your test scenario.

leninmehedy: If we pass `namespace` in those two methods of K8, we should update the API for all methods in K8 for consistency.

jeromy-cannon: updated

leninmehedy (approved):
Thanks for this very important fix.

PR #234, opened 2024-04-16
feat: add `solo node add` subcommand

### LINKED ISSUE #115: Implement node add command
While `solo node refresh` is useful, but we need to allow adding a new node (e.g. support `solo node add`).
`solo` state should maintain a list of all members so that we can generate config.txt when provisioning a new node or refreshing a node.

`solo node add` flow:
1. Use helm chart to deploy a new pod
2. Setup the new node (with correct `config.txt`)
3. Send `config.txt` and public keys to all nodes
4. Send freeze transaction to all nodes across all clusters
5. Wait for 2/3 majority of nodes to FREEZE.
6. Restart/Recycle all node (`systemctl restart network-node`, for deterministic image it would be to kill the pod).

_Originally posted by @leninmehedy in https://github.com/hashgraph/solo/issues/96#issuecomment-1970199272_

jeromy-cannon:
## Description

* add `solo node add` subcommand

### Related Issues

* Closes #115

### DIFF
--- a/src/commands/node.mjs
+++ b/src/commands/node.mjs
@@ -1127,7 +1127,7 @@ export class NodeCommand extends BaseCommand {
         }
       },
       {
-        title: 'Identify existing network pods',
+        title: 'Identify existing network nodes',
         task: async (ctx, task) => {
           ctx.config.serviceMap = await self.accountManager.getNodeServiceMap(
             ctx.config.namespace)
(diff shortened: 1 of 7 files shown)


src/commands/node.mjs
```diff
@@ -1040,28 +1009,347 @@ export class NodeCommand extends BaseCommand {
...
+              collapseSubtasks: false
+            }
+          })
+        }
+      },
+      {
+        title: 'Generate Gossip keys',
+        task: async (ctx, parentTask) => {
+          const config = ctx.config
+          const subTasks = self._nodeGossipKeysTaskList(config.keyFormat, config.allNodeIds, config.keysDir, config.curDate, false)
```
leninmehedy, line 1203: Do we really need to generate keys for all nodes? 

We should only need to generate keys for the new node and send its public key to all other nodes.

leninmehedy (changes requested):

src/commands/node.mjs
jeromy-cannon: Good call.  I thought I added the logic, but on review my code was incomplete.  I've added the code and also added a test case to verify that the existing keys are maintained.

leninmehedy (approved):
Thank you!

PR #240, opened 2024-04-19
fix: simplify stop port forward function

leninmehedy:
## Description

* remove unnecessary `async` and `await new Promise....`
* remove the port open check as the `close` event will be emitted when the port is closed (I know we've found it to be not behaving as it says although doc says the `close` event will be emitted: https://nodejs.org/api/net.html#serverclosecallback). 

### Related Issues

* Closes #


leninmehedy:
Closing it as I'm not working on it anymore.

PR #261, opened 2024-05-06
ci: increase chart version

JeffreyDallas:
## Description

* Increase chart version so we can use newer version of mirror node chart, which takes less time to spin up GRPC

### Related Issues

* Closes #

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hashgraph/solo",
-  "version": "0.24.0",
+  "version": "0.24.5",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hashgraph/solo",
-  "version": "0.24.0",
+  "version": "0.24.5",
```
jeromy-cannon, line 3: shouldn't this be auto-updating as part of our artifact build during tag/release?

JeffreyDallas:
Lenin already updated chart version in another PR.

PR #285, opened 2024-05-09
ci: split e2e tests

jeromy-cannon:
## Description

* split `.github/workflows/zxc-compile-code.yaml` to three separate reusable workflows:
  * `.github/workflows/zxc-code-style.yaml`
  * `.github/workflows/zxc-e2e-test.yaml`
  * `.github/workflows/zxc-unit-test.yaml`
* made separate test-e2e runs for:
  * `/e2e/commands/mirror_node.test.mjs`
  * `/e2e/commands/node.test.mjs`
  * `/e2e/commands/relay.test.mjs` 
* reduce number of nodes to one where applicable 
* fixed bug in `.github/workflows/flow-update-readme.yaml` where wget was not available for kind setup

### Related Issues

* Closes #


.github/workflows/flow-build-application.yaml
```diff
@@ -43,34 +43,29 @@ defaults:
     shell: bash
 
 jobs:
-  code:
```
nathanklick, line 46: This should not be removed.

nathanklick (changes requested):

.github/workflows/flow-build-application.yaml
jeromy-cannon: it is empty.  All it does it checkout the code and do an `npm ci`, it doesn't actually compile anything.

jeromy-cannon: code-style is a dependency and will do the same

nathanklick (approved):

PR #310, opened 2024-05-16
feat: mount application.env as a configMap

### LINKED ISSUE #110: `application.env` should be mounted from a config-map to `/etc/network-node/` directory
```
As discussed, we shall need to work on a separate PR where `application.env` will be mounted from a config-map to `/etc/network-node/` directory. 

This will also require updating the docker images in `full-stack-testing/docker` to ensure `application.env` is loaded from `/etc/network-node/`:
`docker/ubi8-init-java21/network-node.service`
`docker/ubi8-init-java21/Dockerfile`
```
_Originally posted by @leninmehedy in https://github.com/hashgraph/solo/pull/84#discussion_r1506596909_
            

- [ ] Requires #https://github.com/hashgraph/full-stack-testing/pull/797

jeromy-cannon:
## Description

* mount `/etc/network-node/application.env` as a configMap to the network nodes

### Related Issues

* Closes #110

### DIFF
--- a/src/core/platform_installer.mjs
+++ b/src/core/platform_installer.mjs
@@ -234,14 +234,19 @@ export class PlatformInstaller {
     if (!podName) throw new MissingArgumentError('podName is required')
     if (!destPath) throw new MissingArgumentError('destPath is required')
 
+    const recursiveFlag = recursive ? '-R' : ''
     try {
-      const recursiveFlag = recursive ? '-R' : ''
       await this.k8.execContainer(podName, container, `chown ${recursiveFlag} hedera:hedera ${destPath}`)
-      await this.k8.execContainer(podName, container, `chmod ${recursiveFlag} ${mode} ${destPath}`)
-      return true
     } catch (e) {
-      throw new FullstackTestingError(`failed to set permission in '${podName}': ${destPath}`, e)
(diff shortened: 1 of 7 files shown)


JeffreyDallas (approved):

src/core/platform_installer.mjs
```diff
@@ -234,14 +234,19 @@ export class PlatformInstaller {
     if (!podName) throw new MissingArgumentError('podName is required')
     if (!destPath) throw new MissingArgumentError('destPath is required')
 
+    const recursiveFlag = recursive ? '-R' : ''
     try {
-      const recursiveFlag = recursive ? '-R' : ''
       await this.k8.execContainer(podName, container, `chown ${recursiveFlag} hedera:hedera ${destPath}`)
+    } catch (e) {
```
leninmehedy, line 240: Not sure silencing errors is solid solution.. Maybe create a separate ticket (low priority) to fine tune this by creating a list of files that needs to be changed using `chown` command here.

leninmehedy (approved):

src/core/platform_installer.mjs
jeromy-cannon: good catch: https://github.com/hashgraph/solo/issues/338

PR #327, opened 2024-05-24
fix: fix freeze timestamp for node add command

### LINKED ISSUE #326: Incorrect freeze timestamp used in node add
### Describe the bug

Current implementation `futureDate.setDate(futureDate.getDate() + (1 / 24 / 60))`
actually would set freeze time as exactly current time, instead of one minute layer.
This is due to both setDate and getDate are dealing with integer.
so `interger + (1 / 24 / 60)` resulted in the same integer, so the operation did not change
date or hours, giving the exactly current timestamp as freeze timestamp.

The transaction receipt if enabled would give error `FREEZE_START_TIME_MUST_BE_FUTURE`

Should set correct timestamp and receive SUCEED as transaction receipt

### Whole JUnit/CLI Logs

```shell
No
```

### Additional Context

_No response_

JeffreyDallas:
## Description

* Fixed timestamp setting since SetDate and GetDate will not provide a unix timestamp, but a day number within a month

### Related Issues

* Closes #326

### DIFF
--- a/src/commands/node.mjs
+++ b/src/commands/node.mjs
@@ -77,7 +77,7 @@ export class NodeCommand extends BaseCommand {
     const podName = Templates.renderNetworkPodName(nodeId)
 
     try {
-      await this.k8.waitForPodReady([
+      await this.k8.waitForPods([constants.POD_PHASE_RUNNING], [
         'fullstack.hedera.com/type=network-node',
         `fullstack.hedera.com/node-name=${nodeId}`
       ], 1, maxAttempts, delay)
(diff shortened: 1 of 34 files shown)


leninmehedy (dismissed):

src/commands/node.mjs
```diff
@@ -1467,9 +1470,13 @@ export class NodeCommand extends BaseCommand {
...
         .execute(client)
 
-      this.logger.debug(`Upgrade frozen with transaction id: ${freezeUpgradeTx.transactionId.toString()}`)
+      const freezeUpgradeReceipt = await freezeUpgradeTx.getReceipt(client)
+
+      this.logger.debug(`Upgrade frozen with transaction id: ${freezeUpgradeTx.transactionId.toString()}`,
+        freezeUpgradeReceipt.status.toString())
     } catch (e) {
       this.logger.error(`Error in freeze upgrade: ${e.message}`, e)
+      await sleep(360000)
```
jeromy-cannon, line 1479: what is the purpose for sleeping for 6 minutes?  when it throws the error, then it will exit the app and stop processing.

jeromy-cannon (changes requested):

src/commands/node.mjs
JeffreyDallas: was added during debug, removed it.

jeromy-cannon (approved):

PR #337, opened 2024-05-30
feat: add node logs command

### LINKED ISSUE #336: Add a node logs command to download swirlds.log and hgcaa.log
### Requested Feature

```markdown
Download node logs before tests were cleaned up
```

### Business Need

```markdown
Help to debug issues when hedera network node could not enter ACTIVE mode or FREEZE mode
```

### Additional Context

_No response_

JeffreyDallas:
## Description

* Add help function to download swirlds.log and hgcaa.log
* Add logs command so we can use `solo node logs` to download logs to ~/.solo directory

### Related Issues

* Closes #336

### DIFF
--- a/.github/workflows/zxc-e2e-test.yaml
+++ b/.github/workflows/zxc-e2e-test.yaml
@@ -182,7 +182,7 @@ jobs:
         uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
         with:
           name: solo-${{ inputs.npm-test-script }}.log
-          path: ~/.solo/logs/solo.log
+          path: ~/.solo/logs/*
           overwrite: true
           if-no-files-found: error
 
--- a/src/commands/node.mjs
+++ b/src/commands/node.mjs
@@ -1092,7 +1092,7 @@ export class NodeCommand extends BaseCommand {
       {
         title: 'Copy logs from all nodes',
         task: (ctx, _) => {
-          getNodeLogs(this.k8)
+          getNodeLogs(this.k8, flags.namespace)
         }
       }
     ], {
@@ -1687,7 +1687,7 @@ export class NodeCommand extends BaseCommand {
           })
           .command({
             command: 'logs',
-            desc: 'Download logs from nodes',
+            desc: 'Download application logs from the network nodes and stores them in <SOLO_LOGS_DIR>/<namespace>/<podName>/ directory',
             builder: y => flags.setCommandFlags(y,
               flags.nodeIDs
             ),
--- a/src/commands/node.mjs
+++ b/src/commands/node.mjs
@@ -21,7 +21,7 @@ import { Listr } from 'listr2'
 import path from 'path'
 import { FullstackTestingError, IllegalArgumentError } from '../core/errors.mjs'
 import * as helpers from '../core/helpers.mjs'
-import { getTmpDir, sleep, validatePath } from '../core/helpers.mjs'
+import { getNodeLogs, getTmpDir, sleep, validatePath } from '../core/helpers.mjs'
 import { constants, Templates } from '../core/index.mjs'
 import { BaseCommand } from './base.mjs'
 import * as flags from './flags.mjs'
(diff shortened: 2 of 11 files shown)


src/core/helpers.mjs
```diff
@@ -182,3 +183,20 @@ export function validatePath (input) {
   }
   return input
 }
+
+/**
+ * Download logs files from all network pods and save to local solo log directory
+ * @param k8
+ *    an instance of core/K8
+ * @returns A promise that resolves when the logs are downloaded
```
jeromy-cannon, line 191:
```suggestion
@param {K8} k8 an instance of core/K8
@returns {Promise<void>} A promise that resolves when the logs are downloaded
```

src/core/helpers.mjs
```diff
@@ -182,3 +183,20 @@ export function validatePath (input) {
...
+ */
+export async function getNodeLogs (k8) {
+  const pods = await k8.getPodsByLabel(['fullstack.hedera.com/type=network-node'])
+  for (const pod of pods) {
+    const podName = pod.metadata.name
+    const targetDir = `${SOLO_LOGS_DIR}/${podName}`
```
jeromy-cannon, line 197: I'm thinking to include the namespace, something like: 
`${SOLO_LOGS_DIR}/${namespace}/${podName}`

src/core/helpers.mjs
```diff
@@ -182,3 +183,20 @@ export function validatePath (input) {
...
+    fs.mkdirSync(targetDir, { recursive: true })
```
jeromy-cannon, line 198: if the directory already exists do we want to remove the contents or just overwrite them an possibly leave an existing file in place?

src/core/helpers.mjs
```diff
@@ -182,3 +183,20 @@ export function validatePath (input) {
...
+    await k8.copyFrom(podName, ROOT_CONTAINER, `${HEDERA_HAPI_PATH}/output/swirlds.log`, targetDir)
+    await k8.copyFrom(podName, ROOT_CONTAINER, `${HEDERA_HAPI_PATH}/output/hgcaa.log`, targetDir)
```
jeromy-cannon, line 200: it might be useful to put this into a try/catch block, log the error, then push it into an array.  after the for loop if there are any errors in the array throw an exception saying which nodes failed.

That way, if the first node fails, we could still get the logs for the other nodes which might be useful to diagnose the issue.

test/e2e/e2e_node_util.js
```diff
@@ -75,6 +76,7 @@ export function e2eNodeKeyRefreshAddTest (keyFormat, testName, mode, releaseTag
     }, defaultTimeout)
 
     afterAll(async () => {
+      await getNodeLogs(k8)
```
jeromy-cannon, line 79: I think it would be useful to also add this into:
- [ ] `network.test.mjs`
- [ ] `account.test.mjs`
- [ ] `cluster.test.mjs`
- [ ] `relay.test.mjs`

.github/workflows/zxc-e2e-test.yaml
```diff
@@ -182,7 +182,7 @@ jobs:
         uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
         with:
           name: solo-${{ inputs.npm-test-script }}.log
-          path: ~/.solo/logs/solo.log
+          path: ~/.solo/logs/*
```
jeromy-cannon, line 185: assuming that you add the namespace as I mentioned below, then will want to retest this to make sure it uploads subdirectories and their files

src/commands/node.mjs
```diff
@@ -1646,6 +1685,25 @@ export class NodeCommand extends BaseCommand {
               })
             }
           })
+          .command({
+            command: 'logs',
+            desc: 'Download logs from nodes',
```
jeromy-cannon, line 1690:
```suggestion
            desc: 'Download application logs from the network nodes and stores them in <SOLO_LOGS_DIR>/<namespace>/<podName>/ directory
```

src/core/helpers.mjs
JeffreyDallas: Updated

src/core/helpers.mjs
JeffreyDallas: Added removing the target directory first

src/core/helpers.mjs
JeffreyDallas: Added try catch block
But did not throw exception so the test can continue to move on with deleting namespace / shutting down properly

test/e2e/e2e_node_util.js
JeffreyDallas: Updated

.github/workflows/zxc-e2e-test.yaml
JeffreyDallas: updated and tested

src/core/helpers.mjs
```diff
@@ -182,3 +183,30 @@ export function validatePath (input) {
   }
   return input
 }
+
+/**
+ * Download logs files from all network pods and save to local solo log directory
+ * @param k8
```
jeromy-cannon, line 189:
```suggestion
 * @param {K8} k8
```

jeromy-cannon (approved):

src/core/helpers.mjs
JeffreyDallas: Updated

jeromy-cannon (approved):

PR #342, opened 2024-05-31
feat: node log command also download config.txt and settings.txt

### LINKED ISSUE #336: Add a node logs command to download swirlds.log and hgcaa.log

JeffreyDallas:
## Description

* Minor change to also download config.txt and settings.txt
* Update setting.txt to align with latest on hedera repo
* update how to use name space in calling getNodeLogs()

### Related Issues

* Closes #336


resources/templates/settings.txt
```diff
@@ -1,25 +1,15 @@
 checkSignedStateFromDisk,                      1
 csvFileName,                                   MainNetStats
-csvOutputFolder,                               data/stats
```
jeromy-cannon, line 3: question: are the changes to this file backwards compatible?  Just want to make sure we won't indirectly stop supporting older versions.  I think previously we had 0.42.5 working.  Although, I don't really see it referenced in any of our e2e tests anymore.

JeffreyDallas: HEDERA_PLATFORM_VERSION used by e2e tests already has been changed to 'v0.49.0
so it should work with v0.49
If someone has to run some solo with on old version hedera repo, then the local settings.txt can be
modified or customized

jeromy-cannon (approved):

PR #364, opened 2024-06-07
refactor: cleanup flag usage in `*Command` classes

### LINKED ISSUE #354: The flags, config, and prompt logic in the commands is difficult to determine what is present, what is missing, and what is no longer needed
as part of #136 I need to move flags from NodeCommand to NetworkCommand, I will also need to delete various functions.  While doing this I noticed it was challenging to see where the flags were being used or not used.  Additionally, the flag values were stored in ctx.config as an object, and it was challenging to determine if these were being used anywhere else or if it was dead code.  Need to come up with a better way to:
- flags should be sorted in the command list to make it easy to compare between commands
- add a flag disablePrompt so that we don't have to keep a separate list of what needs to be prompted and not prompted
- construct code in such a way that we can determine which flags are being used and not used within a command and subcommand

jeromy-cannon:
## Description

* add logic in `BaseCommand` to generate a dynamic config class instance based on the supplied flags and extra properties
* add `constName` in flags to aid in property names for new dynamic config class
* rename command configs to match flag names
* add ability to disable a prompt for a given flag

### Related Issues

* Closes #354


src/commands/node.mjs
jeromy-cannon, line 1: TODO remember to add checks for unused flags in tests

src/commands/base.mjs
```diff
@@ -61,5 +61,93 @@ export class BaseCommand extends ShellRunner {
...
+   * Dynamically builds a class with properties from the provided list of flags
+   * and extra properties, will keep track of which properties are used.  Call
+   * getUnusedConfigs() to get an array of unused properties.
+   *
+   * @param {string} configName the name of the configuration
+   * @param {CommandFlag[]} flags an array of flags
+   * @param {string[]} [extraProperties] an array of extra properties
+   * @returns {Object} the instance of the new class
+   */
+  getConfig (configName, flags, extraProperties = []) {
```
jeromy-cannon, line 77: FYI, Nathan and I reviewed this in a 1:1.  He is okay with how I'm using this here and what I'm using it for.  It is more of an exception though, and not necessarily to be used as a pattern.

jeromy-cannon, line 77: FYI, Nathan and I reviewed this in a 1:1.  He is okay with how I'm using this here and what I'm using it for.  It is more of an exception though, and not necessarily to be used as a pattern.

leninmehedy: I am most likely missing something here as I don't understand the need to have this dynamic class generation and if I ever look at or touch this code, most likely I shall be tempted to refactor.

If we need a class to track the used vs unused flags, we could just have a config class with the full set of configs (so a centralized management of all flags), then the constructor would take a list of flags and mark the corresponding flags as used (or mark the other ones as unused), and then `getUnusedConfigs` would return the unused flags.

However, if two of you are ok with this, then let's go with it.

leninmehedy (approved):

src/commands/base.mjs
jeromy-cannon: @leninmehedy , thank you for the review and feedback.  I do like what you suggested.  I think I was also trying to have the goal of having the IDE provide tool tips.  Like this:
image
Which also requires the JSDoc to work correctly.  However, you lose the IDE tool tips here:
image
Unless you do something like this:
image

PR #365, opened 2024-06-07
fix: use 0.0.58 freeze admin account for freeze transaction

### LINKED ISSUE #308: Node add stops working when you run account init before
The network consensus nodes are configured to use a specific account ID to do the freeze transaction.  Currently, Solo is using the treasury account 0.0.2.  It needs to be updated to use the default freeze transaction account 0.0.58.

To recreate the issue:
```
  SOLO_CLUSTER_NAME=solo-e2e
  SOLO_NAMESPACE=solo-e2e
  SOLO_CLUSTER_SETUP_NAMESPACE=solo-e2e-cluster
  FST_CHARTS_DIR=~/source/full-stack-testing/charts
  kind delete cluster -n "${SOLO_CLUSTER_NAME}" || true
  kind create cluster -n "${SOLO_CLUSTER_NAME}" || return
  solo init -d ${FST_CHARTS_DIR} --namespace "${SOLO_NAMESPACE}" -i node0,node1,node2 -t v0.47.0 -s "${SOLO_CLUSTER_SETUP_NAMESPACE}" || return
  solo node keys --gossip-keys --tls-keys --key-format pem || return
  solo cluster setup  || return
  solo network deploy || return
  solo node setup || return
  solo node start || return
  solo account init || return
  solo node add -i node3 || return
```

JeffreyDallas:
## Description

* Changed to use 0.0.58 freeze admin account to sent freeze upgrade transactions

### Related Issues

* Closes #308


src/core/constants.mjs
```diff
@@ -79,6 +79,7 @@ export const DEFAULT_CHART_REPO = new Map()
 
 // ------------------- Hedera Account related ---------------------------------------------------------------------------------
 export const OPERATOR_ID = process.env.SOLO_OPERATOR_ID || '0.0.2'
+export const FREEZE_ADMIN_ACCOUNT = '0.0.58'
```
jeromy-cannon, line 82: allow user override with env var, I noticed that this is actually configurable in Hedera Services.

```suggestion
export const FREEZE_ADMIN_ACCOUNT = process.env.FREEZE_ADMIN_ACCOUNT || '0.0.58'
```

jeromy-cannon (dismissed):

src/commands/node.mjs
```diff
@@ -1472,8 +1474,21 @@ export class NodeCommand extends BaseCommand {
...
+
+      // query the balance
+      const balance = await new AccountBalanceQuery()
+        .setAccountId(FREEZE_ADMIN_ACCOUNT)
+        .execute(this.accountManager._nodeClient)
+      this.logger.debug(`Freeze admin account balance: ${balance.hbars}`)
+
+      // set operator of freeze transaction as freeze admin account
+      const accountKeys = await this.accountManager.getAccountKeysFromSecret(FREEZE_ADMIN_ACCOUNT, config.namespace)
+      const freezeAdminPrivateKey = accountKeys.privateKey
```
JeffreyDallas, line 1489: E2E test of adding node failed here due to accoutnKeys is "undefined",
seems 0.0.58 accout does not preexist yet.
If I added node init() first, then there are error during update account

```
failed to update account keys for accountId 0.0.300
```
[solo.log](https://github.com/user-attachments/files/15755290/solo.log)

jeromy-cannon: in your logs it looks like `solo account init` is failing.  Does it ever pass?  Originally, I had it running the full set of accounts in an E2E test, but Lenin reduced it to a subset to make it run faster.  So, we haven't been running the full set on a regular basis.

JeffreyDallas: If running the following commands from terminal, everything is working,


```
  kind delete cluster -n "${SOLO_CLUSTER_NAME}" || true
  kind create cluster -n "${SOLO_CLUSTER_NAME}" || return
  solo init -d ${FST_CHARTS_DIR} --namespace "${SOLO_NAMESPACE}" -i node0,node1,node2 -t v0.49.0-alpha.2 -s "${SOLO_CLUSTER_SETUP_NAMESPACE}" || return
  solo node keys --gossip-keys --tls-keys --key-format pem || return
  solo cluster setup  || return
  solo network deploy || return
  solo node setup || return
  solo node start || return
  solo account init || return
  solo node add -i node3 || return
```

but E2E add not test did not have `solo account init` part, so i added:
```
      it('should succeed with init command', async () => {
        const status = await accountCmd.init(argv)
        expect(status).toBeTruthy()
      }, 180000)
```

jeromy-cannon: originally when I wrote this code, I found that the first time the nodes were doing a write transaction they would create the system accounts and pause for a few hundred milliseconds.  This caused an issue where the grpc calls all being released at once would fail.

Later, when Lenin was trying to fix the logic, he removed all of that logic and instead of running them in parallel, just made them run sequentially.  Later, he found the bug was actually in @hashgraph/sdk, and put in a fix to correct the loggers they were creating and not closing.  I asked him to re-enable the parallel, which he did.  But, I don't think he updated the logic to do the initial single account update call before releasing the rest.  He also removed the throttling I had. The throttling was less than ideal, however, it helped me get past the issue created by the logger leak (which I did not know existed at that time), and prevented me from overwhelming the consensus nodes.

I'm guessing that this might be related to either:
1) the consensus nodes pause to do genesis system account creation logic and rejecting the flood of grpc calls
2) the haproxy and/or consensus node is being overwhelmed with connections and not keeping up correctly

In the case that it is #1, you can check the timestamps.  I believe in the hgcaa.log, you can see when it is creating the system accounts and line that up with your grpc call timestamps.

If it is #2, you could try figuring out how to fix the consensus node or haproxy to handle the connections without error; or, you could figure out how to throttle the transactions to prevent them from releasing too many at once.

It has been awhile since I looked at this code, I think Lenin in one version of the code released a certain number of calls before waiting and releasing some more.  If that code is still there, you could reduce that number.

JeffreyDallas: Thanks for the clarification.
Just wonder why the same piece of code was called
 when `solo account init` is used in command line, working fine 

But when called using accountCmd.init(argv), it failed.
The throttling, or parallel/sequential mechanism should be the same when used in the above two
different scenario.

jeromy-cannon: when you run from the command line, it will dump the configs into the logs file.  I think it also dumps them when you run accountCmd.init(argv).  You could take the rows and compare them to see what parameters might be different.

JeffreyDallas:
To make sure we compare apple to apple I created a script `./cmdline.sh` to setup cluster, network and nodes.

Then if following by `solo account init || return`, updateAccount each batch finishied in 2~3 seconds.

if followed by a modified e2e test `npm` ( removed all setup part, only run accountCmd.init() ) then it will show
each updateAccount batch would take 20~30 seconds.

src/commands/node.mjs
JeffreyDallas: Not seeing any difference in argv dumped values.

JeffreyDallas:
no longer needed

PR #402, opened 2024-06-28
fix: Resolve various problems to enable tests from local-node and smart-contracts to run agains solo network

Ivo-Yankov:
## Description

- Fixes an issue with the mirror-node-web3 module in which all calls to `/api/v1/contract/call` resulted in `error: 500`. Adds a new step at the end of `solo mirror-node deploy` that executes a script that inserts 2 rows into the `public.file_data` table - the data for `fileData` and for `exchangeRates`.
- Enables `sidecars` to be processed by the importer. Because they were not enabled all contract state data was not imported into the Mirror node and all requests to `/api/v1/contract/call` returned `0x`.
- Applied all config from `bootstrap.properties` to `application.properties` to override the default configuration for post-modularization versions of Services.

### Related Issues

* Closes #


src/commands/mirror_node.mjs
```diff
@@ -45,6 +49,8 @@ export class MirrorNodeCommand extends BaseCommand {
     }
 
     valuesArg += ` --set hedera-mirror-node.enabled=true --set hedera-explorer.enabled=${deployHederaExplorer}`
+    valuesArg += ` --set hedera-mirror-node.importer.config.hedera.mirror.importer.parser.record.sidecar.enabled=true`;
```
leninmehedy, line 52: If this is a must, then we should update it in the helm chart: https://github.com/hashgraph/full-stack-testing/blob/main/charts/fullstack-deployment/values.yaml#L289C15-L289C22

src/commands/mirror_node.mjs
```diff
@@ -172,7 +178,50 @@ export class MirrorNodeCommand extends BaseCommand {
...
+                    timestamp + '000001'
+                }, ${exchangeRatesFileIdNum}, 17);`;
+                const sqlPath = path.resolve(constants.TEMP_DIR, 'importFeesAndExchangeRates.sql');
+
+                fs.writeFileSync(  sqlPath, [importFeesQuery, importExchangeRatesQuery].join(`\n`) );
+
+                const scriptName = 'insert-fee-data.sh'
+                const scriptPath = path.join(constants.RESOURCES_DIR, scriptName)
+
+                exec(`sh ${scriptPath}`, (error, stdout, stderr) => {
```
leninmehedy, line 209: Instead of running the bash script, we need to leverage the `K8` class that uses the Kubernetes API: https://github.com/hashgraph/solo/blob/main/src/core/k8.mjs#L606

resources/templates/application.properties
```diff
@@ -1,2 +1,9 @@
-autoRenew.targetTypes=
-hedera.config.version=0
+ledger.id=0x01
```
leninmehedy, line 1: Can you please confirm the `ledger.id` param here. Currently `solo` uses ledger_id 298: https://github.com/hashgraph/solo/blob/df04772b846e8af62090ac67e829a8bc56131ed7/src/core/constants.mjs#L37

leninmehedy: This is same as `0x12a`: https://github.com/hashgraph/hedera-json-rpc-relay?tab=readme-ov-file#configuration

leninmehedy, line 1: Please confirm the values of `ledger.id` and `contracts.chainId`

Ivo-Yankov: Those are the values from `bootstrap.properties`. From my testing when running a version of the Network Node with the modularization enabled, only the values from `application.properties` are used.

I was running into an issue with specifically `contracts.chainId`, because it was using some other default value and not `298`, which should be used local networks

leninmehedy: yes, I recently came to know that `bootstrap.properties` file is no longer used.

src/commands/mirror_node.mjs
```diff
@@ -206,7 +207,53 @@ export class MirrorNodeCommand extends BaseCommand {
...
+
+                const fees = await getFileContents(this.accountManager, namespace, feesFileIdNum);
+                const exchangeRates = await getFileContents(this.accountManager, namespace, exchangeRatesFileIdNum);
+
+                const importFeesQuery = `INSERT INTO public.file_data(file_data, consensus_timestamp, entity_id, transaction_type) VALUES (decode('${fees}', 'hex'), ${timestamp + '000000'}, ${feesFileIdNum}, 17);`
+                const importExchangeRatesQuery = `INSERT INTO public.file_data(file_data, consensus_timestamp, entity_id, transaction_type) VALUES (decode('${exchangeRates}', 'hex'), ${
+                    timestamp + '000001'
+                }, ${exchangeRatesFileIdNum}, 17);`;
+                const sqlQuery = [importFeesQuery, importExchangeRatesQuery].join(`\n`);
+                const postgresPodName = 'fullstack-deployment-postgres-postgresql-0';
```
leninmehedy, line 232: Instead of hardcoding the pod name, you could use something like below:
```
          const pods = await this.k8.getPodsByLabel(['app.kubernetes.io/name=postgres'])
          if (pods.length !== 1) {
            throw new FullstackTestingError("postgres pod not found")
          }
          const postgresPodName = pods[0].name
```

JeffreyDallas (approved):

JeffreyDallas (approved):

PR #500, opened 2024-08-20
feat: add node delete and update command

### LINKED ISSUE #473: Add `solo node update` subcommand
- add a `solo node update` subcommand, much of the code can be reused from `solo node add`, you will use the `NodeUpdateTransaction`
- use DRY principle
- flags should be added to support setting:
  - gossipCaCertificate
  - certificateHash
  - adminKey
- depends on #381

### LINKED ISSUE #474: Add `solo node delete` subcommand
- add a `solo node delete` subcommand, much of the code can be reused from `solo node add`, you will use the NodeDeleteTransaction
- use DRY principle
- stop the Hedera service, but leave the pod running
- depends on https://github.com/hashgraph/solo/issues/381

JeffreyDallas:
## Description

* added node delete and update command
* added accordingly E2E tests 

### Related Issues

* Closes #474
* Closes #473


src/commands/node.mjs
```diff
@@ -2314,8 +2352,829 @@ export class NodeCommand extends BaseCommand {
...
+          const treasuryAccountPrivateKey = treasuryAccount.privateKey
+          config.treasuryKey = PrivateKey.fromStringED25519(treasuryAccountPrivateKey)
+
+          self.logger.debug('Initialized config', { config })
+        }
+      },
+      {
+        title: 'Identify existing network nodes',
+        task: async (ctx, task) => {
+          const config = /** @type {NodeAddConfigClass} **/ ctx.config
```
jeromy-cannon, line 2493:
```suggestion
          const config = /** @type {NodeUpdateConfigClass} **/ ctx.config
```

src/commands/node.mjs
```diff
@@ -2314,8 +2352,829 @@ export class NodeCommand extends BaseCommand {
...
+            config.existingNodeIds.push(networkNodeServices.nodeName)
+          }
+
+          return self.taskCheckNetworkNodePods(ctx, task, config.existingNodeIds)
+        }
+      },
+      {
+        title: 'Determine new node account number',
+        task: (ctx, task) => {
+          const config = /** @type {NodeAddConfigClass} **/ ctx.config
```
jeromy-cannon, line 2506:
```suggestion
          const config = /** @type {NodeUpdateConfigClass} **/ ctx.config
```

src/commands/node.mjs
```diff
@@ -2314,8 +2352,829 @@ export class NodeCommand extends BaseCommand {
...
+          ctx.newNode = {
+            accountId: `${constants.HEDERA_NODE_ACCOUNT_ID_START.realm}.${constants.HEDERA_NODE_ACCOUNT_ID_START.shard}.${++maxNum}`,
+            name: config.nodeId
+          }
+        }
+      },
+      {
+        title: 'Load signing key certificate',
+        task: async (ctx, task) => {
+          const config = /** @type {NodeAddConfigClass} **/ ctx.config
```
jeromy-cannon, line 2530:
```suggestion
          const config = /** @type {NodeUpdateConfigClass} **/ ctx.config
```

src/commands/node.mjs
```diff
@@ -2314,8 +2352,829 @@ export class NodeCommand extends BaseCommand {
...
+          const treasuryAccountPrivateKey = treasuryAccount.privateKey
+          config.treasuryKey = PrivateKey.fromStringED25519(treasuryAccountPrivateKey)
+
+          self.logger.debug('Initialized config', { config })
+        }
+      },
+      {
+        title: 'Identify existing network nodes',
+        task: async (ctx, task) => {
+          const config = /** @type {NodeAddConfigClass} **/ ctx.config
```
jeromy-cannon, line 2916:
```suggestion
          const config = /** @type {NodeDeleteConfigClass} **/ ctx.config
```

JeffreyDallas: Updated

src/commands/node.mjs
```diff
@@ -2314,8 +2352,829 @@ export class NodeCommand extends BaseCommand {
...
+          if (!decodedDers || decodedDers.length === 0) {
+            throw new FullstackTestingError('unable to decode public key: ' + signingCertFile)
+          }
+          ctx.signingCertDer = new Uint8Array(decodedDers[0])
+        }
+      },
+      {
+        title: 'Compute mTLS certificate hash',
+        task: async (ctx, task) => {
+          const config = /** @type {NodeAddConfigClass} **/ ctx.config
```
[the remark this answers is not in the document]
JeffreyDallas: Updated

src/commands/node.mjs
JeffreyDallas: Updated

src/commands/node.mjs
JeffreyDallas: Updated

src/commands/node.mjs
JeffreyDallas: Updated

src/commands/node.mjs
```diff
@@ -2314,8 +2352,829 @@ export class NodeCommand extends BaseCommand {
...
+            config.namespace)
+          for (/** @type {NetworkNodeServices} **/ const networkNodeServices of config.serviceMap.values()) {
+            config.existingNodeIds.push(networkNodeServices.nodeName)
+          }
+
+          return self.taskCheckNetworkNodePods(ctx, task, config.existingNodeIds)
+        }
+      },
+      {
+        title: 'Determine new node account number',
```
jeromy-cannon, line 2504: I don't see a reason to change the node account number to the next in sequence.  I don't think this would ever happen in the real world, and the old number would now create a gap.

src/commands/node.mjs
```diff
@@ -2314,8 +2352,829 @@ export class NodeCommand extends BaseCommand {
...
+          for (/** @type {NetworkNodeServices} **/ const networkNodeServices of config.serviceMap.values()) {
+            config.existingNodeIds.push(networkNodeServices.nodeName)
+          }
+
+          return self.taskCheckNetworkNodePods(ctx, task, config.existingNodeIds)
+        }
+      },
+
+      {
+        title: 'Prepare gossip endpoints',
```
jeromy-cannon, line 2912: I don't think this is needed for node delete

src/commands/node.mjs
```diff
@@ -2314,8 +2352,829 @@ export class NodeCommand extends BaseCommand {
...
+            ]
+          } else {
+            endpoints = helpers.splitFlagInput(config.gossipEndpoints)
+          }
+
+          ctx.gossipEndpoints = this.prepareEndpoints(config.endpointType, endpoints, constants.HEDERA_NODE_INTERNAL_GOSSIP_PORT)
+        }
+      },
+      {
+        title: 'Prepare grpc service endpoints',
```
jeromy-cannon, line 2933: I don't think this is needed for node delete

test/e2e/commands/node-delete.test.mjs
```diff
@@ -0,0 +1,67 @@
...
+
+  it('should delete a new node to the network successfully', async () => {
+    await nodeCmd.delete(argv)
+    expect(nodeCmd.getUnusedConfigs(NodeCommand.DELETE_CONFIGS_NAME)).toEqual([
+      flags.app.constName,
+      flags.devMode.constName
+    ])
+    await nodeCmd.accountManager.close()
+  }, 600000)
+})
```
jeromy-cannon, line 67: maybe use serviceMap or some other code to pull a list of current nodes and validate that the node was actually deleted?

src/commands/node.mjs
JeffreyDallas: This is needed to get a bew account id used in NodeUpdateTransaction,
for a three node network, 
changing 0.0.3 to 0.0.6, I think I will change to use a flag parameter to get a new account number.

Meanwhile, what could be problem if account number has a gap ? ( not a node id gap)

test/e2e/commands/node-update.test.mjs
```diff
@@ -0,0 +1,94 @@
...
+  it('cache current version of private keys', async () => {
+    existingServiceMap = await nodeCmd.accountManager.getNodeServiceMap(namespace)
+    existingNodeIdsPrivateKeysHash = await getNodeIdsPrivateKeysHash(existingServiceMap, namespace, constants.KEY_FORMAT_PEM, k8, getTmpDir())
+  }, defaultTimeout)
+
+  it('should succeed with init command', async () => {
+    const status = await accountCmd.init(argv)
+    expect(status).toBeTruthy()
+  }, 450000)
+
```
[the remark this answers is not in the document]
JeffreyDallas: do we need to change keys ? since keys are generated based on node name ,which did not change.

We started with

node1, 0.0.3
node2, 0.0.4
node3, 0.0.5

after update, it changed to

node1, 0.0.6
node2, 0.0.4
node3, 0.0.5

node name still the same .

jeromy-cannon (changes requested):
see other feedback

src/commands/node.mjs
jeromy-cannon: I talked with Nathan, this is valid.  but, we should make a flag to decide whether or not to change the account number.

test/e2e/commands/node-update.test.mjs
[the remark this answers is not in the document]
jeromy-cannon: they are optional to change.  if they were afraid their private keys were compromised they might want to change them.  we pull the grpcTLS and gossip key from the solo cache keys directory.  so, you could just write an updated one to that directory for the test, your code in node update should be fine.

Also, we should add a flag to pass in the node admin key, in case we want to change that as well.

test/e2e/commands/node-delete.test.mjs
JeffreyDallas: Updated

src/commands/node.mjs
JeffreyDallas: Updated

src/commands/node.mjs
JeffreyDallas: Updated

test/e2e/commands/node-update.test.mjs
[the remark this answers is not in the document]
JeffreyDallas: Updated

src/commands/node.mjs
JeffreyDallas: Updated

src/commands/node.mjs
```diff
@@ -154,6 +156,46 @@ export class NodeCommand extends BaseCommand {
...
+      flags.newAdminKey,
+      flags.newCACertificate,
+      flags.newGRPCHash,
+      flags.devMode,
+      flags.endpointType,
+      flags.gossipEndpoints,
+      flags.grpcEndpoints,
+      flags.namespace,
+      flags.nodeID,
+      flags.releaseTag
```
jeromy-cannon, line 195: please sort alphabetical

src/commands/node.mjs
```diff
@@ -2314,8 +2355,766 @@ export class NodeCommand extends BaseCommand {
...
+        title: 'Initialize',
+        task: async (ctx, task) => {
+          self.configManager.update(argv)
+
+          // disable the prompts that we don't want to prompt the user for
+          prompts.disablePrompts([
+            flags.app,
+            flags.chartDirectory,
+            flags.devMode,
+            flags.localBuildPath,
```
jeromy-cannon, line 2411: remove?  flags.localBuildPath, isn't used?

src/commands/node.mjs
```diff
@@ -2314,8 +2355,766 @@ export class NodeCommand extends BaseCommand {
...
+           * @property {string[]} allNodeIds
+           * @property {string[]} existingNodeIds
+           * @property {string} freezeAdminPrivateKey
+           * @property {string} keysDir
+           * @property {Object} nodeClient
+           * @property {Object} podNames
+           * @property {Map<String, NetworkNodeServices>} serviceMap
+           * @property {PrivateKey} treasuryKey
+           * @property {string} stagingDir
+           * @property {string} stagingKeysDir
```
jeromy-cannon, line 2452: please sort alphabetical

src/commands/node.mjs
```diff
@@ -2314,8 +2355,766 @@ export class NodeCommand extends BaseCommand {
...
+
+          self.logger.info(`ctx.signingCertDer: ${ctx.signingCertDer}`)
+          self.logger.info(`ctx.tlsCertHash: ${ctx.tlsCertHash}`)
+          try {
+            const nodeUpdateTx = await new NodeUpdateTransaction()
+              .setNodeId(nodeId)
+              .setGossipEndpoints(ctx.gossipEndpoints)
+              .setServiceEndpoints(ctx.grpcServiceEndpoints)
+              .setGossipCaCertificate(ctx.signingCertDer)
+              .setCertificateHash(ctx.tlsCertHash)
```
jeromy-cannon, line 2628: why set it and then overlay it.  it would be better to only set it once with whatever the correct value is.

src/commands/node.mjs
```diff
@@ -2314,8 +2355,766 @@ export class NodeCommand extends BaseCommand {
...
+          self.logger.info(`config.newAccountNumber: ${config.newAccountNumber}`)
+
+          self.logger.info(`ctx.signingCertDer: ${ctx.signingCertDer}`)
+          self.logger.info(`ctx.tlsCertHash: ${ctx.tlsCertHash}`)
+          try {
+            const nodeUpdateTx = await new NodeUpdateTransaction()
+              .setNodeId(nodeId)
+              .setGossipEndpoints(ctx.gossipEndpoints)
+              .setServiceEndpoints(ctx.grpcServiceEndpoints)
+              .setGossipCaCertificate(ctx.signingCertDer)
```
jeromy-cannon, line 2627: why set it and then overlay it. it would be better to only set it once with whatever the correct value is.

src/commands/node.mjs
```diff
@@ -2314,8 +2355,766 @@ export class NodeCommand extends BaseCommand {
...
+           * @property {string[]} allNodeIds
+           * @property {string[]} existingNodeIds
+           * @property {string} freezeAdminPrivateKey
+           * @property {string} keysDir
+           * @property {Object} nodeClient
+           * @property {Object} podNames
+           * @property {Map<String, NetworkNodeServices>} serviceMap
+           * @property {PrivateKey} treasuryKey
+           * @property {string} stagingDir
+           * @property {string} stagingKeysDir
```
jeromy-cannon, line 2863: please sort alphabetical

test/e2e/commands/node-update.test.mjs
```diff
@@ -0,0 +1,101 @@
...
+  accountCreationShouldSucceed(nodeCmd.accountManager, nodeCmd, namespace)
+
+  it('existing nodes private keys should not have changed', async () => {
+    const currentNodeIdsPrivateKeysHash = await getNodeIdsPrivateKeysHash(existingServiceMap, namespace, constants.KEY_FORMAT_PEM, k8, getTmpDir())
+
+    for (const [nodeId, existingKeyHashMap] of existingNodeIdsPrivateKeysHash.entries()) {
+      const currentNodeKeyHashMap = currentNodeIdsPrivateKeysHash.get(nodeId)
+
+      for (const [keyFileName, existingKeyHash] of existingKeyHashMap.entries()) {
+        expect(`${nodeId}:${keyFileName}:${currentNodeKeyHashMap.get(keyFileName)}`).toEqual(
```
jeromy-cannon, line 96: shouldn't they not match for node3?  It looks like you updated the gossip and grpcTLS keys.

JeffreyDallas: should hedera app generate new keys after handling nodeUpdateTransaction
it seems only regenerated public key

```
ls -ltr /opt/hgcapp/services-hedera/HapiApp2.0/data/upgrade/current/
total 32
drwxr-xr-x 3 hedera hedera 4096 Aug 28 03:37 data
-rwxr-xr-x 1 hedera hedera 1464 Aug 28 03:37 s-public-node3.pem
-rwxr-xr-x 1 hedera hedera 1464 Aug 28 03:37 s-public-node2.pem
-rwxr-xr-x 1 hedera hedera 1464 Aug 28 03:37 s-public-node1.pem
-rwxr-xr-x 1 hedera hedera    3 Aug 28 03:37 execute_immediate.mf
-rwxr-xr-x 1 hedera hedera  484 Aug 28 03:37 config.txt
-rwxr-xr-x 1 hedera hedera   10 Aug 28 03:37 freeze_scheduled.mf
-rwxr-xr-x 1 hedera hedera    3 Aug 28 03:37 now_frozen.mf
```

jeromy-cannon: I don't think it is generating new keys.  There is a download task that will copy down the public keys from nodeId[0] from data/upgrade/current along with the config.txt, then update the configmap, which you have to killl/start all the pods to pick up the change.

JeffreyDallas: The operation of download files from dat/upgrade/current and update the configmap are already there.
but the helper function getNodeIdsPrivateKeysHash only get hashes of private key files,
since private key files are untouched, of course the check would pass.

We could manually copy the test key pem file to the key directory, of course it will guarantee the mismatch
but seems an unnecessary move, 

maybe we could just skip the check step.

src/commands/flags.mjs
```diff
@@ -538,6 +538,50 @@ export const localBuildPath = {
...
+
+/** @type {CommandFlag} **/
+export const newCACertificate = {
+  constName: 'newCACertificate',
+  name: 'new-ca-certificate',
+  definition: {
+    describe: 'file name of the new CA certificate for gossip',
+    defaultValue: '',
+    type: 'string'
+  }
```
jeromy-cannon, line 571:
```suggestion
export const gossipKey = {
  constName: 'gossipKey',
  name: 'gossip-key',
  definition: {
    describe: 'path and file name of the gossip key in PEM key format to be used',
    defaultValue: '',
    type: 'string'
  }
```

src/commands/flags.mjs
```diff
@@ -538,6 +538,50 @@ export const localBuildPath = {
...
+
+/** @type {CommandFlag} **/
+export const newGRPCHash = {
+  constName: 'newGRPCHash',
+  name: 'new-grpc-hash',
+  definition: {
+    describe: 'file name of the new grpc certificate',
+    defaultValue: '',
+    type: 'string'
+  }
```
jeromy-cannon, line 582:
```suggestion
export const tlsKey = {
  constName: 'tlsKey',
  name: 'tls-key',
  definition: {
    describe: 'path and file name of the private TLS key to be used',
    defaultValue: '',
    type: 'string'
  }
```

src/commands/node.mjs
```diff
@@ -2316,8 +2362,1051 @@ export class NodeCommand extends BaseCommand {
...
+          }
+
+          return self.taskCheckNetworkNodePods(ctx, task, config.existingNodeIds)
+        }
+      },
+      {
+        title: 'Load signing key certificate',
+        task: async (ctx, task) => {
+          const config = /** @type {NodeUpdateConfigClass} **/ ctx.config
+          const signingCertFile = Templates.renderGossipPemPublicKeyFile(constants.SIGNING_KEY_PREFIX, config.nodeId)
```
jeromy-cannon, line 2527: if the flag is set for passing in the gossip signing key, generate the public key from the private key, and copy them into the cache/keys dir.  then load that.

src/commands/node.mjs
```diff
@@ -2316,8 +2362,1051 @@ export class NodeCommand extends BaseCommand {
...
+          const signingCertFullPath = path.join(config.keysDir, signingCertFile)
+          const signingCertPem = fs.readFileSync(signingCertFullPath).toString()
+          const decodedDers = x509.PemConverter.decode(signingCertPem)
+          if (!decodedDers || decodedDers.length === 0) {
+            throw new FullstackTestingError('unable to decode public key: ' + signingCertFile)
+          }
+          ctx.signingCertDer = new Uint8Array(decodedDers[0])
```
jeromy-cannon, line 2534: can you use your new method: loadPermCertificate()?

src/commands/node.mjs
```diff
@@ -2316,8 +2362,1051 @@ export class NodeCommand extends BaseCommand {
...
+          const config = /** @type {NodeUpdateConfigClass} **/ ctx.config
+          const tlsCertFile = Templates.renderTLSPemPublicKeyFile(config.nodeId)
+          const tlsCertFullPath = path.join(config.keysDir, tlsCertFile)
+          const tlsCertPem = fs.readFileSync(tlsCertFullPath).toString()
+          const tlsCertDers = x509.PemConverter.decode(tlsCertPem)
+          if (!tlsCertDers || tlsCertDers.length === 0) {
+            throw new FullstackTestingError('unable to decode tls cert: ' + tlsCertFullPath)
+          }
+          const tlsCertDer = new Uint8Array(tlsCertDers[0])
+          ctx.tlsCertHash = crypto.createHash('sha384').update(tlsCertDer).digest()
```
jeromy-cannon, line 2549: can you use your new method: loadPermCertificate()?

src/commands/node.mjs
```diff
@@ -2316,8 +2362,1051 @@ export class NodeCommand extends BaseCommand {
...
+            throw new FullstackTestingError('unable to decode public key: ' + signingCertFile)
+          }
+          ctx.signingCertDer = new Uint8Array(decodedDers[0])
+        }
+      },
+      {
+        title: 'Compute mTLS certificate hash',
+        task: async (ctx, task) => {
+          const config = /** @type {NodeUpdateConfigClass} **/ ctx.config
+          const tlsCertFile = Templates.renderTLSPemPublicKeyFile(config.nodeId)
```
jeromy-cannon, line 2541: if the flag is set for passing in the tls key, generate the public key from the private key, and copy them into the cache/keys dir. then load that.

src/commands/node.mjs
```diff
@@ -2316,8 +2362,1051 @@ export class NodeCommand extends BaseCommand {
...
+            // set up the sub-tasks
+            return task.newListr(subTasks, {
+              concurrent: false, // no need to run concurrently since if one node is up, the rest should be up by then
+              rendererOptions: {
+                collapseSubtasks: false
+              }
+            })
+          }
+      },
+      {
```
jeromy-cannon, line 2803: prepare staging directory: https://github.com/hashgraph/solo/blob/681d5b63a94324e73aace70d7d4190a4535a129d/src/commands/node.mjs#L1921-L1921 to copy the keys from the cache/keys to the staging directory

src/commands/node.mjs
```diff
@@ -2316,8 +2362,1051 @@ export class NodeCommand extends BaseCommand {
...
+        task: async (ctx, parentTask) => {
+          const config = /** @type {NodeUpdateConfigClass} **/ ctx.config
+
+          const subTasks = []
+          for (const nodeId of config.allNodeIds) {
+            const podName = config.podNames[nodeId]
+            subTasks.push({
+              title: `Node: ${chalk.yellow(nodeId)}`,
+              task: () =>
+                self.platformInstaller.taskInstall(podName, config.stagingDir, config.allNodeIds, config.keyFormat, config.force)
```
jeromy-cannon, line 2825: this will copy the keys into the correct directories.  with the current code version, you are copying in the old keys.  I'm surprised that the NodeUpdateTransaction has a new hash for the TLS public key, and a new public signing gossip PEM key, but it the system is still working while the node is still using the old key files?

test/e2e/commands/node-update.test.mjs
jeromy-cannon: the private and public keys for the TLS and the signing gossip should be updated, so it should not match.  I made notes in other places, as to why I think the code that uploads them are missing.

I'm actually wondering if there is a bug as to why services and platform is working with node2. having the old keys.  or perhaps it is throwing an error and we just aren't looking at the logs?

src/commands/node.mjs
```diff
@@ -2627,21 +2598,61 @@ export class NodeCommand extends BaseCommand {
...
-          self.logger.info(`ctx.tlsCertHash: ${ctx.tlsCertHash}`)
           try {
             const nodeUpdateTx = await new NodeUpdateTransaction()
               .setNodeId(nodeId)
 
-            if (config.newGRPCHash) {
-              const tlsCertDer = await this.loadPermCertificate(config.newCACertificate)
+            if (config.tlsKey) {
+              const tlsCertDer = await this.loadPermCertificate(config.gossipKey)
               const tlsCertHash = crypto.createHash('sha384').update(tlsCertDer).digest()
```
jeromy-cannon, line 2607: I think the hash is supposed to be of the public key, right?

src/commands/node.mjs
```diff
@@ -2627,21 +2598,61 @@ export class NodeCommand extends BaseCommand {
...
-          self.logger.info(`ctx.signingCertDer: ${ctx.signingCertDer}`)
-          self.logger.info(`ctx.tlsCertHash: ${ctx.tlsCertHash}`)
           try {
             const nodeUpdateTx = await new NodeUpdateTransaction()
               .setNodeId(nodeId)
 
-            if (config.newGRPCHash) {
-              const tlsCertDer = await this.loadPermCertificate(config.newCACertificate)
+            if (config.tlsKey) {
+              const tlsCertDer = await this.loadPermCertificate(config.gossipKey)
```
jeromy-cannon, line 2606: did you mean to load the config.tlsKey?

src/commands/node.mjs
jeromy-cannon: I think you need to generate the public key from the private key, then take the hash of the public key.

src/commands/node.mjs
```diff
@@ -2316,8 +2362,1062 @@ export class NodeCommand extends BaseCommand {
...
+
+            // config.adminKey contains the original key, needed to sign the transaction
+            let signedTx
+            if (config.newAdminKey) {
+              await nodeUpdateTx.sign(parsedNewKey)
+              signedTx = await nodeUpdateTx.sign(config.adminKey)
+            } else {
+              signedTx = await nodeUpdateTx.sign(config.adminKey)
+            }
+            const txResp = await signedTx.execute(config.nodeClient)
```
jeromy-cannon, line 2677:
```suggestion
            if (config.newAdminKey) {
              await nodeUpdateTx.sign(parsedNewKey)
            }
            signedTx = await nodeUpdateTx.sign(config.adminKey)
            const txResp = await signedTx.execute(config.nodeClient)
```

src/commands/node.mjs
```diff
@@ -2316,8 +2362,1062 @@ export class NodeCommand extends BaseCommand {
...
+           * -- flags --
+           * @property {string} app
+           * @property {string} cacheDir
+           * @property {string} charDirectory
+           * @property {boolean} devMode
+           * @property {string} endpointType
+           * @property {string} fstChartVersion
+           * @property {string} gossipEndpoints
+           * @property {string} grpcEndpoints
+           * @property {string} keyFormat
```
jeromy-cannon, line 2998:
```suggestion
           * @property {string} keyFormat
```
I don't think we use gossipEndpoints or grpcEndpoints in node delete

JeffreyDallas: Removed

src/commands/node.mjs
JeffreyDallas: Updated

src/commands/node.mjs
```diff
@@ -2316,8 +2362,1062 @@ export class NodeCommand extends BaseCommand {
...
+                if (err) {
+                  self.logger.error(`Error copying tls private key: ${err.message}`)
+                  throw new FullstackTestingError(`Error copying tls private key: ${err.message}`, err)
+                }
+              })
+            }
+
+            if (config.gossipKey) {
+              const signingCertDer = await this.loadPermCertificate(config.gossipKey)
+              nodeUpdateTx.setGossipCaCertificate(signingCertDer)
```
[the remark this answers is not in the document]
JeffreyDallas: Upated

src/commands/node.mjs
JeffreyDallas: Upated

src/commands/node.mjs
JeffreyDallas: Updated

src/commands/node.mjs
JeffreyDallas: Updated

src/commands/node.mjs
JeffreyDallas: Updated

src/commands/node.mjs
JeffreyDallas: Updated

src/commands/node.mjs
JeffreyDallas: Updated

test/e2e/commands/node-update.test.mjs
```diff
@@ -0,0 +1,130 @@
...
+  bootstrapNetwork,
+  getDefaultArgv, getNodeIdsPrivateKeysHash, getTmpDir,
+  HEDERA_PLATFORM_VERSION_TAG
+} from '../../test_util.js'
+import { getNodeLogs } from '../../../src/core/helpers.mjs'
+import { NodeCommand } from '../../../src/commands/node.mjs'
+import { HEDERA_HAPI_PATH, ROOT_CONTAINER } from '../../../src/core/constants.mjs'
+import fs from 'fs'
+
+describe('Node add', () => {
```
jeromy-cannon, line 33:
```suggestion
describe('Node Update', () => {
```

test/e2e/commands/node-update.test.mjs
```diff
@@ -0,0 +1,130 @@
...
+    expect(status).toBeTruthy()
+  }, 450000)
+
+  it('should update a new node property successfully', async () => {
+    // generate gossip and tls keys for the updated node
+    const tmpDir = getTmpDir()
+    const signingKey = await nodeCmd.keyManager.generateSigningKey(updateNodeId)
+    const signingKeyFiles = await nodeCmd.keyManager.storeSigningKey(updateNodeId, signingKey, tmpDir)
+    nodeCmd.logger.debug(`generated test gossip signing keys for node ${updateNodeId} : ${signingKeyFiles.certificateFile}`)
+    argv[flags.gossipKey.name] = signingKeyFiles.certificateFile
```
jeromy-cannon, line 82:
```suggestion
    argv[flags.gossipKey.name] = signingKeyFiles.privateKeyFile
```

src/commands/node.mjs
jeromy-cannon: As I'm thinking through this, I think we need the user to pass in both the public and the private for the TLS and/or gossip keys.

If it is TLS, we could generate a self-signed, but one with a CA would be better and they could do that and pass it in.

similar with the Gossip PEM key.  Then we wouldn't have to generate it.

src/commands/flags.mjs
```diff
@@ -538,6 +538,50 @@ export const localBuildPath = {
...
+  name: 'new-admin-key',
+  definition: {
+    describe: 'new admin key for the Hedera account',
+    defaultValue: '',
+    type: 'string'
+  }
+}
+
+/** @type {CommandFlag} **/
+export const gossipKey = {
```
jeromy-cannon, line 564: make two:
- publicGossipKey
- privateGossipKey

src/commands/flags.mjs
```diff
@@ -538,6 +538,50 @@ export const localBuildPath = {
...
+  name: 'gossip-key',
+  definition: {
+    describe: 'path and file name of the private signing gossip key in PEM key format to be used',
+    defaultValue: '',
+    type: 'string'
+  }
+}
+
+/** @type {CommandFlag} **/
+export const tlsKey = {
```
jeromy-cannon, line 575: make two:
- privateTlsKey
- publicTlsKey

JeffreyDallas: Updated with separate flags for public and private key

src/commands/flags.mjs
JeffreyDallas: Updated with separate flags for public and private key

test/e2e/commands/node-update.test.mjs
```diff
@@ -0,0 +1,130 @@
...
+
+    const tlsKey = await nodeCmd.keyManager.generateGrpcTLSKey(updateNodeId)
+    const tlsKeyFiles = await nodeCmd.keyManager.storeTLSKey(updateNodeId, tlsKey, tmpDir)
+    nodeCmd.logger.debug(`generated test TLS keys for node ${updateNodeId} : ${tlsKeyFiles.certificateFile}`)
+    argv[flags.tlsKey.name] = tlsKeyFiles.certificateFile
```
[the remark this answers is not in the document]
JeffreyDallas: Updated with separate flags for public and private key

test/e2e/commands/node-update.test.mjs
JeffreyDallas: Updated with separate flags for public and private key

test/e2e/commands/node-update.test.mjs
JeffreyDallas: Fixed

src/commands/node.mjs
```diff
@@ -2316,8 +2364,1048 @@ export class NodeCommand extends BaseCommand {
...
+          prompts.disablePrompts([
+            flags.app,
+            flags.chainId,
+            flags.chartDirectory,
+            flags.devMode,
+            flags.endpointType,
+            flags.force,
+            flags.fstChartVersion,
+            flags.gossipEndpoints,
+            flags.grpcEndpoints,
```
jeromy-cannon, line 2976:
```suggestion
```

JeffreyDallas: Fixed

test/e2e/commands/node-update.test.mjs
```diff
@@ -0,0 +1,134 @@
...
+  it('should update a new node property successfully', async () => {
+    // generate gossip and tls keys for the updated node
+    const tmpDir = getTmpDir()
+
+    const signingKey = await nodeCmd.keyManager.generateSigningKey(updateNodeId)
+    const signingKeyFiles = await nodeCmd.keyManager.storeSigningKey(updateNodeId, signingKey, tmpDir)
+    nodeCmd.logger.debug(`generated test gossip signing keys for node ${updateNodeId} : ${signingKeyFiles.certificateFile}`)
+    // Temporarily disable gossip keys generation due to unsolved bug
+    argv[flags.gossipPublicKey.name] = signingKeyFiles.certificateFile
+    argv[flags.gossipPrivateKey.name] = signingKeyFiles.privateKeyFile
```
jeromy-cannon, line 85: did you mean to comment this out?

JeffreyDallas: I added .skip to it() block

jeromy-cannon (approved):

PR #512, opened 2024-08-23
feat: add configuration for Monitor Pinger service

### LINKED ISSUE #511: Setup mirror node monitor Pinger service
### Feature to Enhance

Currently there is an issue in the network node that causes record files to only be exported when a new one is generated. This means that in networks with sparce transactions the latest one may become "stuck". This can be prevented with setting up a Pinger service in the Mirror node Monitor pod, which periodically creates small `CRYPTO_TRANSFER` transactions. After some quick testing I found that this can double the performance of the Solo network. There are also some test suites in the [Smart Contracts repo](https://github.com/hashgraph/hedera-smart-contracts) that fail with a `TIMEOUT` errors, which should be resolved by this.

### Requested Enhancement

Implement a similar solution to the one in the [local node](https://github.com/hashgraph/hedera-local-node).

### Business Need

The user experience when running end-to-end tests against a locally setup solo network is really poor due to the slow transaction processing.

### Additional Context

_No response_

Ivo-Yankov:
## Description

- Adds dynamic configuration to the Pinger service.
- Requires https://github.com/hashgraph/full-stack-testing/pull/959 to be merged and a new release to be created

### Related Issues

* Closes #511

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ Then run the following command to set the kubectl context to the new cluster:
 ```bash
 kind create cluster -n "${SOLO_CLUSTER_NAME}"
 ```
+
 Example output
 
 ```
@@ -184,6 +185,7 @@ Kubernetes Namespace	: solo
 ✔ Generate gRPC TLS keys
 ✔ Finalize
 ```
+
 Key files are generated in `~/.solo/keys` directory.
 
 ```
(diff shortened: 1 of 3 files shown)


JeffreyDallas:
Maybe add a flag so user has an option to turn on or off pinger service,
my experience is there are many background transactions from pingers, sometimes had to check log when debugging other issues.

jeromy-cannon:
pick back up later

PR #547, opened 2024-09-05
refactor(liveness check): switch the network node pod check strategy from: 'wait phase running'…

### LINKED ISSUE #209: Update checks for Network Node ACTIVE to leverage PodCondition[Ready]
Update Solo's logic for checking if the Network Node is ACTIVE to leverage the PodCondition[Ready] status.

- [ ] the prometheus endpoint might need to be enabled and configured with a port depending on what the defaults are
- [ ] https://github.com/hashgraph/hedera-services/blob/3747e61503c22d5424d7cd3ff3d0d04c5909f5f7/platform-sdk/swirlds-platform-core/src/main/java/com/swirlds/platform/config/internal/ConfigMappings.java#L59-L59
- [ ] I think they are stored in settings.txt
- [ ] the settings are probably already enabled by default
- [ ] we need to open a port forward on the port 9090, then hit the prometheus metric endpoint, check the results.  It seems we are looking for the state, Nathan mentioned enums, we mostly use ACTIVE, but also sometimes FREEZE_COMPLETE

instamenta:
## Description

Updated the network node pod checks to leverage the 'pod condition ready' instead of just the 'pod phase running'.
Refactored the code for the method that does this check to improve readability 

closes #209

### DIFF
--- a/src/commands/node.mjs
+++ b/src/commands/node.mjs
@@ -303,7 +303,7 @@ export class NodeCommand extends BaseCommand {
     const podName = Templates.renderNetworkPodName(nodeId)
 
     try {
-      await this.k8.waitForPods([constants.POD_PHASE_RUNNING], [
+      await this.k8.waitForPodReady([
         'fullstack.hedera.com/type=network-node',
         `fullstack.hedera.com/node-name=${nodeId}`
       ], 1, maxAttempts, delay)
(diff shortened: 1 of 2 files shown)


jeromy-cannon (changes requested):
reviewing.

what is the metrics port returning when doing the liveness probe?  I think we wanted to wait until it got something from the metrics port that said that the network node was ACTIVE (during `solo node start`, when the network becomes usable).  This was so that we could get away from parsing the log in `checkNetworkNodeState`.  

I don't think that is what we accomplished.  I might have approved your FST chart changes in https://github.com/hashgraph/full-stack-testing/pull/962 too quickly.

PR #549, opened 2024-09-06
feat: add support for attach jvm debugger

### LINKED ISSUE #298: Add ability to set the debugger on a network node JVM
I believe this was mentioned it DevCon 2024.

We did this on the fly by doing the following: https://github.com/hashgraph/solo/pull/469/commits/5057474e7dda57d19d7cee3e7d3df23bf338508f

image

We also made a similar alteration during the helm upgrade in `solo node add`.

This is extremely useful to help debug issues encountered inside the platform or services code.

----

While trying to determine root cause as to why platform sdk was throwing an incorrect signature after a node add we needed to add a debugger to the JVM.  There was not an easy way to do this since current code dynamically generates the hedera.nodes array for the values file, and node add needs to also dynamically generate this.  We hardcoded it, but it would be good to have a better more long term solution.

Some possible features:
* be able to set which node the debugger will attach to, if this is too difficult, then just hard code it to the first one
* increase timeouts, the node will wait for the debugger to attach, and then the debug session will take time, current code would timeout during check for node ACTIVE
* work for both `node start` and `node add` (as well as future `node update` and `node delete`)

see the commented out code below as to how it was solved for a temporary solution.

`src/core/profile_manager.mjs`
```
  resourcesForConsensusPod (profile, nodeIds, yamlRoot) {
    if (!profile) throw new MissingArgumentError('profile is required')
    const accountMap = getNodeAccountMap(nodeIds)
    // set consensus pod level resources
    for (let nodeIndex = 0; nodeIndex < nodeIds.length; nodeIndex++) {
      this._setValue(`hedera.nodes.${nodeIndex}.name`, nodeIds[nodeIndex], yamlRoot)
      this._setValue(`hedera.nodes.${nodeIndex}.accountId`, accountMap.get(nodeIds[nodeIndex]), yamlRoot)
      this._setChartItems(`hedera.nodes.${nodeIndex}`, profile.consensus, yamlRoot)
      // if (nodeIndex === 0) {
      //   this._setValue(`hedera.nodes.${nodeIndex}.root.extraEnv.0.name`, 'JAVA_OPTS', yamlRoot)
      //   this._setValue(`hedera.nodes.${nodeIndex}.root.extraEnv.0.value`, '-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005', yamlRoot)
      // }
    }
```

`src/commands/node.mjs`
```
      {
        title: 'Deploy new network node',
        task: async (ctx, task) => {
          const config = /** @type {NodeAddConfigClass} **/ ctx.config
          const index = config.existingNodeIds.length
          let valuesArg = ''
          for (let i = 0; i < index; i++) {
            valuesArg += ` --set "hedera.nodes[${i}].accountId=${config.serviceMap.get(config.existingNodeIds[i]).accountId}" --set "hedera.nodes[${i}].name=${config.existingNodeIds[i]}"`
            // if (i === 0) {
            //   valuesArg += ` --set "hedera.nodes[${i}].root.extraEnv[0].name=JAVA_OPTS"`
            //   valuesArg += ` --set "hedera.nodes[${i}].root.extraEnv[0].value=-agentlib:jdwp=transport=dt_socket\\,server=y\\,suspend=y\\,address=*:5005"`
            // }
          }
          valuesArg += ` --set "hedera.nodes[${index}].accountId=${ctx.newNode.accountId}" --set "hedera.nodes[${index}].name=${ctx.newNode.name}"`

          await self.chartManager.upgrade(
            config.namespace,
            constants.FULLSTACK_DEPLOYMENT_CHART,
            config.chartPath,
            valuesArg,
            config.fstChartVersion
          )

          config.allNodeIds = [...config.existingNodeIds, config.nodeId]
        }
      },
```

JeffreyDallas:
## Description

* add support for attach jvm debugger

### Related Issues

* Closes #298 

Instructions

https://www.notion.so/swirldslabs/Using-solo-with-JVM-Debugger-70aea26ce00341ed89d31c931e1ca02b


README.md.template
```diff
@@ -377,6 +377,55 @@ You can find log for running solo command under the directory `~/.solo/logs/`
 The file `solo.log` contains the logs for the solo command. 
 The file `hashgraph-sdk.log` contains the logs from solo client when sending transactions to network nodes.
 
+## Using Intellj remote debug with solo
+
```
jeromy-cannon, line 381:
```suggestion

NOTE: the hedera-services path referenced '../hedera-services/hedera-node/data' may need to be updated based on what directory you are currently in.  This also assumes that you have done an assemble/build and the directory contents are up-to-date.

```

jeromy-cannon (dismissed):

jeromy-cannon (approved):

Issue #503, opened 2024-08-21
In the README maintain a compatibility matrix for the version of Solo with the version of Hedera, and other technologies.


jeromy-cannon:
- [ ] required by: https://github.com/hashgraph/solo/issues/57

PR #555, opened 2024-09-10
docs: add the requirement table for the current version of solo

instamenta:
## Description

Add the requirement matrix table to the readme with all the requirements for the current version of solo

## Closes 

https://github.com/hashgraph/solo/issues/503

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,9 @@ An opinionated CLI tool to deploy and manage standalone test networks.
 
 ## Requirements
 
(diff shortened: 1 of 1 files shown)


jeromy-cannon:
@instamenta , would you please add the GH Issue number into the description?  something like this: 
closes #503 

Also, would you add the Hedera Platform version to the matrix.  I think we can use v0.53.0+ as the version for the current row you put in there.

PR #567, opened 2024-09-12
feat(network node livecheck): implement a functionality to determine network node activeness with prometheus metrics

instamenta:
## Description

Implemented functionality that utilizes the node root container prometheus metrics endpoint to determine node's activeness

### Related Issues

* Closes # [#209](https://github.com/hashgraph/solo/issues/209)


src/commands/node.mjs
```diff
@@ -385,6 +387,119 @@ export class NodeCommand extends BaseCommand {
...
+
+      try {
+        const url = `http://${LOCAL_HOST}:${localPort}/metrics`
+        console.log(url)
+        const response = await fetch(url, { signal: controller.signal })
+
+        if (!response.ok) {
+          task.title = `${title} - status ${chalk.yellow('UNKNOWN')}, attempt ${chalk.blueBright(`${attempt}/${maxAttempts}`)}`
+          clearTimeout(timeoutId)
+          throw new Error()
```
jeromy-cannon, line 430: here you left the port forward open without closing it

src/commands/node.mjs
```diff
@@ -385,6 +387,119 @@ export class NodeCommand extends BaseCommand {
...
+
+        const text = await response.text()
+        const statusLine = text
+          .split('\n')
+          .find(line => line.startsWith('platform_PlatformStatus'))
+
+        if (!statusLine) {
+          task.title = `${title} - status ${chalk.yellow('STARTING')}, attempt: ${chalk.blueBright(`${attempt}/${maxAttempts}`)}`
+          clearTimeout(timeoutId)
+          throw new Error()
```
jeromy-cannon, line 441: here you left the port forward open without closing it

src/commands/node.mjs
```diff
@@ -385,6 +387,119 @@ export class NodeCommand extends BaseCommand {
...
+        }
+        clearTimeout(timeoutId)
+      } catch {}
+
+      attempt++
+      clearTimeout(timeoutId)
+      await sleep(delay)
+    }
+
+    await this.k8.stopPortForward(srv)
```
jeromy-cannon, line 465: if anything happens above that would exit the method or throw an exception, then you would leave the port forward open

src/commands/node.mjs
instamenta: its in a try catch and a while cycle it exits after the while loop is broken

src/commands/node.mjs
instamenta: its in a try catch

instamenta: > here you left the port forward open without closing it

its in a try catch

src/commands/node.mjs
instamenta: > here you left the port forward open without closing it

its in a try catch and a while cycle it exits after the while loop is broken

instamenta: it's in a try catch and a while loop, it exits after the while loop is broken

src/commands/node.mjs
instamenta: here you left the port forward open without closing it

its in a try catch

src/commands/node.mjs
instamenta: the only thing that can prevent this from executing is the 'this.k8.portForward(podName, localPort, podPort)' which would mean the port is not open, the other code is inside a try catch

src/commands/node.mjs
jeromy-cannon: instead of throwing a new Error why not just use continue, I think continue would be more clear for what you are using it for.

src/commands/node.mjs
jeromy-cannon: instead of throwing a new Error why not just use continue, I think continue would be more clear for what you are using it for.

src/commands/node.mjs
instamenta: there is code after the catch that won't execute if 'continue' is used

jeromy-cannon (approved):

PR #583, opened 2024-09-16
chore: Relocate flags used for mirror node deployment

### LINKED ISSUE #431: Move mirror node options from `solo network deploy` to `solo mirror-node deploy`
The following options are still loaded under `solo network deploy`; however, they aren't actually used until `solo mirror-node deploy` and should be moved there.

```
      flags.deployHederaExplorer,
      flags.deployMirrorNode,
      flags.enableHederaExplorerTls,
      flags.hederaExplorerTlsHostName,
      flags.hederaExplorerTlsLoadBalancerIp,
      flags.tlsClusterIssuerType,
```

should be done after https://github.com/hashgraph/solo/issues/136 to avoid conflicts from refactoring

JeffreyDallas:
## Description

* Relocate flags used for mirror node deployment

### Related Issues

* Closes #431

### DIFF
--- a/src/core/k8.mjs
+++ b/src/core/k8.mjs
@@ -343,8 +343,7 @@ export class K8 {
    * @param {string} containerName
    * @param {string} destPath - path inside the container
    * @param {number} [timeout] - timeout in ms
-   * @returns {Promise<{owner: string, size: number, modifiedAt: string, name: string, directory: boolean, group: string}[]>}
-   * array of directory entries, custom object
+   * @returns array of directory entries, custom object
    */
   async listDir (podName, containerName, destPath, timeout = 5000) {
     try {
--- a/src/core/k8.mjs
+++ b/src/core/k8.mjs
@@ -343,7 +343,7 @@ export class K8 {
    * @param {string} containerName
    * @param {string} destPath - path inside the container
    * @param {number} [timeout] - timeout in ms
-   * @returns array of directory entries, custom object
+   * @returns a promise that returns array of directory entries, custom object
    */
   async listDir (podName, containerName, destPath, timeout = 5000) {
     try {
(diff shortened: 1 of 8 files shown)


test/e2e/commands/network.test.mjs
```diff
@@ -103,15 +103,10 @@ describe('NetworkCommand', () => {
         flags.applicationProperties.constName,
         flags.bootstrapProperties.constName,
         flags.chainId.constName,
-        flags.deployHederaExplorer.constName,
```
jeromy-cannon, line 106: I think we can just remove flags.deployHederaExplorer, since it is now redundant.

test/e2e/commands/network.test.mjs
```diff
@@ -103,15 +103,10 @@ describe('NetworkCommand', () => {
...
-        flags.deployMirrorNode.constName,
```
jeromy-cannon, line 107: I think we can remove flags.deployMirrorNode from the flags file and the entire system, since it is implied when we call `solo mirror-node deploy`

JeffreyDallas: done

src/core/k8.mjs
```diff
@@ -343,8 +343,7 @@ export class K8 {
    * @param {string} containerName
    * @param {string} destPath - path inside the container
    * @param {number} [timeout] - timeout in ms
-   * @returns {Promise<{owner: string, size: number, modifiedAt: string, name: string, directory: boolean, group: string}[]>}
-   * array of directory entries, custom object
+   * @returns array of directory entries, custom object
```
jeromy-cannon, line 346: this is async, so it still returns a promise

JeffreyDallas: Updated

jeromy-cannon (approved):

PR #621, opened 2024-09-19
feat: add quiet flag

### LINKED ISSUE #580: Add `--quiet` mode to suppress all command/subcommand prompting
we should add a `-q` or `--quiet` flag to all commands/subcommands to suppress all prompting.  If a required value was not set, we should inform the user of the required value that was not set and exit gracefully.  This is to allow a good CLI experience for advanced users and for using in scripts where no human interaction is available.

Except InitCommand, it is scheduled to be removed: #578

JeffreyDallas:
## Description

* add quiet flag to command that might need prompt to set flags with unknown values

### Related Issues

* Closes #580


src/commands/mirror_node.mjs
```diff
@@ -158,7 +159,9 @@ export class MirrorNodeCommand extends BaseCommand {
             flags.mirrorNodeVersion
           ])
 
-          await prompts.execute(task, self.configManager, MirrorNodeCommand.DEPLOY_FLAGS_LIST)
+          if (!MirrorNodeCommand.DEPLOY_FLAGS_LIST.includes(flags.quiet)) {
```
jeromy-cannon, line 162: instead of doing the checks here, if we do it here:
https://github.com/hashgraph/solo/blob/ea01a94088ec09eb7848441b947c5e7377f2219a/src/commands/prompts.mjs#L490-L490
```
  if (configManager.getFlag(flags.quiet)) {
    return
  }
```
then we can remove all of these if checks.

jeromy-cannon (changes requested):
see comments

jeromy-cannon (approved):

Issue #360, opened 2024-06-06
Add an integration test suite


jeromy-cannon:
currently there is only Unit Tests and E2E Tests.  There are tests in both that should be in a new integration test suite.  This is to create the integration test suite and move the `tests/e2e/core` tests into the new integration folder.  Also, prove it is running correctly locally and in the CICD workflows.

- [ ] add `test/int` directory
- [ ] move `test/e2e/core` tests into `test/e2e/int/core`
- [ ] add `package.json` script to run the integration tests and ensure that unit and e2e are not running the integration tests
- [ ] update CICD workflows to call the integration tests
  - [ ] `flow-build-application.yaml`
  - [ ] `flow-pull-request-checks.yaml`

required by https://github.com/hashgraph/solo/issues/359

PR #624, opened 2024-09-24
fix(tests): start Kind cluster on demand if not having a kube config errors

instamenta:
## Description

Start Kind cluster on demand if not having a kube config errors

### Related Issues

* Closes # [359](https://github.com/hashgraph/solo/issues/359)

### Note

This pull request might also render [issue #360](https://github.com/hashgraph/solo/issues/360) obsolete

### DIFF
--- a/test/test_util.js
+++ b/test/test_util.js
@@ -370,6 +370,7 @@ async function addKeyHashToMap (k8, nodeId, keyDir, uniqueNodeDestDir, keyHashMa
 export function getK8Instance (configManager) {
   try {
     return new K8(configManager, testLogger)
+    // TODO: return a mock without running the init within constructor after we convert to Mocha, Jest ESModule mocks are broke.
   } catch (e) {
     if (!(e instanceof FullstackTestingError)) {
       throw e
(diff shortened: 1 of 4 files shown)


test/test_util.js
```diff
@@ -360,3 +362,26 @@ async function addKeyHashToMap (k8, nodeId, keyDir, uniqueNodeDestDir, keyHashMa
...
   keyHashMap.set(privateKeyFileName, crypto.createHash('sha256').update(keyString).digest('base64'))
 }
+
+/**
+ * @param {ConfigManager} configManager
+ * @returns {K8}
+ */
+export function getK8Instance (configManager) {
+  try {
+    return new K8(configManager, testLogger)
```
jeromy-cannon, line 372: so, on the pipeline this command runs and passes and allows the test cases to execute without interacting with a live cluster, but when my machine it would fail if kube context is not set to an active cluster. 

I see below that you have added code to create a kind cluster, but, the goal was to some how possibly load a mock or something or update the code so that it would not fail when a kube config is found but the kubernetes cluster is not running, since it isn't really needed for the unit test case.  

Also, we don't want to bind ourselves to Kind.

jeromy-cannon (changes requested):
update so that unit test case can run without needing an active k8s cluster.

test/test_util.js
jeromy-cannon:
```suggestion
    return new K8(configManager, testLogger)
    // TODO: return a mock without running the init within constructor after we convert to Mocha, Jest ESModule mocks are broke.
```

jeromy-cannon (approved):

PR #713, opened 2024-10-18
feat: removed caching of ConfigManager and its flags

### LINKED ISSUE #676: Remove ConfigManager caching capabilities to improve user experience
While the caching of the ConfigManager reduces the number of flags that a user has to enter, it also causes a lot of confusion and sometimes doesn't automatically get uncached.  Considering our future designs with multi-cluster the caching of these flags needs to be removed.  We want the user to either use the prompts or the flags for each command, and the number of commands will be reduced in the future.

Here is a list, although there might be need to be more items that aren't documented:

```[tasklist]
### Tasks
- [x] update: `ConfigManager.constructor`, remove 2nd parameter `cachedConfigFile`
- [x] update: `ConfigManager.applyPrecedence` jsdoc refers to cached config
- [x] update: `ConfigManager.update`, remove 2nd parameter `persist`
- [x] update: `ConfigManager.persist`
- [x] remove: `ConfigManager.load`
- [x] remove: `constants.SOLO_CONFIG_FILE`
- [ ] many test cases will need updating and refactoring since they rely upon caching of the flags
- [ ] the workflow that generates the README will need to be updated
- [ ] the README template file may also need to be updated
```

jeromy-cannon:
## Description

* removed caching of ConfigManager and its flags
* changes to remove lint warnings
* enhanced update readme workflow so that it will fail correctly if any of the solo calls fail with an error
* deduplicated .gitignore (lint warnings)
* updated readme template and readme due to impacted changes
* added esling: tseslint.configs.recommended
* added some missing flags for commands that were getting overlooked due to caching
* fail flag prompt if non-interactive terminal detected
* made getNodeLogs concurrent
* set chai config to not truncate error messages
* removed some TODOs by addressing them

### Related Issues

* Closes #676

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,11 +80,17 @@ Example output
 
 ```
 Creating cluster "solo" ...
- ✓ Ensuring node image (kindest/node:v1.29.1) 🖼
+ • Ensuring node image (kindest/node:v1.27.3) 🖼  ...
+ ✓ Ensuring node image (kindest/node:v1.27.3) 🖼
+ • Preparing nodes 📦   ...
  ✓ Preparing nodes 📦 
+ • Writing configuration 📜  ...
  ✓ Writing configuration 📜
+ • Starting control-plane 🕹️  ...
  ✓ Starting control-plane 🕹️
+ • Installing CNI 🔌  ...
  ✓ Installing CNI 🔌
+ • Installing StorageClass 💾  ...
  ✓ Installing StorageClass 💾
 Set kubectl context to "kind-solo"
 You can now use your cluster with:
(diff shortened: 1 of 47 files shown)


instamenta (approved):

README.md
```diff
@@ -144,24 +150,29 @@ solo init -t v0.54.0-alpha.4 -i node1,node2,node3 -n "${SOLO_NAMESPACE}" -s "${S
 Version			: 0.31.1
 Kubernetes Context	: kind-solo
 Kubernetes Cluster	: kind-solo
-Kubernetes Namespace	: solo
+Kubernetes Namespace	: undefined
```
JeffreyDallas, line 153: somehow the namespace is not output correctly here

jeromy-cannon: solo init no longer has a namespace option since there is no caching, that is why it is undefined.

README.md
```diff
@@ -208,11 +231,14 @@ solo cluster setup
...
```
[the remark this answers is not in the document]
jeromy-cannon: solo cluster setup has a cluster setup namespace but not a regular namespace, thus it is undefined.

README.md
jeromy-cannon: I've opened a GH Issue to collapse that when it is undefined.: https://github.com/hashgraph/solo/issues/721

PR #743, opened 2024-10-25
feat: change mirror node and explorer from subchart to individual chart

### LINKED ISSUE #617: Update mirror node deployment to deploy mirror node external of Solo Charts, similar to JSON RPC Relay
- mirror node deployment using chart directly instead of leveraging solo-deployment sub-chart settings
- mirror node explorer deployment using chart directly instead of leveraging solo-deployment sub-chart settings

---
- relates to: https://github.com/hashgraph/full-stack-testing/issues/974
- depends on: https://github.com/hashgraph/full-stack-testing/issues/675
- depends on: https://github.com/hashgraph/hedera-mirror-node-explorer/issues/877

JeffreyDallas:
## Description

* change mirror node and explorer from subchart to individual chart

### Related Issues

* Closes #617 

---

### Notes

Mus not use `hedera-mirror` as release name, since
k8 would auto generate env variable `${SERVICE_NAME}_PORT`
which lead to a new `HEDERA_MIRROR_REST_PORT` clash with pre-existing env variable
used by mirror node.

https://stackoverflow.com/questions/45759205/what-environment-variables-are-created-in-kubernetes-by-default

image


resources/hedera-explorer-values.yaml
```diff
@@ -0,0 +1,49 @@
+ingress:
```
jeromy-cannon, line 1: ingress might require `haproxyIngressController`, let me know if that is the case or if I'm mistaken.

src/commands/flags.ts
```diff
@@ -323,7 +323,7 @@ export const enableHederaExplorerTls: CommandFlag = {
   constName: 'enableHederaExplorerTls',
   name: 'enable-hedera-explorer-tls',
   definition: {
-    describe: 'Enable the Hedera Explorer TLS, defaults to false',
+    describe: 'Enable the Hedera Explorer TLS, defaults to false, must also enable deployCertManager and deployCertManagerCrds',
```
jeromy-cannon, line 326:
```suggestion
    describe: 'Enable the Hedera Explorer TLS, defaults to false, requires certManager and certManagerCrds, which can be deployed through solo-cluster-setup chart or standalone',
```

jeromy-cannon (approved):

resources/hedera-explorer-values.yaml
jeromy-cannon: @JeffreyDallas

JeffreyDallas: I think this ingress is different one vs haproxy-ingress

jeromy-cannon: haproxyIngressController is what we used to configure the ingress in hedera mirror node explorer, those ingress value configurations are used here:

https://github.com/hashgraph/hedera-mirror-node-explorer/blob/main/chart/templates/ingress.yaml

As long as it tests all okay we should be good.  self-signed can be tested locally.  I had previously used GKE to test the letsencrypt signed certificates.

JeffreyDallas: Test passed with new explorer chart 0.2.1 
with both `selfSignedClusterIssuer` enabled or disabled

PR #765, opened 2024-10-30
fix: nextNode ID for release version > 0.56.0

### LINKED ISSUE #742: Remove NextNodeId from config.txt if version of hedera is >= 0.56
As of release 0.56, the use of nextNodeId in the config.txt is going away.  The PR for this code change is here: https://github.com/hashgraph/hedera-services/pull/15791
Config files which are automatically generated in software upgrades through DAB are automatically correct.
How does this impact you?
In release 0.56 and later you will need to remove the field nextNodeId from any stored config.txt files and any tools which render config.txt files.
```[tasklist]
### Tasks
- [x] update the [`profile_manager.ts prepareConfigTxt()`](https://github.com/hashgraph/solo/blob/de6968c7a83c1c0da09eff4581825aeed86da4b1/src/core/profile_manager.ts#L454-L454) method condition to exclude adding the `nextNodeId` if the `releaseVersion.minor` is >= 56
- [x] the config.txt generated by Solo should not add the nextNodeId into the file
```
NOTE: as mentioned above, when we copy the config.txt from the nodes `${constants.HEDERA_HAPI_PATH}/data/upgrade/current/config.txt`, it will already be correct, so no updates are needed for these.

JeffreyDallas:
## Description

* no longer need nextNodeID for release version >= 0.56.0

### Related Issues

* Closes #742


jeromy-cannon (dismissed):

andrewb1269hg (approved):
Reviewed `.github/workflows/zxc-e2e-test.yaml`

Looks good, approved.

jeromy-cannon (approved):

PR #854, opened 2024-11-20
fix: Error in `node update` separate commands when new admin key is not supplied

Ivo-Yankov:
## Description

* TBD

### Related Issues

* Closes #


src/core/helpers.ts
jeromy-cannon, line 1: I see it took 9 runs before the e2e test passed, do you think this issue is resolved?  or are those errors unrelated to this fix?

jeromy-cannon (approved):

src/core/helpers.ts
Ivo-Yankov: It's unrelated to the issue in this PR because I am seeing the same error in other PRs

PR #856, opened 2024-11-20
fix: resolve node update test failures

### LINKED ISSUE #855: Node Update tests fail with INVALID_NODE_ACCOUNT
### Describe the bug

Node Update tests are failing.

Node Update tests should succeed.

### To Reproduce

1. Run `npm run build && npm run test-e2e-node-update`

### Additional Context

_No response_

nathanklick:
## Description

* Fixes the `INVALID_NODE_ACCOUNT` failure consistently observed on the node update tests.
* Disables the Node Add suite of E2E tests which are failing because the teacher crashes when reconnecting the new node. Oleg & Alex are currently having similar issues with larger HashSphere nodes (4 CPU, 32 GB RAM, -Xmx22g). 

### Related Issues

* Closes #855


andrewb1269hg (approved):
Review applies to: `.github/workflows/flow-pull-request-checks.yaml`

rbarkerSL (approved):

JeffreyDallas (approved):

src/core/account_manager.ts
```diff
@@ -214,25 +214,19 @@ export class AccountManager {
    * @returns a node client that can be used to call transactions
    */
   async _getNodeClient (namespace: string, networkNodeServicesMap: Map<string, NetworkNodeServices>, operatorId: string,
-    operatorKey: string) {
-    const nodes = {}
+    operatorKey: string, useFirstNodeOnly = true) {
```
jeromy-cannon, line 217: what happens when set useFirstNodeOnly, and first node is target for `solo node delete --node-aliase node1`

jeromy-cannon: I think it should be fine.  We rebuild the node client after the network is frozen.

nathanklick: bad stuff, I have a change in flight to use the second node only.

jeromy-cannon: but, what if they decide to delete the second node and keep the first?

PR #862, opened 2024-11-22
feat: Create remote config 

instamenta:
## Description

Adding functionality to create remote config which is kept inside the Kubernetes cluster in a configmap.
  The remote config includes data about:
  * migrations
  * components state like consensus nodes, relays, explorers
  * history of executed commands
  * clusters

Added new command `deployment create` for creating the remote config

### Related Issues

* Closes #
- https://github.com/hashgraph/solo/issues/772
- https://github.com/hashgraph/solo/issues/773
- https://github.com/hashgraph/solo/issues/771
- https://github.com/hashgraph/solo/issues/808


package.json
```diff
@@ -87,6 +87,7 @@
     "@types/got": "^9.6.12",
     "@types/inquirer": "^9.0.7",
     "@types/ip": "^1.1.3",
+    "@types/js-yaml": "^4.0.9",
```
jeromy-cannon, line 90: If we can use `yaml` instead of `js-yaml` that would be better.  Nathan wants us to switch to using `yaml` because it is more maintained.   `js-yaml` hasn't been updated in about 4 years.

instamenta: Isn't js-yaml more popular? 
Also if it's stable there shouldn't really be any updates since the core functionality doesn't change I wouldn't call it stale

instamenta: also js-yaml seems more stable

src/commands/deployment.ts
```diff
@@ -0,0 +1,149 @@
...
+    interface Context { config: Config }
+
+    const tasks = new Listr<Context>([
+      {
+        title: 'Initialize',
+        task: async (ctx, task): Promise<Listr<Context, any, any>> => {
+          self.configManager.update(argv)
+          self.logger.debug('Loaded cached config', { config: self.configManager.config })
+
+          await prompts.execute(task, self.configManager, [flags.namespace])
```
jeromy-cannon, line 58: might need some more required prompts:
```
❯ solo deployment create -n solo-e2e

Version			: 0.99.0
Kubernetes Context	: kind-solo-e2e
Kubernetes Cluster	: kind-solo-e2e
Kubernetes Namespace	: solo-e2e
(node:62323) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
✖ Cannot read properties of undefined (reading 'split')
◼ Prompt local configuration
◼ Validate cluster connections
◼ Create remote config
*********************************** ERROR *****************************************
Error installing chart solo-deployment
***********************************************************************************
```

jeromy-cannon: when I gave it the namespace, it prompted me for the deployment name.  but, the deployment name and the namespace are the same thing, so I think we should either make them interchangeable or get rid of deployment-name

src/core/config/remote/remote_config_data_wrapper.ts
```diff
@@ -0,0 +1,155 @@
...
+import * as yaml from 'yaml'
+import { RemoteConfigMetadata } from './metadata.js'
+import { ComponentsDataWrapper } from './components_data_wrapper.js'
+import * as constants from '../../constants.js'
+import type { Cluster, Version, Namespace, RemoteConfigData, RemoteConfigDataStructure } from './types.js'
+import type * as k8s from '@kubernetes/client-node'
+import type { ToObject, Validate } from '../../../types/index.js'
+
+export class RemoteConfigDataWrapper implements Validate, ToObject<RemoteConfigDataStructure> {
+  private readonly _version: Version = version.HEDERA_PLATFORM_VERSION
```
jeromy-cannon, line 28:
```suggestion
  private readonly _version: Version = '1.0.0'
```
the comment in the [notion page](https://www.notion.so/swirldslabs/Multi-user-and-Multi-cluster-Support-Solo-State-and-Lease-Mechanism-DRAFT-8b013e8befe04e3daffc889619c851c1?pvs=4) was
`# solo state maintaines its own version increments, not same as solo version`

src/core/config/remote/enumerations.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Enumerations that represent the component types used in remote config
+ * {@link ComponentsDataWrapper}
+ */
+export enum ComponentType {
+  ConsensusNode = 'consensus nodes',
+  HaProxy = 'ha proxies',
+  EnvoyProxy = 'envoy proxies',
+  MirrorNode = 'mirror nodes',
+  MirrorNodeExplorer = 'mirror node explorers',
+  Relay = 'replays',
```
jeromy-cannon, line 28:
```suggestion
  ConsensusNode = 'consensus-node',
  HaProxy = 'ha-proxy',
  EnvoyProxy = 'envoy-proxy',
  MirrorNode = 'mirror-node',
  MirrorNodeExplorer = 'mirror-node-explorer',
  Relay = 'relay',
```
since this seems to be ending up in the solo-remote-config as yaml, then it should probably be well formed.  Instead of:
```
 apiVersion: v1                                                                                                                                                                                                                           
 data:                                                                                                                                                                                                                                    
   remote-config-data: |                                                                                                                                                                                                                  
     metadata:                                                                                                                                                                                                                            
       name: solo-e2e                                                                                                                                                                                                                     
       lastUpdatedAt: 2024-11-26T22:23:55.286000Z                                                                                                                                                                                         
       lastUpdateBy: jeromy@swirldslabs.com                                                                                                                                                                                               
     version: v0.56.5                                                                                                                                                                                                                     
     clusters:                                                                                                                                                                                                                            
       kind-solo-e2e: solo-e2e                                                                                                                                                                                                            
     components:                                                                                                                                                                                                                          
       replays: {}                                                                                                                                                                                                                        
       ha proxies: {}                                                                                                                                                                                                                     
       mirror nodes: {}                                                                                                                                                                                                                   
       envoy proxies: {}                                                                                                                                                                                                                  
       consensus nodes: {}                                                                                                                                                                                                                
       mirror node explorers: {}                                                                                                                                                                                                          
     commandHistory:                                                                                                                                                                                                                      
       - deployment create                                                                                                                                                                                                                
     lastExecutedCommand: deployment create                                                                                                                                                                                               
 kind: ConfigMap                                                                                                                                                                                                                          
 metadata:                                                                                                                                                                                                                                
   creationTimestamp: "2024-11-26T22:23:55Z"                                                                                                                                                                                              
   labels:                                                                                                                                                                                                                                
     solo.hedera.com/type: remote-config                                                                                                                                                                                                  
   name: solo-remote-config                                                                                                                                                                                                               
   namespace: solo-e2e                                                                                                                                                                                                                    
   resourceVersion: "2842"                                                                                                                                                                                                                
   uid: 1d9fc502-54ed-4d0e-a983-66e1195fc6e1                                                                                                                                                                                              
```

src/commands/deployment.ts
```diff
@@ -0,0 +1,149 @@
...
+import type { CommandFlag, ContextClusterStructure } from '../types/index.js'
+
+export class DeploymentCommand extends BaseCommand {
+  private static get DEPLOY_FLAGS_LIST (): CommandFlag[] {
+    return [
+      flags.quiet,
+      flags.namespace,
+      flags.deploymentName,
+      flags.userEmailAddress,
+      flags.deploymentClusters,
```
jeromy-cannon, line 36: you give the option to supply `--deployment-clusters`, but when I use it, it still looks for `--context-cluster`.   There is some more work that Ivo is doing in LocalConfig in his current PR that should help us get the context when you already have the cluster.  I'm wondering if we want to use that.  Or, if you want to just remove deploymentClusters?   Not sure what the purpose would be otherwise.

```
❯ cat ~/.solo/cache/local-config.yaml
userEmailAddress: jeromy@swirldslabs.com
deployments:
  solo-e2e:
    clusters:
      - kind-solo-e2e
currentDeploymentName: solo-e2e
❯ k9s -A
❯ ls -al ~/.solo/logsq
❯ solo deployment create -n solo-e2e --deployment-clusters kind-solo-e2e

Version			: 0.99.0
Kubernetes Context	: kind-solo-e2e
Kubernetes Cluster	: kind-solo-e2e
Kubernetes Namespace	: solo-e2e
(node:63822) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
✖ Cannot read properties of undefined (reading 'split')
◼ Prompt local configuration
◼ Validate cluster connections
◼ Create remote config
*********************************** ERROR *****************************************
Error installing chart solo-deployment
***********************************************************************************
```

src/core/templates.ts
```diff
@@ -220,4 +222,29 @@ export class Templates {
...
+
+    unparsed.split(',').forEach((data) => {
+      const [context, cluster] = data.split('=') as [Context, Cluster]
+
+      if (!context || typeof context !== 'string') {
+        throw new SoloError('Invalid context in context-cluster', null, { data })
+      }
+
+      if (!cluster || typeof cluster !== 'string') {
+        throw new SoloError('Invalid cluster in context-cluster', null, { data })
```
jeromy-cannon, line 242: maybe a better error message, and explain the format so that they don't have to do a `--help`.

```
❯ solo deployment create --namespace solo-e2e --deployment-name solo-e2e --deployment-clusters kind-solo-e2e --context-cluster kind-solo-e2e

Version			: 0.99.0
Kubernetes Context	: kind-solo-e2e
Kubernetes Cluster	: kind-solo-e2e
Kubernetes Namespace	: solo-e2e
(node:89410) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
✖ Invalid cluster in context-cluster
◼ Prompt local configuration
◼ Validate cluster connections
◼ Create remote config
*********************************** ERROR *****************************************
Error installing chart solo-deployment
***********************************************************************************
```

src/core/config/remote/remote_config_manager.ts
```diff
@@ -0,0 +1,288 @@
...
+
+  private setDefaultNamespaceIfNotSet (): void {
+    if (this.configManager.hasFlag(flags.namespace)) return
+
+    if (!this.localConfig?.currentDeploymentName) {
+      this.logger.error('Current deployment name is not set in local config', this.localConfig)
+      throw new SoloError('Current deployment name is not set in local config')
+    }
+
+    // TODO: Current quick fix for commands where namespace is not passed
```
jeromy-cannon, line 258: Which commands don't have namespace and are loading remoteConfig?

src/commands/node/handlers.ts
```diff
@@ -552,6 +570,7 @@ export class NodeCommandHandlers {
 
     const action = helpers.commandActionBuilder([
       this.tasks.initialize(argv, keysConfigBuilder.bind(this), null),
+      RemoteConfigTasks.loadRemoteConfig.bind(this)(argv),
```
jeromy-cannon, line 573: `solo node keys` is non mutating it has no need for RemoteConfig or LeaseManager.  It just creates keys on the local users file system for use later.  It can be ran at anytime.

src/core/config/remote/remote_config_tasks.ts
```diff
@@ -0,0 +1,246 @@
...
+
+  /* ----------- Create and Load ----------- */
+
+  /**
+   * Loads the remote config from the config class.
+   *
+   * @param argv - used to update the last executed command and command history
+   */
+  public static loadRemoteConfig (this: BaseCommand, argv: any): ListrTask<any, any, any> {
+    return this.remoteConfigManager.buildLoadTask(argv)
```
jeromy-cannon, line 46: I'm concerned about how frequently this is being called.  

Looking through handlers.ts, I see the issue you are trying to solve.  I think if you only do the load and modify in the functions that have the commandActionBuilder, then you will remove all of the duplication.

src/core/config/remote/components/base_component.ts
```diff
@@ -0,0 +1,120 @@
...
+  /* -------- Utilities -------- */
+
+  /**
+   * Compares two BaseComponent instances for equality.
+   *
+   * @param x - The first component to compare
+   * @param y - The second component to compare
+   * @returns boolean - true if the components are equal
+   */
+  public static compare (x: BaseComponent, y: BaseComponent): boolean {
```
jeromy-cannon, line 87: did you mean to include name?  or is this only meant to compareType?

src/core/k8.ts
```diff
@@ -1256,11 +1337,20 @@ export class K8 {
...
+  /* ------------- Utilities ------------- */
+
+  /**
+   * @param response - response object from the kubeclient call
+   * @param error - body of the response becomes the error if the status is not OK
+   * @param errorMessage - the error message to be passed in case it fails
+   *
+   * @throws SoloError - if the status code is not OK
+   */
+  private _handleKubernetesClientError (response: http.IncomingMessage, error: Error | any, errorMessage: string): void {
```
jeromy-cannon, line 1349:
```suggestion
  private handleKubernetesClientError (response: http.IncomingMessage, error: Error | any, errorMessage: string): void {
```
https://google.github.io/styleguide/tsguide.html#identifiers-underscore-prefix-suffix

src/core/k8.ts
```diff
@@ -1190,6 +1205,72 @@ export class K8 {
...
+    const configMap = new k8s.V1ConfigMap()
+    configMap.data = data
+
+    const metadata = new k8s.V1ObjectMeta()
+    metadata.name = name
+    metadata.namespace = namespace
+    metadata.labels = labels
+    configMap.metadata = metadata
+    try {
+      const resp  =await this.kubeClient.replaceNamespacedConfigMap(name, namespace, configMap)
```
jeromy-cannon, line 1266:
```suggestion
      const resp = await this.kubeClient.replaceNamespacedConfigMap(name, namespace, configMap)
```
I think you need to run `npm run format`

src/core/config/remote/remote_config_manager.ts
instamenta: node logs, and etc. Since we need the remote config loaded to update the lastExecuted command and command History

src/core/config/remote/components/base_component.ts
```diff
@@ -0,0 +1,120 @@
...
+    if (!this.namespace || typeof this.namespace !== 'string') {
+      throw new SoloError(`Invalid namespace: ${this.namespace}`)
+    }
+
+    if (!Object.values(ComponentType).includes(this.type)) {
+      throw new SoloError(`Invalid component type: ${this.type}`)
+    }
+  }
+
+  public toObject (): Component {
```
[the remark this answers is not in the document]
instamenta: return types are not always written explicitly if this is what you asked

src/core/config/remote/remote_config_manager.ts
jeromy-cannon: I think it is okay if we don't capture history for non-mutating commands.  the history was an optional non-MVP item anyway.

src/commands/node/tasks.ts
```diff
@@ -608,6 +600,9 @@ export class NodeCommandTasks {
     const self = this
     return new Task('Download generated files from an existing node', async (ctx: any, task: ListrTaskWrapper<any, any, any>) => {
       const config = ctx.config
+
+      // TODO: Note seems this doesn't work if you have only one node
```
jeromy-cannon, line 604: if you have only one node, you cannot do a node delete.

jeromy-cannon (approved):

jeromy-cannon (approved):

PR #863, opened 2024-11-22
feat: Implement solo context use

Ivo-Yankov:
## Description

* TBD

### Related Issues

* Closes #

### DIFF
--- a/src/commands/context/flags.ts
+++ b/src/commands/context/flags.ts
@@ -14,10 +14,17 @@
  * limitations under the License.
  *
  */
-export class Task {
-  constructor (public title: string, public task: Function, public skip: Function | boolean = false) {
-    this.title = title
-    this.task = task
-    this.skip = skip
-  }
+
+import * as flags from '../flags.js'
+
+export const DEFAULT_FLAGS = {
+    requiredFlags: [],
+    requiredFlagsWithDisabledPrompt: [flags.namespace, flags.cacheDir, flags.releaseTag],
+    optionalFlags: [flags.devMode]
+}
+
+export const USE_FLAGS = {
+    requiredFlags: [],
+    requiredFlagsWithDisabledPrompt: [],
+    optionalFlags: [flags.devMode, flags.quiet, flags.clusterName, flags.context, flags.force, flags.namespace]
 }
--- a/src/commands/context/tasks.ts
+++ b/src/commands/context/tasks.ts
@@ -1,0 +1,88 @@
+/**
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the ""License"");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an ""AS IS"" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+import {
+  Task, Templates
+} from '../../core/index.js'
+import * as flags from '../flags.js'
+import type { ListrTaskWrapper } from 'listr2'
+import { type BaseCommand } from '../base.js'
+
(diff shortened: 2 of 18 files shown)


src/commands/context/flags.ts
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import * as flags from '../flags.js'
+
+export const DEFAULT_FLAGS = {
+    requiredFlags: [],
+    requiredFlagsWithDisabledPrompt: [flags.namespace, flags.cacheDir, flags.releaseTag],
```
jeromy-cannon, line 22: what makes releaseTag required or used for `solo context use`?

src/commands/context/tasks.ts
```diff
@@ -0,0 +1,104 @@
...
+      const deployments = this.parent.getLocalConfig().deployments
+      deployments[currentDeploymentName].clusterAliases = clusterAliases
+      this.parent.getLocalConfig().setDeployments(deployments)
+
+      this.parent.getK8().getKubeConfig().setCurrentContext(contextName)
+
+      this.parent.logger.info(`currentDeploymentName: ${currentDeploymentName}`)
+      this.parent.logger.info(`contextName: ${contextName}`)
+      this.parent.logger.info(`clusterAliases: ${clusterAliases.join(' ')}`)
+      this.parent.logger.info('Save LocalConfig file')
```
jeromy-cannon, line 85:
```suggestion
      this.parent.logger.info('Save LocalConfig file: [currentDeploymentName: ${currentDeploymentName}, contextName: ${contextName}, clusterAliases: ${clusterAliases.join(' ')}]')
```
less verbose in the log file if it is on one line

src/core/config/local_config.ts
jeromy-cannon, line 1: I recall us talking about storing the cluster to context mapping by reading it in, but not saving it to the local config file, so that it would be rebuilt each time and can be used by other parts of the application.  Maybe I'll find it somewhere else while reviewing this PR?

src/core/task.ts
```diff
@@ -15,11 +15,9 @@
...
-    return {
-      title,
-      task: taskFunc,
-      skip
-    }
+  constructor (public title: string, public task: Function, public skip: Function | boolean = false) {
+    this.title = title
+    this.task = task
+    this.skip = skip
   }
```
jeromy-cannon, line 22:
```suggestion
  constructor (public title: string, public task: Function, public skip: Function | boolean = false) { }
```
I think this means the same thing?

src/core/k8.ts
```diff
@@ -565,28 +565,28 @@ export class K8 {
         inputStream.pipe(inputPassthroughStream)
 
         execInstance.exec(namespace, podName, containerName, command, null, errPassthroughStream, inputPassthroughStream, false,
-          ({ status }) => self.handleCallback(status, localContext, messagePrefix))
-          .then(conn => {
-            self.logger.info(`${messagePrefix} connection established`)
-            localContext.connection = conn
+            ({ status }) => self.handleCallback(status, localContext, messagePrefix))
```
jeromy-cannon, line 568: what formatter are you using?  I noticed that `npm run format` does not do this.  I have an outstanding item to implement prettier.  although now that we agreed on google style, there is probably a better way.

In IntelliJ, I had the Palantir Style set, previously in Java, so some of those settings carried over.  If I reformat this file, it changes everything back to how I had it.  We need to all get on the same page and use the same settings.

src/commands/flags.ts
```diff
@@ -748,6 +748,16 @@ export const userEmailAddress: CommandFlag = {
   }
 }
 
+export const context: CommandFlag = {
+  constName: 'contextName',
+  name: 'context',
+  definition: {
+    describe: 'The kind context name to be used',
```
jeromy-cannon, line 755:
```suggestion
    describe: 'The Kubernetes context name to be used',
```

src/core/k8.ts
Ivo-Yankov: I usually run `npm run format` but I think in this case the change was due to a merge conflict.

jeromy-cannon: `npm run format` is not really doing whitespace until we configure prettier, which is something I was going to do soon now that we are aligned on Google code style.

jeromy-cannon (approved):

PR #901, opened 2024-12-02
chore: google code style integration

jeromy-cannon:
## Description

* adds prettier, eslint, and tsconfig settings from google code style
* sets numerous eslint items to warn and added a `// TODO remove` for follow up in separate PRs
* eslint setting no this alias except self (added // TODO remove for allowing self)
* only ran npm run format using the new eslint rules
* and fixed an error caused by // @ts-ignore moving to a different line in test/e2e/integration/core/account_manager.test.ts

### Related Issues

* Closes #


instamenta (approved):

.prettierrc.json
instamenta, line 1: will trailingComma set to "all" put comma on this for example
```
const fooBar = {  foo: 'foo', bar: 'bar', }
```

instamenta (approved):

.prettierrc.json
jeromy-cannon: no.  Only when it wraps, see the two examples below.  

```
import {it, describe, after, before} from 'mocha';
import {
  AccountCreateTransaction,
  AccountId,
  Client,
  Hbar,
  HbarUnit,
  Logger,
  LogLevel,
  PrivateKey,
  Status,
  TopicCreateTransaction,
  TopicMessageSubmitTransaction,
} from '@hashgraph/sdk';
```

PR #912, opened 2024-12-03
ci: Reverting the codecov upload action to an older version to fix an ongoing issue.

### LINKED ISSUE #911: ci: Fix issues with latest version of Codecov by reverting
The latest version of Codecov upload action can not correctly input the codecov-token. Issue is described [here](https://github.com/codecov/codecov-action/issues/1709). We need to revert to older version 5.0.2 until issue is fixed.

san-est:
## Description

* Currently there is an issue with the latest version of codecov-upload action where it does not properly use the codecov_token. We are now using version 5.0.2 everywhere as we know its stable until they fix the issue in the latest version.

### Related Issues

* Closes #911


jeromy-cannon (approved):

andrewb1269hg (approved):
There are 2 changes in this. I believe the changes in `templates` are unnecessary but approving both since it brings the codecov action to version 5.0.2.

jeromy-cannon:
> There are 2 changes in this. I believe the changes in `templates` are unnecessary but approving both since it brings the codecov action to version 5.0.2.

the templates get out of date until we run them, then we have to update them.  definitely better to keep them up to date as we go, thank you for updating the template too!

PR #923, opened 2024-12-04
feat: add extended math utils and duration support

### LINKED ISSUE #867: Production Readiness: Improve time handling logic
### Feature to Enhance

Time handling logic

### Requested Enhancement

Currently, all logic dealing with time conversions and computations are spread throughout the code base. The only consistent factor is most of these code locations depend on a set of constant values.

### Business Need

In order to improve code readability, reliability, and testability, there should only be one implementation of the time operations which is used throughout the code.

### Additional Context

_No response_

nathanklick:
## Description

* Adds a new `Duration` class based on the official Java implementation. 
* Adds a new `MathEx` class to support the `Duration` implementation.
* Adds a new `ArithmeticError` class to support proper exception handling.
* Replaces all usages of the `SECONDS` and `MINUTES` constants with the appropriate `Duration` methods.
* Updates the `sleep()` method to accept `Duration` instances as the argument.

### Related Issues

* Closes #867 

### Notes to Reviewers

- ~~This draft PR is still missing unit test coverage for the `Duration` class. If anyone has bandwidth please feel to contribute unit tests.~~

### DIFF
--- a/src/core/time/duration.ts
+++ b/src/core/time/duration.ts
@@ -1,0 +1,490 @@
+/**
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the ""License"");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an ""AS IS"" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+import {IllegalArgumentError} from '../errors.js';
+import {MathEx} from '../util/math_ex.js';
+import {Time} from './time.js';
+
+/**
+ * A time-based amount of time, such as '34.5 seconds'.
+ *
+ * This class models a quantity or amount of time in terms of seconds and nanoseconds. It can be accessed using other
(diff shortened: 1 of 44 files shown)


src/core/time/duration.ts
```diff
@@ -0,0 +1,517 @@
...
+  private readonly _nanos: number;
+
+  /**
+   * Creates a new instance of Duration with the specified number of seconds and nanoseconds.
+   * This is a private constructor and not intended to be called directly.
+   *
+   * @param seconds - the number of seconds
+   * @param nanos - the number of nanoseconds
+   */
+  private constructor(seconds: number, nanos: number) {
```
instamenta, line 70: You can just define them as readonly and you will achieve the same effect without needing a getter :)

src/core/time/time.ts
instamenta, line 1: Having the explicit types like this is a bit of anti-pattern in my opinion, I know this is the Java way but it might not be the best choice in our case. ( just my opinion )

instamenta (commented):
I like what I see definitely. Left some small suggestions that are up for debate

jeromy-cannon (dismissed):

src/core/time/duration.ts
nathanklick: Kotlin, Java, C/C++, and C# habits die hard.

src/core/time/time.ts
nathanklick: Yeah, in this case since these are constants I agree omitting the explicit type would be fine..

src/core/time/duration.ts
nathanklick: Fixed

src/core/time/time.ts
nathanklick: Fixed

jeromy-cannon (approved):

PR #924, opened 2024-12-04
chore: add eslint config @typescript-eslint/consistent-type-imports

jeromy-cannon:
## Description

* add eslint config @typescript-eslint/consistent-type-imports to reduce chances of circular dependencies

### Related Issues

* Relates to #928

### DIFF
--- a/src/commands/node/tasks.ts
+++ b/src/commands/node/tasks.ts
@@ -69,7 +69,13 @@ import chalk from 'chalk';
 import {flags} from '../index.js';
 import {type SoloLogger} from '../../core/logging.js';
 import type {Listr, ListrTaskWrapper} from 'listr2';
-import {ConfigBuilder, type NodeAlias, type NodeAliases, type PodName, SkipCheck} from '../../types/aliases.js';
+import {
+  type ConfigBuilder,
+  type NodeAlias,
+  type NodeAliases,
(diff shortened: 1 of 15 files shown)


Ivo-Yankov (approved):

src/commands/node/tasks.ts
```diff
@@ -69,7 +69,13 @@ import chalk from 'chalk';
 import {flags} from '../index.js';
 import {type SoloLogger} from '../../core/logging.js';
```
instamenta, line 70: I think there is a better option which would put it like `import type {SoloLogger} from '../../core/logging.js';
` instead of `import {type SoloLogger} from '../../core/logging.js';` which reduces code when only types are imported

instamenta (approved):

src/commands/node/tasks.ts
jeromy-cannon: It is true that it would reduce the number of characters in the file; but, I found it easier to add non-types if it already had braces on the outside.  If the braces are around just the type and not the keyword, then it needs to be manually converted, I don't think the IDE will fix it for you. (at least not in some situations).  This is why I chose to set the option this way.  I did notice previously where you had purposefully changed some to put the type on the outside of the braces, so it did make me pause and think about this.

Issue #42, opened 2024-01-16
Inspect existing cluster resources during cluster setup


leninmehedy:
`solo` needs to check for existing operator setup/configuration before installing `cluster setup chart`. 

```[tasklist]
### Tasks
- [ ] scan all pods for all namespaces for a label of 'app:cert-manager', if none exists install cert-manager and install acme-cluster-issuer, if it does exist install acme-cluster-issuer
- [ ] detect if minio operator has been installed if not, then install it, unless flag is set to not install minio
- [ ] detect if ingress controller is installed, if not, install the haproxy ingress controller
- [x] don't remove minio during destroy/uninstall ( obsolete )
- [ ] before destroying/uninstalling solo cluster setup chart check if any namespace has a remote config, if so prompt the user with a warning and if they wish to continue
- [ ] if all of the required components are installed then solo cluster setup chart does not need to be installed and can exit successfully
```
- relates to https://github.com/hashgraph/solo-charts/issues/58

PR #1094, opened 2025-01-03
feat: inspect existing cluster resources during cluster setup

instamenta:
## Description

* During `cluster setup` an instance of won't be deployed if it's already present:
  * `Minio`
  * `Prometheus Stack`
  * `Cert Manager`
  * `Cert Manager CRDS`

  If all of the above are present `solo-setup-chart` won't be installed.

* During `mirror-node deploy` an instance of won't be deployed if it's already present:
  * `Ingress Controller`

### Related Issues

* Closes [#42](https://github.com/hashgraph/solo/issues/42)


src/commands/cluster.ts
```diff
@@ -226,8 +261,24 @@ export class ClusterCommand extends BaseCommand {
...
-          task: async (ctx, _) => {
+          task: async (ctx, task) => {
             const clusterSetupNamespace = ctx.config.clusterSetupNamespace;
+
+            if (!argv.force && (await self.k8.isRemoteConfigPresentInAnyNamespace())) {
+              const confirm = await task.prompt(ListrEnquirerPromptAdapter).run({
+                type: 'toggle',
+                default: false,
+                message:
+                  'There is remote config for one of the deployments' +
```
jeromy-cannon, line 272: would be better if we can tell the user which deployment is still on the cluster

jeromy-cannon (dismissed):

jeromy-cannon (approved):

PR #1148, opened 2025-01-13
fix: use "double fork" to invoke port forward within Taskfile

### LINKED ISSUE #1115: Port-forward called from Taskfile exit itself
### Describe the bug

I just did a debug session with Pranali, she is trying to write gitbook documentation for solo.
One thing we found is that is using task to deploy solo, some port-forward would mysteriously exit itself without any errors/logs/warning.
This is the same thing I found when adding a github test flow for Task last December.
In summay, if calling kubectl port-forward  within task file, sometimes kubectl port-forward 
would quit itself, sometimes is haproxy-node1-svc , one sometimes is hedera-explorer one.
The workaround I used in github workflow is to re-calling port-forward outside taskfile.

Port-forward should keep running

### To Reproduce

run task default-with-mirror

### Additional Context

_No response_

JeffreyDallas:
## Description

* Tried various method suggested by stackoverflow, chatGPT, Gemini, all suggesting Taskfile may have some bugs/limitation that causing "port-forward" was created in incorrect process groups. The workaround is to use double fork to invoke a program.
* Remove re-enable port-foward from test script

### Related Issues

* Closes #1115


JeffreyDallas:
## Reference from Gemini

The Problem: Process Groups and Signal Propagation

The core issue is very likely related to process groups and how signals are propagated within them. Even with setsid (which creates a new session), if the kubectl port-forward process isn't properly placed in its own process group, Taskfile (or its parent process) might be sending signals that inadvertently terminate the kubectl process when a connection is established and then closed.

Here's why this happens:

Process Groups: Processes are organized into process groups. When a signal is sent to a process group, all processes within that group receive the signal.

Taskfile's Process Management: Taskfile, when it finishes executing a task, might be sending signals (like SIGHUP or SIGTERM) to its child processes and their process groups.

Connection Establishment and Closure: When you connect with nc, the following happens:

nc connects to the forwarded port.
kubectl handles the connection, forwarding traffic to the pod.
nc closes the connection.
This connection closure might be triggering some internal cleanup or signal handling within kubectl. If kubectl is still in the same process group as Taskfile, Taskfile's cleanup signals might then reach kubectl, causing it to exit.

The Solution: Explicit Process Group Management

The most reliable solution is to explicitly create a new process group for the kubectl port-forward process. This ensures that signals sent by Taskfile won't affect it

jeromy-cannon:
@JeffreyDallas , I already fixed this: https://github.com/hashgraph/solo/pull/1132

instamenta (approved):

Taskfile.helper.yml
```diff
@@ -442,7 +441,7 @@ tasks:
     deps:
       - task: "init"
     cmds:
-      - SOLO_HOME_DIR=${SOLO_HOME_DIR} npm run solo -- mirror-node destroy --namespace "${SOLO_NAMESPACE}" --force -q --dev || true
+      - SOLO_HOME_DIR=${SOLO_HOME_DIR} npm run solo -- mirror-node destroy --namespace "${SOLO_NAMESPACE}" --force --dev || true
```
jeromy-cannon, line 444: we should support `-q`/`--quiet` on all commands/subcommands.  Instead of removing the flag, just add the flag to the destroy subcommands.

JeffreyDallas: quiet flag added to mirror & relay destroy subcommands

PR #1154, opened 2025-01-14
chore:  account init before relay deployment

### LINKED ISSUE #1149: ensure that mirror node and json rpc relay deploys correctly when account init is ran after node start
ensure that mirror node and json rpc relay deploys correctly when account init is ran after node start, relay seems to have an option to pass in the operator id/key, is it defaulting to genesis if nothing is passed?  we should do a lookup on the operator id/key incase account init has ran

JeffreyDallas:
## Description

* Add test matrix to default smoke test
* Run "account init" after network deployment and node start , but before mirror & relay deployment
* Use generated account id and private key as parameters for relay node deployment

### Related Issues

* Closes #1149


mishomihov00 (approved):
Review applies to `.github/workflows/flow-task-test.yaml`

jeromy-cannon (changes requested):
I would like mirror node and relay to do a lookup to see if a secret exists for the default operator id's key.  if it exists, then use that instead of the genesis key.  I would like to see that both mirror node and json rpc relay can be deployed after an account init has been ran, without manually fetching the keys, but instead it is automated.

.github/workflows/script/solo_smoke_test.sh
```diff
@@ -86,7 +86,19 @@ function start_sdk_test ()
...
+
+# if first parameter equals to ACCOUNT_INIT,
+# then call solo account init before deploy mirror and relay node
+if [ "$1" == "ACCOUNT_INIT" ]; then
+  echo "Call solo account init"
+  npm run solo-test -- account init -n solo-e2e
+fi
+
+task solo:mirror-node
+task solo:relay
```
jeromy-cannon, line 98: if possible I would like to validate that the monitor/pinger functionality is active and running correctly

mishomihov00 (dismissed):
Review applies to `.github/workflows/flow-task-test.yaml`

src/commands/mirror_node.ts
```diff
@@ -263,9 +264,23 @@ export class MirrorNodeCommand extends BaseCommand {
                 const pod = networkPods[0];
                 ctx.config.valuesArg += ` --set monitor.config.hedera.mirror.monitor.nodes.0.accountId=${startAccId}`;
                 ctx.config.valuesArg += ` --set monitor.config.hedera.mirror.monitor.nodes.0.host=${pod.status.podIP}`;
+                ctx.config.valuesArg += ' --set monitor.config.hedera.mirror.monitor.nodes.0.nodeId=0';
 
                 ctx.config.valuesArg += ` --set monitor.config.hedera.mirror.monitor.operator.accountId=${constants.OPERATOR_ID}`;
-                ctx.config.valuesArg += ` --set monitor.config.hedera.mirror.monitor.operator.privateKey=${constants.OPERATOR_KEY}`;
+
+                // get operatorKey from k8s secret since it could have been changed
```
jeromy-cannon, line 271: The user can use an SDK to change the key, and then the one in the secrets would be invalid.

if the user supplied the operator key, use that
else if the secret exist, use that
else use genesis

src/commands/relay.ts
```diff
@@ -109,8 +110,13 @@ export class RelayCommand extends BaseCommand {
       valuesArg += ` --set config.OPERATOR_ID_MAIN=${operatorID}`;
     }
 
-    if (operatorKey) {
-      valuesArg += ` --set config.OPERATOR_KEY_MAIN=${operatorKey}`;
+    const secrets = await this.k8.getSecretsByLabel([`solo.hedera.com/account-id=${constants.OPERATOR_ID}`]);
```
jeromy-cannon, line 113: The user can use an SDK to change the key, and then the one in the secrets would be invalid.

if the user supplied the operator key, use that
else if the secret exist, use that
else use genesis

.github/workflows/script/solo_smoke_test.sh
JeffreyDallas: added helm acceptance, and monitor & importer log checks

src/commands/relay.ts
JeffreyDallas: Updated

src/commands/mirror_node.ts
JeffreyDallas: Updated

src/commands/mirror_node.ts
```diff
@@ -263,18 +272,36 @@ export class MirrorNodeCommand extends BaseCommand {
...
-
-                ctx.config.valuesArg += ` --set monitor.config.hedera.mirror.monitor.operator.accountId=${constants.OPERATOR_ID}`;
-                ctx.config.valuesArg += ` --set monitor.config.hedera.mirror.monitor.operator.privateKey=${constants.OPERATOR_KEY}`;
+                ctx.config.valuesArg += ' --set monitor.config.hedera.mirror.monitor.nodes.0.nodeId=0';
+
+                const operatorId = ctx.config.operatorId || constants.OPERATOR_ID;
+                ctx.config.valuesArg += ` --set monitor.config.hedera.mirror.monitor.operator.accountId=${operatorId}`;
+
+                if (ctx.config.operatorKey) {
+                  this.logger.info(`Using provided operator key: ${ctx.config.operatorKey}`);
```
jeromy-cannon, line 281: don't log keys or secrets

JeffreyDallas: removed

jeromy-cannon (approved):

Issue #1178, opened 2025-01-17
save reusable solo flags in RemoteConfig


jeromy-cannon:
We removed the local caching of flags in ConfigManager to avoid issues with multi-cluster support.  Now that we have RemoteConfig, it makes sense to store certain values in it to leverage for commands and subcommands for a better user experience.

- Add all of the flags that are re-usable across multiple subcommands that are needed for `network deploy` to the `deployment create` subcommand to store in the remote config: 
  - node aliases
  - release tag
  - solo chart version
  - (anything else?)
- During the `deployment create` subcommand, as well as the `network deploy` subcommand store common flag values in remote config for reuse (`specification: RemoteConfigSpecification`)
	- node aliases unparsed (`nodeAliasesUnparsed: string`, `getNodeAliases(): string[]`)
	- solo chart version (`soloChartVersion: string`)
	- hedera platform release tag (`hederaPlatformReleaseTag: string`)
	- chart directory (`soloChartsDirectory: string`)
	- local build path (`hederaPlatformLocalBuildPath: string`)
- update commands/subcommands that have use these re-usable flags so that:
	- if a user does not specify the flag value as an command line argument and the flag/value exists in the remote config, use the value from the remote config and do not prompt the user for a value.
	- if the value specified by the user does not match what is in the remote config, throw a warning telling the user that the remote config value and the specified value does not match, and display those values.
	    - Furthermore, if the force flag AND quiet flag is not set, then prompt the user if they wish to continue with the specified values or the remote config values.

NOTE: this could be broken up into multiple separate issues for smaller PRs, or have multiple smaller PRs relate to this single GitHub Issue.

PR #1191, opened 2025-01-21
feat(remote-config): save reusable solo flags in remoteconfig

instamenta:
## Description

adds new class `CommonFlagsDataWrapper` to handle keeping the state of the flags inside the remote config, it's added to the middleware which loads and and it will populate the flags if needed.

### Related Issues

* Closes [#1178](https://github.com/hashgraph/solo/issues/1178)


package.json
```diff
@@ -48,6 +48,7 @@
   "license": "Apache2.0",
   "dependencies": {
     "@hashgraph/sdk": "^2.56.0",
+    "@inquirer/prompts": "^7.2.3",
```
jeromy-cannon, line 51: I see that there is some confusion in our package.json that has probably been there a while.

We have: 
- "@listr2/prompt-adapter-enquirer": "^2.0.12",
- "enquirer": "^2.4.1",
- "inquirer": "^12.3.2",
- and now:  "@inquirer/prompts": "^7.2.3",

It looks like `enquirer` is really old.  `inquirer` is newer, but has the legacy post on their GitHub site.  `@inquirer/prompts` is the newest and should probably be used.

Would you see if we can remove:
- "@listr2/prompt-adapter-enquirer": "^2.0.12",
- "enquirer": "^2.4.1",
- "inquirer": "^12.3.2",

And add: `@listr2/prompt-adapter-inquirer`

If it works, this is much better than having all of those other libraries that will cause confusion.

src/core/config/remote/common_flags_data_wrapper.ts
```diff
@@ -0,0 +1,118 @@
...
+        }
+      }
+    };
+
+    // if the flag is set, inspect the value
+    if (this.configManager.hasFlag(flag)) {
+      await detectFlagMismatch();
+    }
+
+    // if the value is not set and exists, override it
```
jeromy-cannon, line 92:
```suggestion
    // use remote config value if no user supplied value
```

src/core/config/remote/remote_config_data_wrapper.ts
```diff
@@ -39,13 +33,15 @@ export class RemoteConfigDataWrapper implements Validate, ToObject<RemoteConfigD
   private _components: ComponentsDataWrapper;
   private _commandHistory: string[];
   private _lastExecutedCommand: string;
+  private readonly _flags: CommonFlagsDataWrapper;
```
jeromy-cannon, line 36:
```suggestion
  private readonly flags: CommonFlagsDataWrapper;
```
https://google.github.io/styleguide/tsguide.html#naming-style : Do not use trailing or leading underscores for private properties or methods.

src/core/config/remote/remote_config_data_wrapper.ts
```diff
@@ -113,22 +109,26 @@ export class RemoteConfigDataWrapper implements Validate, ToObject<RemoteConfigD
...
+  public get flags() {
+    return this._flags;
+  }
+
   //! -------- Utilities -------- //
 
   public static compare(x: RemoteConfigDataWrapper, y: RemoteConfigDataWrapper): boolean {
-    // TODO
-    return true;
+    return true; // TODO
```
jeromy-cannon, line 119: TODO

src/core/config/remote/common_flags_data_wrapper.ts
```diff
@@ -0,0 +1,118 @@
...
+    else if (this.flags[flag.constName]) {
+      argv[flag.constName] = this.flags[flag.constName];
+      this.configManager.setFlag(flag, this.flags[flag.constName]);
+    }
+  }
+
+  public static async initializeEmpty(configManager: ConfigManager, argv: AnyObject): Promise<CommonFlagsDataWrapper> {
```
jeromy-cannon, line 105:
```suggestion
  public static async initialize(configManager: ConfigManager, argv: AnyObject): Promise<CommonFlagsDataWrapper> {
```
initializeEmpty is misleading, as it is calling handle flags which updates the flags.

src/core/config/remote/remote_config_data_wrapper.ts
instamenta: With Nathan we discussed and agreed that values that have a getter `get flags()` we will use _underscore prefix in naming convention, this is same as for the hedera javascript sdk

src/core/config/remote/remote_config_manager.ts
```diff
@@ -212,10 +215,69 @@ export class RemoteConfigManager {
...
+   *
+   * @param argv - arguments containing command input for historical reference.
+   * @returns a Listr task which loads the remote configuration.
+   */
+  public buildLoadTask(argv: {_: string[]}): SoloListrTask<EmptyContextConfig> {
+    const self = this;
+
+    return {
+      title: 'Load remote config',
+      task: async (): Promise<void> => {
```
jeromy-cannon, line 234: anything that returns listr tasks belongs in `listr_config_tasks.ts`, except for the business logic.

src/core/config/remote/remote_config_manager.ts
```diff
@@ -212,10 +215,69 @@ export class RemoteConfigManager {
...
+    cluster: Cluster,
+    context: Context,
+    namespace: Namespace,
+    argv: AnyObject,
+  ): SoloListrTask<ListrContext> {
+    const self = this;
+
+    return {
+      title: `Create remote config in cluster: ${cluster}`,
+      task: async (_, task): Promise<void> => {
```
jeromy-cannon, line 256: anything that returns listr tasks belongs in listr_config_tasks.ts, except for the business logic.

test/unit/core/remote_config/remote_config_data_wrapper.test.ts
```diff
@@ -34,13 +42,15 @@ function createRemoteConfigDataWrapper() {
   const components = componentsDataWrapper;
   const lastExecutedCommand = 'lastExecutedCommand';
   const commandHistory = [];
+  const flags = await CommonFlagsDataWrapper.initializeEmpty(configManagerMock as any, {}); // TODO MOCK
```
jeromy-cannon: TODO

package.json
instamenta: New issue created to address the issue with the libs
https://github.com/hashgraph/solo/issues/1201

jeromy-cannon (approved):

PR #1193, opened 2025-01-21
feat: Add adminKey flag to node setup command

### LINKED ISSUE #1042: add flags to supply keys to use for node admin keys when creating `genesis-network.json` file

Ivo-Yankov:
### Related Issues

* Closes #1042

### DIFF
--- a/src/commands/flags.ts
+++ b/src/commands/flags.ts
@@ -1733,6 +1733,7 @@ export class Flags {
 
   static readonly allFlags: CommandFlag[] = [
     Flags.accountId,
+    Flags.adminKey,
     Flags.amount,
     Flags.apiPermissionProperties,
     Flags.app,
--- a/src/commands/flags.ts
+++ b/src/commands/flags.ts
@@ -1351,6 +1351,17 @@ export class Flags {
     prompt: undefined,
   };
 
+  static readonly adminPublicKeys: CommandFlag = {
+    constName: 'adminPublicKeys',
+    name: 'admin-public-keys',
+    definition: {
+      describe: 'Comma separated list of DER encoded ED25519 public keys',
+      defaultValue: constants.GENESIS_KEY,
+      type: 'string',
+    },
+    prompt: undefined,
+  };
+
   static readonly quiet: CommandFlag = {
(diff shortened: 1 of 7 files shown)


jeromy-cannon (changes requested):
@Ivo-Yankov ,

A few questions/thoughts:

1. As someone using Solo, how do I provide a different key for each node?  When looking at the code I see that you allow a comma separated list of keys.  However, when looking at the `flags.ts` code, I only see the description: `Admin key`.  I think we need to provide a better description and provide a usage example.
2. Since we don't sign anything, I think it would be more friendly to just supply the public key. It will be smaller, easier to supply on the command line, and will be safer, since the private key isn't really needed here.
3. rename the flag from `--admin-key` to `--admin-public-keys`

Ivo-Yankov:
> A few questions/thoughts:
> 
> 1. As someone using Solo, how do I provide a different key for each node?  When looking at the code I see that you allow a comma separated list of keys.  However, when looking at the `flags.ts` code, I only see the description: `Admin key`.  I think we need to provide a better description and provide a usage example.
> 2. Since we don't sign anything, I think it would be more friendly to just supply the public key. It will be smaller, easier to supply on the command line, and will be safer, since the private key isn't really needed here.
> 3. rename the flag from `--admin-key` to `--admin-public-keys`

@jeromy-cannon I addressed points 2 and 3 and I think that mostly covers point 1. But I'm not sure on what is the appropriate way to provide a usage example.

src/commands/flags.ts
```diff
@@ -1351,6 +1351,17 @@ export class Flags {
     prompt: undefined,
   };
 
+  static readonly adminPublicKeys: CommandFlag = {
+    constName: 'adminPublicKeys',
+    name: 'admin-public-keys',
+    definition: {
+      describe: 'Comma separated list of DER encoded ED25519 public keys',
```
jeromy-cannon, line 1358:
```suggestion
      describe: 'Comma separated list of DER encoded ED25519 public keys and must match the order of the node aliases',
```

jeromy-cannon (approved):

PR #1205, opened 2025-01-24
fix: Refactor task and handler classes

Ivo-Yankov:
## Description

* TBD

### Related Issues

* Closes #


src/commands/cluster/index.ts
```diff
@@ -32,7 +32,8 @@ export class ClusterCommand extends BaseCommand {
   constructor(opts: Opts) {
     super(opts);
 
-    this.handlers = new ClusterCommandHandlers(this, new ClusterCommandTasks(this, this.k8), this.remoteConfigManager);
+    // TODO properly inject
```
jeromy-cannon, line 35: TODO

src/commands/node/index.ts
```diff
@@ -51,48 +49,18 @@ export class NodeCommand extends BaseCommand {
...
-      platformInstaller: opts.platformInstaller,
-      logger: opts.logger,
-      k8: opts.k8,
-      tasks: this.tasks,
-      parent: this,
-      leaseManager: opts.leaseManager,
-      remoteConfigManager: opts.remoteConfigManager,
-    });
+
+    // TODO properly inject
```
jeromy-cannon, line 53: TODO

src/core/command_handler.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  constructor(
+    @inject(SoloLogger) public readonly logger?: SoloLogger,
+    @inject(ConfigManager) private readonly configManager?: ConfigManager,
+  ) {
+    this.logger = patchInject(logger, SoloLogger, this.constructor.name);
+    this.configManager = patchInject(configManager, ConfigManager, this.constructor.name);
+  }
+
+  commandActionBuilder(actionTasks: any, options: any, errorString: string, lease: Lease | null) {
```
jeromy-cannon, line 42: add scope modifier public to method, add return type

src/core/command_handler.ts
```diff
@@ -0,0 +1,107 @@
...
+        await Promise.all(promises);
+      }
+    };
+  }
+
+  /**
+   * Setup home directories
+   * @param dirs a list of directories that need to be created in sequence
+   */
+  setupHomeDirectory(
```
jeromy-cannon, line 68: add scope modifier public to method, add return type

src/core/command_handler.ts
```diff
@@ -0,0 +1,107 @@
...
+      });
+    } catch (e: Error | any) {
+      self.logger.error(e);
+      throw new SoloError(`failed to create directory: ${e.message}`, e);
+    }
+
+    return dirs;
+  }
+
+  setupHomeDirectoryTask() {
```
jeromy-cannon, line 93: add scope modifier public to method, add return type

src/core/command_handler.ts
```diff
@@ -0,0 +1,107 @@
...
+    return new Task('Setup home directory', async () => {
+      this.setupHomeDirectory();
+    });
+  }
+
+  // Config related methods:
+  getConfig(configName: string, flags: CommandFlag[], extraProperties: string[] = []): object {
```
jeromy-cannon, line 100: add scope modifier public to method, add return type

test/unit/commands/cluster.test.ts
```diff
@@ -228,16 +241,26 @@ describe('ClusterCommand unit tests', () => {
         certificateManager: sandbox.createStubInstance(CertificateManager),
         remoteConfigManager: remoteConfigManagerStub,
       } as Opts;
+
+      // stubDependencies([Object.values(options)]);
```
jeromy-cannon, line 245: delete commented out code

test/unit/commands/cluster.test.ts
```diff
@@ -228,16 +241,26 @@ describe('ClusterCommand unit tests', () => {
...
+
+      return options;
     };
 
+    function stubDependencies(deps: any[]) {
```
jeromy-cannon, line 250: is this still being used if it is commented out on line 245?

src/commands/base.ts
```diff
@@ -197,86 +159,13 @@ export abstract class BaseCommand extends ShellRunner {
...
-    };
-
-    const newConfigInstance = new NewConfigClass();
-
-    // add the new instance to the configMaps so that it can be used to get the
-    // unused configurations using the configName from the BaseCommand
-    this._configMaps.set(configName, newConfigInstance);
-
-    return newConfigInstance;
+    return getConfig(this.configManager, this._configMaps, configName, flags, extraProperties);
```
jeromy-cannon, line 168: This is really confusing.  upon first glance I thought you were doing recursion.  I see you are importing a function with the same name as the function calling it.  I think it would be better to rename one of them to avoid confusion.

src/core/container_init.ts
jeromy-cannon, line 1: what is the purpose of this file?

src/commands/node/configs.ts
```diff
@@ -31,406 +40,436 @@ export const KEYS_CONFIGS_NAME = 'keyConfigs';
...
-    deployment: this.configManager.getFlag(flags.deployment),
-    consensusNodes: this.parent.getConsensusNodes(),
-    contexts: this.parent.getContexts(),
-  };
-
-  if (!(await this.k8Factory.default().namespaces().has(ctx.config.namespace))) {
-    throw new SoloError(`namespace ${ctx.config.namespace} does not exist`);
+
+  public async keysConfigBuilder(argv, ctx, task, configMaps?: ConfigMap) {
+    const config: NodeKeysConfigClass = getConfig(this.configManager, configMaps, KEYS_CONFIGS_NAME, argv.flags, [
```
jeromy-cannon, line 384: argv is an array.  But, it has been mutated and had objects stored on it.  We really should keep it pure and not mutate it.  Instead coming up with another way to pass the values instead of reusing argv for that purpose.

src/core/container_init.ts
Ivo-Yankov: I wanted to make `getConfig` accessible to classes that are not extending BaseCommand. But actually maybe it makes more sense to move it to `configManager` and access it through an injection instead of including it like a helper function.

jeromy-cannon: I think the way you have it as an export function is fine right now.  If this file is still empty, then just delete it.

jeromy-cannon: also, we plan on getting rid of that getConfig function.  It was temporary when I was doing a lot of refactoring for the deterministic image

jeromy-cannon: I see you moved it into configManager.  That is fine for now.

jeromy-cannon (dismissed):

jeromy-cannon (dismissed):

jeromy-cannon (approved):

Issue #1052, opened 2024-12-31
enhance logging to include a command with banner so that you can see in the logs when a new command is ran and what was passed, mask any sensitive flag values


jeromy-cannon:
for masking see the related issue that added the `dataMask` property to specific flags.

- relates to https://github.com/hashgraph/solo/issues/960

Issue #960, opened 2024-12-09
Capture the full command line call, not just the command and subcommand, but all flags and their parameters as it was called


jeromy-cannon:
currently in the remote config history, it only tracks down to the sub command.  We want to track the full argv.  However, we want to provide a mask **** for certain private/sensitive values such as private keys or local directories.

Update the `CommandFlag` interface to add an optional `dataMask: string` property.  In `flags.ts` set the `dataMask` property to `***` for the following flags:
- `operatorKey`
- `privateKey`
- `gossipPrivateKey`
- `tlsPrivateKey`
- `ed25519PrivateKey`
- `ecdsaPrivateKey`
- `adminKey`
- `grpcTlsKeyPath`
- `grpcWebTlsKeyPath`
- `storageAccessKey`
- `storageSecrets`
- `storageEndpoint`
- `storageBucket`
- `backupBucket`
- `googleCredential`

## related issues

- relates to https://github.com/hashgraph/solo/issues/1052

PR #1208, opened 2025-01-24
feat(remote-config): instead of keeping just the command and the subcommand inside the remote config keep the flags as passed

instamenta:
## Description
* add new function to stringify the flags from the argv and to store it together with the command and subcommand inside the remote config's `commandHistory: string[]`
* add email address of the user that executed the command next to the command inside the history

```yaml
# Example of new structure
commandHistory:                                                                                              
 - "Executed by example@test.com: cluster setup"
 - "Executed by example@test.com: network deploy --node-aliases node1"
```

* Add the stringified flags together with the command and the subcommand inside the logging when command is ran
```yaml
# Example of log
Version                 : 0.34.0
Kubernetes Context      : kind-solo-e2e
Kubernetes Cluster      : kind-solo-e2e
Current Command         : node keys --gossip-keys --tls-keys --node-aliases node1
```

* rename the `flag.nodeAliasesUnparsed` cli name from `node-aliases-unparsed` --> `node-aliases`
* mask user supplied values that are senstive/private before writing to RemoteConfig

### Related Issues

* Closes [#960](https://github.com/hashgraph/solo/issues/960)
* Closes [#1052](https://github.com/hashgraph/solo/issues/1052)
* Closes [#959](https://github.com/hashgraph/solo/issues/959)


src/core/config/remote/remote_config_validator.ts
```diff
@@ -19,7 +19,7 @@ import {SoloError} from '../../errors.js';
 
 import type {K8} from '../../k8.js';
 import type {ComponentsDataWrapper} from './components_data_wrapper.js';
-import {type BaseComponent} from './components/base_component.js';
+import type {BaseComponent} from './components/base_component.js';
```
jeromy-cannon, line 22: I think your settings might be different than mine.  How do we adjust the setting so that it is the same for everyone?  I don't really care which way the setting is, just want it consistent.

instamenta: i fixed it since all of the others were the other way, and it was bugging me. ps. I can stop doing it np

Issue #1226, opened 2025-01-28
enhance mirror node external database feature


jeromy-cannon:
During the milestone # 1 deployment when deploying Mirror Node using an external Database to GKE a lot of workarounds were applied.  This issue is to close some of the gaps and reduce the amount of scripting required to workaround these gaps.

```[tasklist]
### Tasks
- [x] Rename the flag `--custom-mirror-node-database-values-path` to `--use-external-database`. Set the description to: "Set to true if you have an external database to use instead of the database that the Mirror Node Helm chart supplies"
- [x] When `--use-external-database` is true write the following to the templated values file: `stackgres.enabled: false`, `postgresql.enabled: false`, `db.name: mirror_node`
- [x] When `--use-external-database` is true and the `database-seeding-query.sql` is written out, show a message to the user: "Please run the following SQL script against the external database to enable Mirror Node to function correctly: ${path.join(constants.SOLO_CACHE_DIR, 'database-seeding-query.sql')}"
- [x] Add a flag `--external-database-host` for the user to provide the external database host.  If it is set, then write the following to the templated values file: `db.host: ${externalDatabaseHost}`
- [x] Add a flag `--external-database-owner-username` for the user to provide the external database privileged owner username.  If it is set, then write the following to the templated values file: `db.owner.username: ${username}`, `importer.db.username: ${username}`, `grpc.db.username: ${username}`, `rest.db.username: ${username}`, `restjava.db.username: ${username}`, `web3.db.username: ${username}`
- [x] Add a flag `--external-database-owner-password` for the user to provide the external database privileged owner password.  If it is set, then write the following to the templated values file: `db.owner.password: ${password}`, `importer.db.password: ${password}`, `grpc.db.password: ${password}`, `rest.db.password: ${password}`, `restjava.db.password: ${password}`, `web3.db.password: ${password}`
- [ ] Update the "Test Taskfile Commands" workflow scripts to leverage the external database and the flags added to support it
- [ ] Update the `Taskfile.helper.yml` to be able to leverage environment variables from an `examples` directory `Taskfile.yml` to then enable the external database and pass in the values
- [ ] Create a `examples/custom-mirror-node-database/Taskfile.yml` to provide the necessary environment variables to be used for the updated "Test Taskfile Commands"
[...]

PR #1230, opened 2025-01-29
feat(mirror-node): Enhance mirror-node external database feature

instamenta:
## Description

**Deprecated** the flag `--custom-mirror-node-database-values-path`,
instead there is new strategy together with 3 new flags:
* `--use-external-database` - boolean flag that signalizes you want to enable the feature
* `--external-database-host` - provides the external database host
* `--external-database-owner-username` - provides the external database owner's username
* `--external-database-owner-password` - provides the external database owner's password

### Related Issues

* Closes [#1226](https://github.com/hashgraph/solo/issues/1226)


Taskfile.helper.yml
```diff
@@ -434,6 +434,15 @@ tasks:
           sleep 4
         fi
 
+  solo:mirror-node-only:
```
jeromy-cannon, line 437: instead of creating a new task.  Instead split the `solo:mirror-node` into:
- `solo:mirror-node`
- `solo:explorer`
keep the port forward logic on both, but only do the port forward for the relevant services

doing this will require:
- `default-with-mirror` - add `solo:explorer` to task list
- `default-with-relay` - add `solo:explorer` to task list between `solo:mirror-node` and `solo:relay`
- `examples/external-database-test/Taskfile.yml`:`install` - following `solo:mirror-node`

instamenta: this is now obsolete task, I will remove it

examples/external-database-test/scripts/solo_smoke_test.sh
jeromy-cannon, line 1: It looks like this is a clone of the other smoke test `.github/workflows/script/solo_smoke_test.sh`.  I would say to merge your changes into that version to preserve commit history, and since it is dependent on that `helper.sh` in the same folder. 

Update `.github/workflows/flow-task-test.yaml`:
original:
```
      - name: Run Example Task File Test with type ${{ matrix.type }}
        run: |
          export CONSENSUS_NODE_VERSION=v0.58.3
          SOLO_CLUSTER_NAME=solo-task-test-${{ matrix.type }}-${{ github.run_id }}-${{ github.run_attempt }} task default
          .github/workflows/script/solo_smoke_test.sh ${{ matrix.type }}
          task clean
```
new:
```

File: src/commands/mirror_node.ts [hunk 2]
```diff
@@ -132,6 +141,43 @@ export class MirrorNodeCommand extends BaseCommand {
...
+        'stackgres.enabled': false,
+        'postgresql.enabled': false,
+
+        // Set the host and name
+        'db.host': host,
+        'db.name': 'mirror_node',
+
+        // set the usernames
+        'db.owner.username': username,
+        'importer.db.username': username,
```
jeromy-cannon, line 164: you might want to double check Nathan's script, but I think these two are using the provisioner username, the rest are using the mirror username

jeromy-cannon (approved):

PR #1373, opened 2025-02-12
chore: account init to update node admin key

### LINKED ISSUE #935: Production Readiness: Update account init/rekey logic to also update the Node admin keys
NOTE: this may not be needed after #949, need to validate

JeffreyDallas:
## Description

* account init to update node admin key with NodeUpdateTransaction, then save to k8 secret
* load adminKey from k8 secret first other wise, load default genesis key

### Related Issues

* Closes #935


src/commands/account.ts
```diff
@@ -244,6 +244,20 @@ export class AccountCommand extends BaseCommand {
                     });
                   },
                 },
+                {
+                  title: 'Update node admin key',
+                  task: async ctx => {
+                    const newPrivateKey = PrivateKey.generateED25519();
+                    const genesisAccountId = AccountId.fromString(constants.FREEZE_ADMIN_ACCOUNT);
+                    // update genesis account key
+                    await self.accountManager.updateAccountKeys(
```
jeromy-cannon, line 253: this is not what was meant.  We need to update the Node Admin Keys.  I believe the only way to do this is via the NodeUpdateTransaction: https://github.com/hashgraph/solo/blob/929fdaa22faa1566fb5300d3d387649c0a7c2b12/src/commands/node/tasks.ts#L1438-L1438

jeromy-cannon (changes requested):
Need to use NodeUpdateTransaction to update each of the node's admin keys.

src/commands/account.ts
JeffreyDallas: Updated

jeromy-cannon (approved):

PR #1398, opened 2025-02-14
fix: corrected cache dir flag not working and added dual cluster e2e test

jeromy-cannon:
## Description

* started creating a Dual Cluster Full E2E test
* integrated into CICD workflows
* changed `solo node keys` to use `--deployment` instead of `--node-aliases`
* updated `index.ts` `main(argv)` to be async/await to increase stability in the application
* fixed some issues where the solo cache directory was not being used correctly
* added an optional parameter in `remote_config_manager.ts.loadAndValidate()` to disable the validation logic
* updated container init logic to leverage the `isInitialized` property correctly
* enhanced namespace delete method to loop until namespace is actually deleted to prevent issues in test cases
* changed `test_util.ts.testLogger` to `getTestLogger()` to prevent race condition when DI hasn't been loaded yet

### Related Issues

* Relates to #1554


src/commands/node/flags.ts
```diff
@@ -223,7 +223,13 @@ export const REFRESH_FLAGS = {
 export const KEYS_FLAGS = {
   requiredFlags: [flags.cacheDir, flags.generateGossipKeys, flags.generateTlsKeys, flags.deployment],
   requiredFlagsWithDisabledPrompt: [],
-  optionalFlags: [flags.devMode, flags.quiet, flags.nodeAliasesUnparsed],
+  optionalFlags: [
+    flags.devMode,
+    flags.quiet,
+    flags.nodeAliasesUnparsed,
+    // TODO remove namespace once the remote config manager is updated to pull the namespace from the local config
```
jeromy-cannon, line 230: can I remove this now?

instamenta (approved):

PR #1423, opened 2025-02-17
fix: resolve issues discovered during charlie sphere staging deployment testing

nathanklick:
## Description

* Fixes issues with node staking
* Fixes issues with node activeness checks
* Fixes internal IP handling for the `0.58.5` release
* Improves several error messages and user output


instamenta (approved):

src/commands/node/tasks.ts
```diff
@@ -355,6 +356,10 @@ export class NodeCommandTasks {
           title,
           i,
           status,
+          undefined,
+          undefined,
+          undefined,
+          context,
```
leninmehedy, line 362: Here `undefined` args probably is causing issues. I'll test separately.

instamenta: passing undefined like this, uses the default value of the param 
``` typescript
// Example
function log(txt: string = 'Empty') {
  console.log(txt);
}

log(undefined);
// this will log - Empty
```

nathanklick:
Superseded by:
- #1428 
- #1429 
- #1430 
- #1431 
- #1432 
- #1433

PR #1438, opened 2025-02-18
chore(test): Add argv wrapper class and use it in tests

### LINKED ISSUE #900: Improve tests by adding new wrapper for the Argv
### Description

Create new class to wrap the arguments passed to the cli, this should prepare us for future test improvements

instamenta:
## Description

* Modifies the test, bootstrap functions to avoid having to pass undefined for default values.
* Adds new wrapper class `Argv` for building the arguments passed to the cli

### Related Issues

* Closes #900


test/helpers/argv_wrapper.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  public setArg(flag: CommandFlag, value: any): void {
+    this.args[flag.name] = value;
+  }
+
+  public getArg<T>(flag: CommandFlag): T {
+    return this.args[flag.name];
+  }
+
+  public build(): AnyObject {
```
leninmehedy, line 28: `build` and `clone` methods are a bit confusing and not clear when to use each of this. 

How about renaming `build()` with `cloneArgs()` for clarity? Alternatively, we can remove `build` and only use `clone` everywhere and it would make it easier to remember how to pass `args` obj.

instamenta: well it build the Args, it wraps the data and also it can be cloned into another instance of the same class, I would say the naming is pretty intuitive, and in the future the methods would become vastly different from each-other  when more logic is simplified/abstracted into them, clone would become a lot more useful once we start moving towards building separate Argv for each command executed, instead of relying on the bootstrap helpers

instamenta: With this, I've tried to keep it simple, If more complex solutions are needed in the future, we shouldn't overengineer it from now

jeromy-cannon: build() is usually for the Builder pattern.  Maybe createArgv(), create(), or getArgv().  I'm fine with this for now so that we can get it merged, should be an easy rename later once we figure out what we want.

jeromy-cannon (approved):

PR #1546, opened 2025-02-25
chore: fix warnings in ClusterCommand

instamenta:
## Description

- added interfaces for `ClusterCommandTasks` and `ClusterCommandHandlers`
- removed usage of the listr2 task wrapper `Task`
- added types to have type-safety
- remove lint warnings

### Related Issues

* Closes # https://github.com/hashgraph/solo/issues/1542

### Awaits

* https://github.com/hashgraph/solo/pull/1205
* https://github.com/hashgraph/solo/pull/1601


src/commands/cluster/configs.ts
```diff
@@ -2,27 +2,35 @@
...
 
-export const connectConfigBuilder = async function (argv, ctx, task) {
-  const config = this.getConfig(CONNECT_CONFIGS_NAME, argv.flags, []) as ClusterConnectConfigClass;
-
-  // set config in the context for later tasks to use
-  ctx.config = config;
-
+export const connectConfigBuilder = async function (
+  this: IClusterCommandHandlers & CommandHandlers,
+  argv: AnyArgv,
```
jeromy-cannon, line 21: AnyArgv is a bit of trick to get around Typescript's warning.  Later on, we will need to remove this to properly fix what the true type is, or write it in such a way that any is not needed.  This was something Nathan and I was discussing.  You don't have to change this now, but our preference is to just leave it as `any` for now and it will be easy to search and find those warnings.

jeromy-cannon (approved):

src/commands/cluster/configs.ts
instamenta: if we do decide to put a type on Argv it would still be easier to find all the places where AnyArgv is used and just change it, Wouldn't it?

JeffreyDallas (dismissed):

jeromy-cannon (approved):

PR #1563, opened 2025-02-27
ci: switch to different working directory for e2e test

### LINKED ISSUE #1534: update workflow job that does `npm install -g @hashgraph/solo` to purposefully switch into a different directory
one of the tests that does npm install -g and npm link but purposefully switch into a different directory other than solo root dir.  this would have revealed an error that we missed during one of our deployments in the past.

JeffreyDallas:
## Description

* switch to different working directory when running Solo after `npm install -g` to verify we are directory independent when the user installs Solo from npmjs

### Related Issues

* Closes #1534


jeromy-cannon (changes requested):
Nice idea.  But, I think this needs to be on the update readme job, since it is the only one that runs off of a `npm install -g @hashgraph/solo` + `npm link`.  Unless there is another one doing that, also.

PR #1584, opened 2025-03-03
chore: Enhance Grammar and Clarity in Solo Documentation

SITADRITA1:
Original:
Below we describe how you can set up local environment and contribute to `solo`.
Updated:
Below we describe how you can set up a local environment and contribute to `solo`.
Reason for Change:
Added "a" before "local environment" because "environment" is a singular countable noun, which requires an article.
This change improves grammatical correctness and readability.
Original:
with `./gradlew assemble`. If need to running nodes with different versions or releases, please duplicate the repo or build directories in
Updated:
with `./gradlew assemble`. If need to run nodes with different versions or releases, please duplicate the repo or build directories in
Reason for Change:
Changed "running" → "run" to correct the verb form. The phrase should use the base verb "run" rather than the continuous form "running" after "need to".
The previous phrasing was grammatically incorrect and needed correction for proper English usage.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -1,6 +1,6 @@
 # Instructions for developers working on solo project
 
-Below we describe how you can set up local environment and contribute to `solo`.
+Below we describe how you can set up a local environment and contribute to `solo`.
 
 * Clone the repo
 * In order to support ES6 modules with `jest`, set an env variable `NODE_OPTIONS` as below:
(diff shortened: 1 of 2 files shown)


SITADRITA1:
If my corrections are not correct, please let me know, and I will fix them, thank you)

jeromy-cannon (approved):

PR #1612, opened 2025-03-07
feat: Improve error handling

### LINKED ISSUE #1505: correct error handling
we are missing try/catch blocks with many calls. When doing an `async/await` if there is not a try/catch or a catch that is not close enough to the throw it can lead to scenarios that totally crashes NodeJS without a correct log error message.  We need to appropriately wrap `await` calls with try/catch blocks.  Additionally, catch blocks like this can fail destruction and throw null pointer exceptions: `.catch(e => e)`

### LINKED ISSUE #1585: Process does not terminate after logAndExitError and logAndExitSuccess
### Describe the bug

After calling `logger.logAndExitSuccess` or `logger.logAndExitError` the command continues execution instead of terminating with the provided message.

When those methods are called they should terminate the execution of the current command.

### To Reproduce

```
ivoyankov@Ivos-MacBook-Pro solo % npm run solo -- cluster connect --cluster-ref ivo-1 --context kind-solo-e2e-c1

> @hashgraph/solo@0.35.0 solo
> node --no-deprecation --no-warnings dist/solo.js cluster connect --cluster-ref ivo-1 --context kind-solo-e2e-c1

*********************************** ERROR *****************************************
Please create a local configuration first. Run "solo deployment create"
***********************************************************************************

Version			: 0.35.0
Kubernetes Context	: kind-solo-e2e-c1
Kubernetes Cluster	: kind-solo-e2e-c1
Current Command		: cluster connect --cluster-ref ivo-1 --context kind-solo-e2e-c1
✔ Initialize
✔ Associate a context with a cluster reference: ivo-1
✔ Test connection to cluster: ivo-1
✔ Save local configuration

```

The error is shown and the command was executed.

### Additional Context

_No response_

Ivo-Yankov:
## Description

- Adds an `ErrorHandler` class
- Removes duplicate logging of errors
- Fixes an issue where `logErrorAndExit` and `logSuccessAndExit` were not actually terminating the process as intended
- Removes the possibility of `cannot access property of undefined` due to some errors being returned as a result instead of being thrown
- fixes the `--dev` flag to correctly set the `devMode` property of the logger. When passing `--dev` you will now see the error stack trace in the console.

### Related Issues

* Closes #1585
* Closes #1505

### DIFF
--- a/src/commands/account.ts
+++ b/src/commands/account.ts
@@ -117,6 +117,7 @@ export class AccountCommand extends BaseCommand {
         const privateKey = PrivateKey.fromStringDer(newAccountInfo.privateKey);
         newAccountInfo.privateKeyRaw = privateKey.toStringRaw();
       } catch {
+        // TODO should this error be thrown or just logged?
         this.logger.error(`failed to retrieve EVM address for accountId ${newAccountInfo.accountId}`);
       }
     }
@@ -169,8 +170,7 @@ export class AccountCommand extends BaseCommand {
           ctx.accountInfo.privateKey,
         ))
       ) {
-        this.logger.error(`failed to update account keys for accountId ${ctx.accountInfo.accountId}`);
-        return false;
+        throw new SoloError(`failed to update account keys for accountId ${ctx.accountInfo.accountId}`);
       }
     } else {
       amount = amount || (flags.amount.definition.defaultValue as number);
@@ -183,8 +183,7 @@ export class AccountCommand extends BaseCommand {
 
     if (hbarAmount > 0) {
       if (!(await this.transferAmountFromOperator(ctx.accountInfo.accountId, hbarAmount))) {
-        this.logger.error(`failed to transfer amount for accountId ${ctx.accountInfo.accountId}`);
-        return false;
+        throw new SoloError(`failed to transfer amount for accountId ${ctx.accountInfo.accountId}`);
       }
       this.logger.debug(`sent transfer amount for account ${ctx.accountInfo.accountId}`);
     }
@@ -731,7 +730,6 @@ export class AccountCommand extends BaseCommand {
                   if (!r) throw new SoloError('Error running init, expected return value to be true');
                 })
                 .catch(err => {
-                  self.logger.showUserError(err);
                   throw new SoloError(`Error running init: ${err.message}`, err);
                 });
             },
@@ -751,7 +749,6 @@ export class AccountCommand extends BaseCommand {
                   if (!r) throw new SoloError('Error running create, expected return value to be true');
                 })
                 .catch(err => {
-                  self.logger.showUserError(err);
                   throw new SoloError(`Error running create: ${err.message}`, err);
                 });
             },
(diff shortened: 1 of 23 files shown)


src/core/logging.ts
```diff
@@ -156,25 +161,6 @@ export class SoloLogger {
     this.showUser(chalk.green('-------------------------------------------------------------------------------'));
     console.log(JSON.stringify(obj, null, ' '));
   }
-
-  logAndExitSuccess(msg: string, ...args: any) {
-    this.winstonLogger.log('info', msg, args, () => {
```
jeromy-cannon, line 161: This is how I forced the log to flush before the application stops, without it, how are you ensuring that we will get the last bit of logs in memory?

src/commands/account.ts
```diff
@@ -117,6 +117,7 @@ export class AccountCommand extends BaseCommand {
         const privateKey = PrivateKey.fromStringDer(newAccountInfo.privateKey);
         newAccountInfo.privateKeyRaw = privateKey.toStringRaw();
       } catch {
+        // TODO should this error be thrown or just logged?
```
jeromy-cannon, line 120: probably should throw an error, if it is unable to correctly convert the key, then I don't see what good it will be if we return it, probably will just throw an error in a different location.

src/commands/init.ts
```diff
@@ -100,11 +100,7 @@ export class InitCommand extends BaseCommand {
       },
     );
 
-    try {
-      await tasks.run();
-    } catch (e: Error | any) {
-      throw new SoloError('Error running init', e);
-    }
+    await tasks.run();
```
jeromy-cannon, line 103: I'm hesitant on this one.  as I mentioned earlier, having a try/catch as close the the await as possible prevents issues in NodeJS.  By doing this, what is catching the error, Yargs?  and, then it re-throws it again, so it is being thrown several levels.

src/core/errors.ts
```diff
@@ -88,3 +88,14 @@ export class DataValidationError extends SoloError {
     super(message, cause, {expected, found});
   }
 }
+
+export class UserBreak extends SoloError {
```
jeromy-cannon, line 92: every class export should have its own file

src/commands/node/tasks.ts
```diff
@@ -438,15 +439,15 @@ export class NodeCommandTasks {
...
+          throw new SoloError('empty response'); // Guard
         }
 
         const statusLine = response.split('\n').find(line => line.startsWith('platform_PlatformStatus'));
 
         if (!statusLine) {
           task.title = `${title} - status ${chalk.yellow('STARTING')}, attempt: ${chalk.blueBright(`${attempt}/${maxAttempts}`)}`;
           clearTimeout(timeoutId);
-          throw new Error('missing status line'); // Guard
+          throw new SoloError('missing status line'); // Guard
```
instamenta, line 450: This is not really a solo error, but I'm fine with that ( just a note )

instamenta (approved):

jeromy-cannon (approved):

jeromy-cannon (approved):

Issue #1463, opened 2025-02-24
implement 'DataManager' and use it throughout the codebase


jan-milenkov:
** DRAFT **

### Description

Currently we have a lot of logic spread out throughout the codebase, a lot of places utilize different methods to get the same data and it has become hard for getting the data since it's spread and comes from different places. For this reason we need to make a new class `DataManager` which would serve as a single source of truth.

### Class Interface

```typescript
export abstract class IDataManager {  
  public namespace: NamespaceName;  
  public deployment: DeploymentName;  
  
  public contextNames: ContextName[];  
  public clusterRefs: ClusterRef[];  
  public contextClusterMapping: Record<ContextName, ClusterRef>;  
  public clusterContextMapping: Record<ClusterRef, ContextName>;  
  
  public abstract nodes: ConsensusNode[];  
  public abstract existingNodes: ConsensusNode[];  
  public abstract nodeAliases: NodeAlias[];  
  public abstract existingNodeAliases: NodeAlias[];  
  
  public abstract leases: Lease[];  
  
  public abstract config: AnyObject;  
  
  /** To make it extensible and for type-safety, this method would be used when config data is needed */  
  public abstract getConfig<T>(): T;  
  
  public abstract readonly argv: AnyObject;  
  
  /** Flags that come from default values instead of passed via cli */  
  public abstract defaultFlags: CommandFlag[];  
}
```

jeromy-cannon:
I had a discussion with Nathan yesterday about our Data Manager.  Not specifically what you have hear, this is the first time I saw it.

We believe this is a good way to split up the data types:
- generated persisted data
- constants (real constants, not the ones that can be overridden by environment variables)
- configuration
  - non-persisted
  - persisted
    - from environment variables with defaults (shouldn't be constants) (persisted and non-persisted)
    - persisted by being written to cache
    - persisted by being: saved to remote config, cluster, or local config
- run-time
  - introspected (from a cluster or other source)
  - computed

jeromy-cannon:
More notes:
- argv should not be mutated, see helper.ts.addFlagsToArgv
- argv should not be mutated, see references to how baseCmd.getConfig() has a 2nd paramater of `argv.flags`
- also, we should remove baseCmd.getConfig()

PR #1655, opened 2025-03-14
feat: adds migration and full schema support for the local config

nathanklick:
## Description

* Fixes the `package.json` version to match the currently published `v0.35.1` release
* Adds a completed migration implementation for `LocalConfig`
* Refactors the `soloVersion` field in `LocalConfig`

### Related Issues

* Partially Addresses #1463

### DIFF
--- a/src/data/schema/migration/impl/local/local-config-v1-migration.ts
+++ b/src/data/schema/migration/impl/local/local-config-v1-migration.ts
@@ -3,7 +3,18 @@
 import {type SchemaMigration} from '../../api/schema-migration.js';
 import {VersionRange} from '../../../../../business/utils/version-range.js';
 import {Version} from '../../../../../business/utils/version.js';
-import {deepClone} from 'deep-clone';
+import deepClone from 'deep-clone';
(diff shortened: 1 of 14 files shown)


instamenta (approved):

src/data/schema/migration/impl/local/local-config-v1-migration.ts
```diff
@@ -17,17 +28,47 @@ export class LocalConfigV1Migration implements SchemaMigration {
...
     if (clone.schemaVersion && clone.schemaVersion !== 0) {
-      //this case should never happen considering the field was not present in version 0 and should default to zero
+      // this case should never happen considering the field was not present in version 0 and should default to zero
       // during this migration
-      throw Error();
+      throw new InvalidSchemaVersionError(clone.schemaVersion, 0);
     }
 
+    // Remove the legacy email address
+    delete clone.userEmailAddress;
```
jeromy-cannon, line 43: are you just calling this out to be more obvious?  it seems that with our current configuration it will just not use properties that are not in the target class with class-transformer.

jeromy-cannon (approved):

PR #1658, opened 2025-03-14
feat: local config cleanup maintenance

### LINKED ISSUE #1498: local-config cleanup maintenance
- Update `solo network destroy` to also remove the deployment from `LocalConfig` and should remove the remote configs on all of the clusters.
- Also, on `solo network destroy` remove the `--force` switch requirement from the delete of the namespace.  if the pvcs and secret delete switches are enabled, that is sufficient.
- Update the `solo network destroy` help to tell the user if the delete pvcs and delete secrets switch is set the namespace and anything in it will also be destroyed.  Put a similar statement in both the delete pvcs and delete secrets switches definition.
- Add a new command `solo deployment delete --deployment X` - Deletes the selected deployment from the LocalConfig.  Check if we have a running deployment on any of the cluster-refs: if (namespace exists AND (configMap with label= 'app.kubernetes.io/managed-by: Helm' OR remote config exists)) then the command should fail and let the user know it found existing resources which need to be removed first.
- `cluster-ref connect` should NOT trigger the RemoteConfig validation logic, but we should verify the connectivity to the cluster that we are adding the cluster-ref to context mapping for.

Also, try to be aware of the ongoing changes in this Issue: https://github.com/hashgraph/solo/issues/1471

- relates to #1471

Ivo-Yankov:
## Description

- Adds a new command `solo deployment delete [--deployment xxx]` which deletes a deployment from the `LocalConfig` if there are no existing remote resources for that deployment.
- Adds a new step in the `solo network destroy` command, which deletes the deployment data from `LocalConfig` and deletes the `RemoteConfig` ConfigMaps in all deployment clusters. 
- Creates the `LocalConfig` during `solo init`, also adds an optional `--user-email-address` flag to the command.
- Adds a middleware that checks if the `LocalConfig` exists and throws an error if it doesn't, notifying the user to run `solo init` first.

### Related Issues

* Closes #1498


instamenta (dismissed):

src/core/middlewares.ts
```diff
@@ -147,4 +151,29 @@ export class Middlewares {
       return argv;
     };
   }
+
+  /**
+   * Checks if the Solo instance has been initiated
+   *
+   * @returns callback function to be executed from listr
+   */
+  public checkIfInitiated() {
```
jeromy-cannon, line 160:
```suggestion
  public checkIfInitialized() {
```

src/core/middlewares.ts
```diff
@@ -147,4 +151,29 @@ export class Middlewares {
       return argv;
     };
   }
+
+  /**
+   * Checks if the Solo instance has been initiated
```
jeromy-cannon, line 156:
```suggestion
   * Checks if the Solo instance has been initialized
```

src/index.ts
```diff
@@ -117,8 +117,16 @@ export async function main(argv: string[], context?: {logger: SoloLogger}) {
...
     .middleware(
-      // @ts-expect-error - TS2322: To assign middlewares
-      [middlewares.setLoggerDevFlag(), middlewares.processArgumentsAndDisplayHeader(), middlewares.loadRemoteConfig()],
+      [
+        // @ts-expect-error - TS2322: To assign middlewares
+        middlewares.setLoggerDevFlag(),
+        // @ts-expect-error - TS2322: To assign middlewares
+        middlewares.processArgumentsAndDisplayHeader(),
+        // @ts-expect-error - TS2322: To assign middlewares
+        middlewares.checkIfInitiated(),
```
jeromy-cannon, line 126:
```suggestion
        middlewares.checkIfInitialized(),
```

jeromy-cannon (dismissed):

jeromy-cannon (approved):

PR #1664, opened 2025-03-17
fix(multi-cluster): make remote config features required

instamenta:
## Description

* Removes conditional logic assuming the remote config and its components might be missing
* Fixes bug when prompting for `--num-consensus-nodes` has no effect 

### Related Issues

* Closes # https://github.com/hashgraph/solo/issues/1663

### DIFF
--- a/src/commands/flags.ts
+++ b/src/commands/flags.ts
@@ -19,6 +19,7 @@ import validator from 'validator';
 import {type AnyListrContext, type AnyObject, type AnyYargs} from '../types/aliases.js';
 import {type ClusterRef} from '../core/config/remote/types.js';
 import {type Optional, type SoloListrTaskWrapper} from '../types/index.js';
+import chalk from 'chalk';
 
 export class Flags {
   public static KEY_COMMON = '_COMMON_';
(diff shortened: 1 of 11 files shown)


Ivo-Yankov (approved):

src/commands/flags.ts
```diff
@@ -2240,7 +2241,25 @@ export class Flags {
       describe: 'Used to specify desired number of consensus nodes for pre-genesis deployments',
       type: 'number',
     },
-    prompt: undefined,
+    prompt: async function promptAmount(task: SoloListrTaskWrapper<AnyListrContext>, input: number): Promise<number> {
```
jeromy-cannon, line 2244:
```suggestion
    prompt: async function promptNumberOfConsensusNodes(task: SoloListrTaskWrapper<AnyListrContext>, input: number): Promise<number> {
```
There is actually another flag titled `amount`, changing the name to avoid confusion.

src/commands/flags.ts
```diff
@@ -2240,7 +2241,25 @@ export class Flags {
...
+      const promptForInput = (): Promise<number> =>
+        Flags.prompt(
+          'number',
+          task,
+          input,
+          Flags.numberOfConsensusNodes.definition.defaultValue,
+          `Enter number of consensus nodes ${chalk.grey('(must be a positive number)')}:`,
```
jeromy-cannon, line 2251:
```suggestion
          `Enter number of consensus nodes to add to the provided cluster ${chalk.grey('(must be a positive number)')}:`,
```

src/commands/network.ts
```diff
@@ -633,6 +633,7 @@ export class NetworkCommand extends BaseCommand {
       flags.gcsEndpoint,
       flags.gcsBucket,
       flags.gcsBucketPrefix,
+      flags.nodeAliasesUnparsed,
```
jeromy-cannon, line 636: why is this being added back?

instamenta: because we don't want to prompt the user so it's disabled

src/core/config/remote/remote-config-manager.ts
```diff
@@ -230,52 +217,40 @@ export class RemoteConfigManager {
...
-      const newError = new SoloError('Failed to load remote config');
-      // TODO throw newError instead of showUserError()
-      self.logger.showUserError(newError);
-      return;
-    }
-    self.logger.info('Remote config loaded');
-    if (!validate) {
-      return;
-    }
+    await this.load();
```
jeromy-cannon, line 224: I believe this.load() throws an error, but there is no try/catch.

src/commands/network.ts
jeromy-cannon: ah, didn't see this was in the disabledPrompts, had to expand further.

jeromy-cannon (approved):

PR #1672, opened 2025-03-18
fix: fixed wrong counting when adding new nodes and the local config not being updated properly 

jan-milenkov:
## Description

* Fixes bug when new nodes have wrong count
* Fixes issue where local config is not updated properly 

### Related Issues

* Closes # https://github.com/hashgraph/solo/issues/1670


jeromy-cannon (changes requested):
still not working.  See the results from this: https://github.com/hashgraph/solo/pull/1674

basically, node1 and node2 are being deployed to both clusters.

JeffreyDallas (approved):

PR #1909, opened 2025-04-28
fix: node add and node delete in CLI mode

### LINKED ISSUE #1902: dataMask cause incorrect default value of adminKey
### Describe the bug

The below function changes the default value of any flag with `dataMask` to '***'
```
  public static setOptionalCommandFlags(y: AnyYargs, ...commandFlags: CommandFlag[]) {
    for (const flag of commandFlags) {
      let defaultValue = flag.definition.defaultValue === '' ? undefined : flag.definition.defaultValue;
      defaultValue = defaultValue && flag.definition.dataMask ? flag.definition.dataMask : defaultValue;
      y.option(flag.name, {
        ...flag.definition,
        default: defaultValue,
      });
    }
  }
```

Then causes the following command line failed, because adminKey value `***` is reported invalid PrivateKey

```
    context_.adminKey = argv[flags.adminKey.name]
      ? PrivateKey.fromStringED25519(argv[flags.adminKey.name])
      : PrivateKey.fromStringED25519(constants.GENESIS_KEY);
```

Output

```
solo node add --deployment "${SOLO_DEPLOYMENT}" --gossip-keys --tls-keys --debug-node-alias node4 --local-build-path ../hiero-consensus-node/hedera-node/data --pvcs true

Version			: 0.35.1
Kubernetes Context	: kind-solo-cluster
Kubernetes Cluster	: kind-solo-cluster
Current Command		: node add --deployment solo-deployment --gossip-keys --tls-keys --debug-node-alias node4 --local-build-path ../hiero-consensus-node/hedera-node/data --pvcs --admin-key ***
✖ invalid private key length: 0 bytes
◼ Check that PVCs are enabled
◼ Identify existing network nodes

```

This bug is not exposed in normal E2E test, since in E2E test this `setOptionalCommandFlags` function is
never invoked.

Should not fail

### To Reproduce

```
solo node add --deployment "${SOLO_DEPLOYMENT}" --gossip-keys --tls-keys --debug-node-alias node4 
```

### Additional Context

_No response_

### LINKED ISSUE #1907: could not create account after node delete operation
### Describe the bug

Could not run any create account operation after node delete
```
failed to load node client: failed to refresh node client: failed to setup node client: failed to configure node access: SoloError: failed to configure node access: failed to ping network node:
  127.0.0.1:30214

```

This is due to refreshClient would try to ping every node service, even for the deleted one.
This is because serviceMap was built by iterate every available node service, even a node is deleted, but
pod and its service still there.

should not fail

### To Reproduce

```
npm run solo-test -- node delete --deployment "${SOLO_DEPLOYMENT}" --node-alias node3
npm run solo-test --  account create --deployment solo-deployment --hbar-amount 100
```

### Additional Context

_No response_

### LINKED ISSUE #1910: node update after node delete would fail at step `Check all nodes are FROZEN`
### Describe the bug

although the delete node nodeId is being marked as `0.0.0`,
but at step `Check all nodes are FROZEN`, it would still check 
status of node1, node2, node3,node4,
This is because previous step `Identify existing network nodes` returns node1,node2,node3,node3
from function `getNodeServiceMap`, even node1 is already deleted, its pod still running

should not fail

### To Reproduce

```
  npm run solo-test -- node delete --deployment "${SOLO_DEPLOYMENT}" --node-alias node1

  npm run solo-test -- node update --deployment "${SOLO_DEPLOYMENT}" --node-alias node2 --new-account-number 0.0.7
```

### Additional Context

_No response_

JeffreyDallas:
## Description

* Fixed failed node add/delete test in cli mode
* Fixed a bug in `prepareValuesArgForNodeDelete`, where `maxNodeId + 1` is used as nodeId 
* add new workflow to test cli node add/delete/update make sure they are working

### Related Issues

* Closes #1907
* Closes #1910

### Testing

* TBD

The following was not tested:

* TBD

<details>
<summary>
Commit message guidelines
</summary>

</details>


.github/workflows/flow-gcs-test.yaml
```diff
@@ -40,7 +40,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        storageType: ["gcs_only", "minio_only", "aws_only"]
+        storageType: ["gcs_only", "minio_only", "aws_only", "node_add_delete"]
```
jeromy-cannon, line 43: We want to limit tests running outside of Mocha.  I understand that you are wanting to properly test the CLI.  The plan is to migrate the E2E tests to use the same methodology of `dual-cluster-full.test.ts`.  To create a new E2E test harness extracting reusable components and refactoring `dual-cluster-full.test.ts` to remain DRY.

Tests ran outside of Mocha will not show up in coverage reports.  Additionally, we want to avoid adding new E2E tests to avoid increasing our pipeline total run time.

I think the functionality here could easily be integrated into our existing E2E mocha tests and possibly removing the current individual node add, delete, update (leave the seperate versions).

To rebuild the E2E harness will take more work.  We probably need to choose a quicker solution to merge this PR, so that we can work the tests harness refactor in a separate future PR.

jeromy-cannon (changes requested):
we don't want to add more Flow GCS Test scenarios

.github/workflows/flow-gcs-test.yaml
JeffreyDallas: reverted the changes to workflow

src/commands/flags.ts
```diff
@@ -129,8 +129,7 @@ export class Flags {
    */
   public static setOptionalCommandFlags(y: AnyYargs, ...commandFlags: CommandFlag[]) {
     for (const flag of commandFlags) {
-      let defaultValue = flag.definition.defaultValue === '' ? undefined : flag.definition.defaultValue;
-      defaultValue = defaultValue && flag.definition.dataMask ? flag.definition.dataMask : defaultValue;
+      const defaultValue = flag.definition.defaultValue === '' ? undefined : flag.definition.defaultValue;
```
jeromy-cannon, line 132: @instamenta , looks like there was a bug in this logic.  The new change is breaking the mask and causing sensitive data to leak to the logs.  Would you be able to chime in to hint Jeffrey in a direction that will fix his issue, but not break the mask?

src/core/config-manager.ts
```diff
@@ -230,6 +230,9 @@ export class ConfigManager {
                 this.usedConfigs.set(flag.constName, this.usedConfigs.get(flag.constName) + 1 || 1);
                 return this[`_${flag.constName}`];
               },
+              set(value) {
+                this[`_${flag.constName}`] = value;
+              },
```
JeffreyDallas, line 235: need this setter, other wise the below logic would trigger exception
"Cannot set property clusterRef of #<NewConfigClass> which has only a getter"

```
    if (!context_.config.clusterRef) {
      context_.config.clusterRef = this.k8Factory.default().clusters().readCurrent();
    }
```

src/commands/node/tasks.ts
```diff
@@ -1079,6 +1079,9 @@ export class NodeCommandTasks {
           config.deployment,
         );
         for (const networkNodeServices of config.serviceMap.values()) {
+          if (networkNodeServices.accountId === constants.IGNORED_NODE_ACCOUNT_ID) {
+            continue;
+          }
```
JeffreyDallas, line 1084: Need this filter logic here, otherwise, serviceMap contains deleted nodes,
and nodeClient would fail to ping the no longer running node

src/commands/flags.ts
JeffreyDallas: Let me add some context here, the existence of dataMask would cause the default value become "***",
then if no command line flag is given to overwrite the default value,
any read of the flag would get value "***" instead of its original value.

instamenta: I think removing this line and using the data mask whenever they are logged to the CLI would fix the issue, similar to what is done in the snippet. We shouldn't be overriding the default value with the datamask

```
// if the flag's data is masked, display it without the value
else if (flag.definition.dataMask) {
   processedFlags.push(`--${flagName} ${flag.definition.dataMask}`);
}
```

```ts
  public static stringifyArgv(argv: AnyObject): string {
    const processedFlags: string[] = [];

    for (const [name, value] of Object.entries(argv)) {
      // Remove non-flag data and boolean presence based flags that are false
      if (name === '_' || name === '$0' || value === '' || value === false || value === undefined || value === null) {
        continue;
      }

      // remove flags that use the default value
      const flag = Flags.allFlags.find(flag => flag.name === name);
      if (!flag || (flag.definition.defaultValue && flag.definition.defaultValue === value)) {
        continue;
      }

      const flagName = flag.name;

      // if the flag is boolean based, render it without value
      if (value === true) {
        processedFlags.push(`--${flagName}`);
      }

      // if the flag's data is masked, display it without the value
      else if (flag.definition.dataMask) {
        processedFlags.push(`--${flagName} ${flag.definition.dataMask}`);
      }

      // else display the full flag data
      else {
        processedFlags.push(`--${flagName} ${value}`);
      }
    }

    return processedFlags.join(' ');
  }
```

src/core/config-manager.ts
jeromy-cannon: This is abused logic.  There was never a setter because it was meant to be immutable.  It is supposed to be read only if it is coming from the user.  In some cases such as in the Command classes I've noticed we are changing it to be a mutable set of data with business logic.  As we move into the new design it will be more straight forward, and we will use and pass around business objects instead of the user supplied values.

jeromy-cannon: I see what you are talking about though, the code that is already in main is expecting this to be mutable.  I'm fine with letting this go through, and addressing this as we migrate to the new architecture.

jeromy-cannon (approved):

src/commands/flags.ts
jeromy-cannon: thank you for opening the ticket.  We do need to come up with a better solution.  Perhaps maybe provide an unmasked value method, so that it is really obvious that we are using the unmasked version, and it should be done with caution sense it is sensitive data.

Issue #2870, opened 2025-11-06
`solo consensus network deploy --enable-monitoring-support`: installing CRDs from Solo instead of solo-deployment Helm chart


jeromy-cannon:
In https://github.com/hashgraph/solo-charts/pull/183 there were changes made for #2533 .  However, the changes were made in solo-deployment chart which will cause impact to deployments that are pushed to clusters with an existing solo-deployment.

NOTE: the Custom Resources (CRs) can stay in the solo-deployment chart, as they are chart dependent

solo-charts:
- [ ] remove the PodLogs CRD from solo-deployment chart
- [ ] remove the ServiceMonitor CRD from solo-deployment chart
- [ ] keep the [`charts/solo-deployment/values.yaml`](https://github.com/hashgraph/solo-charts/blob/main/charts/solo-deployment/values.yaml) variables to default the crds.podLog/serviceMonitor.enabled to false
- [ ] rename from crds to crs

solo:
- [ ] add a flag for `solo consensus network deploy` `--enable-monitoring-support`, default it to true
- [ ] if `--enable-monitoring-support` is true then interrogate and determine if the `podlogs.monitoring.grafana.com` CRD is already installed, if it isn't then download and install: [`monitoring.grafana.com_podlogs.yaml`](https://github.com/grafana/alloy/blob/v1.11.3/operations/helm/charts/alloy/charts/crds/crds/monitoring.grafana.com_podlogs.yaml) , where `v1.11.3` is a const/env in version.ts
- [ ] check to see if each CRD exists in the cluster for the following CRDs, if it already exists, then set the [value](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-operator-crds/values.yaml) to false to be supplied to the prometheus-operators-crds chart:
```
crd-alertmanagerconfigs.yaml
crd-alertmanagers.yaml
crd-podmonitors.yaml
crd-probes.yaml
crd-prometheusagents.yaml
crd-prometheuses.yaml
crd-prometheusrules.yaml
crd-scrapeconfigs.yaml
crd-servicemonitors.yaml
crd-thanosrulers.yaml
```
- [ ] install the [`prometheus-operator-crds`](https://artifacthub.io/packages/helm/prometheus-community/prometheus-operator-crds) Helm chart unless the `--enable-monitoring-support` is false or if all of the CRDs it supports are already installed

Relates to:
- https://github.com/hashgraph/solo-charts/pull/183
[...]

PR #2903, opened 2025-11-12
feat(monitoring): enable monitoring support installing crds from solo instead of solo deployment helm chart

jan-milenkov:
## Description

Adds new flag:
`--enableMonitoringSupport` - used to enable CRDs for Prometheus and Grafana

Adds new envs to versions file
`GRAFANA_PODLOGS_CRD_VERSION`
`PROMETHEUS_OPERATOR_CRDS_VERSION`

### Related Issues

* Closes # https://github.com/hiero-ledger/solo/issues/2870


src/commands/network.ts
```diff
@@ -937,6 +946,101 @@ export class NetworkCommand extends BaseCommand {
...
+        continue;
+      }
+
+      this.logger.info(`Installing missing CRD ${PODLOGS_CRD} from ${CRD_URL} in context ${context}...`);
+
+      const temporaryFile: string = PathEx.join('/tmp', 'podlogs-crd.yaml');
+
+      // download YAML from GitHub
+      const curlCommand: string = `curl -fsSL ${CRD_URL} -o ${temporaryFile}`;
+      await this.run(curlCommand);
```
jeromy-cannon: Instead of adding the curl binary as a dependency, I would rather us use a library, such as this one: 
- https://www.npmjs.com/package/node-libcurl
- https://github.com/JCMais/node-libcurl/tree/v5.0.1

Kubectl was different, we already use their library, but their library has gaps, so we are using the binary to augment the library to make it more complete.

src/integration/kube/k8-client/resources/crd/k8-client-crds.ts
jeromy-cannon:
```suggestion
        this.logger.info(`CRD ${crdName} does not exist.`);
```

Issue #2842, opened 2025-10-31
(blocked by CN#25883) reduce-memory-footprint need to set the Block Node stream mode to BLOCKS, instead of BOTH


jeromy-cannon:
#reduce-memory-footprint 
for the "Performance Test" workflow which is using the one-shot-single.test.ts, we want to set the `application.properties` `blockStream.streamMode=BLOCKS` instead of `BOTH`.  As part of this, change, we will update the system to make this change anytime Block Node is used.
- need to set it to BLOCKS, instead of both to get proper results, and not writing to disk

Update the consensus node configuration to use BLOCKS setting.

Changes needed:

Switch consensus node setting from current configuration to BLOCKS
Verify compatibility with network operations
Update any related documentation or configuration files
Context:

This aligns with recommended consensus node settings
May improve performance or compatibility

bug fixed in: 
- https://github.com/hiero-ledger/hiero-consensus-node/issues/22262

jeromy-cannon:
@jan-milenkov , do you think this can be worked now? or is it perhaps already implemented?

jan-milenkov:
@jeromy-cannon I think once the blocknode becomes mandatory, then we can do this issue. Before that the logic will have to be gated with check if the deployed network has blocknodes deployed

jeromy-cannon:
@jan-milenkov , if we are running block node turned on, then I think we should be using BLOCKS and not BOTH.  This will give us a more accurate RTT for post Block Node deployment.  If this work is already done and being tested in our pipeline, then this can be closed.

JeffreyDallas:
Currently due to upstream bugs, for performance test BLOCK_STREAM_STREAM_MODE is set to BOTH

jeromy-cannon:
If I'm reading the links correctly, it looks like this has been fixed for a long time, but the CN GHI that was referenced was not ever closed out.  I've updated our reference ticket to the newer ticket that was fixed.

jeromy-cannon:
FYI @JeffreyDallas 

> If I'm reading the links correctly, it looks like this has been fixed for a long time, but the CN GHI that was referenced was not ever closed out. I've updated our reference ticket to the newer ticket that was fixed.

JeffreyDallas:
Not totally fixed yet, there is still smart contract failure (current skipped) due to CN bug,
only BN stream mode issue fixed

jeromy-cannon:
[...]

PR #2959, opened 2025-11-20
feat(block-node): reduce-memory-footprint need to set the Block Node stream mode to BLOCKS, instead of BOTH

jan-milenkov:
## Description

switch BLOCK_STREAM_STREAM_MODE to default to BLOCKS

### Related Issues

* Closes # https://github.com/hiero-ledger/solo/issues/2842


jeromy-cannon:
Nana and Nathan talked, this is not really supportable at this point

PR #3120, opened 2026-01-14
feat(block-node): configure custom network topology

jan-milenkov:
## Description

* Add flag `--priority-mapping` to `solo block node add` and `solo block node upgrade`, allowing to affect which node to connect with the block node and the priority of each.

* Add the functionality to deploy block nodes post-genesis and configure which nodes to connect to, effectively allowing to define a custom network topology.

* Simple default logic to configure backfill when a new block node is added to network with an existing one.


src/commands/block-node.ts
```diff
@@ -207,6 +214,99 @@ export class BlockNodeCommand extends BaseCommand {
...
+              title: `Start node: ${chalk.yellow(nodeAlias)}`,
+              task: async (): Promise<void> => {
+                const node: ConsensusNode = nodes.find((node): boolean => node.name === nodeAlias);
+                const namespace: NamespaceName = NamespaceName.of(node.namespace);
+
+                const k8: K8 = this.k8Factory.getK8(node.context);
+
+                const container: Container = await k8.helpers().getConsensusNodeRootContainer(namespace, nodeAlias);
+
+                await container.execContainer(['bash', '-c', 'systemctl stop network-node || true']);
```
jeromy-cannon: If you just run stop, them systemctl will see it is down, and automatically restart it.  instead:
```
systemctl disable --now network-node
```

src/commands/block-node.ts
```diff
@@ -207,6 +214,99 @@ export class BlockNodeCommand extends BaseCommand {
...
+  }
+
+  private handleConsensusNodeUpdating(): SoloListrTask<BlockNodeDeployContext> {
+    return {
+      title: 'Update consensus nodes',
+      task: (_, task): SoloListr<BlockNodeDeployContext> => {
+        const subTasks: SoloListrTask<BlockNodeDeployContext>[] = [this.updateConsensusNodesInRemoteConfig()];
+
+        if (this.remoteConfig.configuration.state.ledgerPhase !== LedgerPhase.UNINITIALIZED) {
+          subTasks.push(this.updateConsensusNodesPostGenesis(), this.startNodesAgainFromGenesis() as any);
```
jeromy-cannon: what is the purpose we are trying to solve by resetting the consensus nodes to genesis?  This is sort of an extreme action to take and need to think hard about it, as well as making sure it is well documented.  and perhaps even prompt the user about it, else abort.

resources/startPodJava.sh
jeromy-cannon: I don't understand what is happening here.  I see that a clean is taking place.  why are we hard coding the JVM parameters and launching the consensus node directly instead of using systemctl?

jeromy-cannon (changes requested):

Issue #3162, opened 2026-01-19
chore(block-node): add support for new `block-nodes.json` field name for port


jan-milenkov:
There is an update to the `block-nodes.json` file inside the consensus node starting verison 0.70.0 where the `port` field is renamed to `streamingPort`. This PR adds logic to figure out the field name and produce the updated schema if required.

PR #3163, opened 2026-01-19
chore(block-node): add support for new block-nodes.json field name for port #3162

jan-milenkov:
## Description

There is an update to the `block-nodes.json` file inside the consensus node starting verison `0.70.0` where the `port` field is renamed to `streamingPort`. This PR adds logic to figure out the field name and produce the updated schema if requried.

### Related Issues

* Closes # https://github.com/hiero-ledger/solo/issues/3162

### DIFF
--- a/version.ts
+++ b/version.ts
@@ -44,8 +44,7 @@ export const MINIMUM_HIERO_BLOCK_NODE_VERSION_FOR_NEW_LIVENESS_CHECK_PORT: SemVe
 
 export const MINIMUM_HIERO_PLATFORM_VERSION_FOR_BLOCK_NODE: string = 'v0.64.0';
 export const MINIMUM_HIERO_PLATFORM_VERSION_FOR_GRPC_WEB_ENDPOINTS: string = 'v0.62.0';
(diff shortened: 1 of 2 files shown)


version.ts
```diff
@@ -44,6 +44,8 @@ export const MINIMUM_HIERO_BLOCK_NODE_VERSION_FOR_NEW_LIVENESS_CHECK_PORT: SemVe
 
 export const MINIMUM_HIERO_PLATFORM_VERSION_FOR_BLOCK_NODE: string = 'v0.64.0';
 export const MINIMUM_HIERO_PLATFORM_VERSION_FOR_GRPC_WEB_ENDPOINTS: string = 'v0.62.0';
+export const MINIMUM_HIERO_CONSENSUS_NODE_VERSION_FOR_LEGACY_PORT_NAME_FOR_BLOCK_NODES_JSON_FILE: string =
+  constants.getEnvironmentVariable('PROMETHEUS_OPERATOR_CRDS_VERSION') || '0.70.0';
```
jeromy-cannon: copy/paste miss?

PR #3170, opened 2026-01-21
feat(block-node): add remove command for block node external node and improve the existing `solo block node destroy`

jan-milenkov:
## Description

* add new command `solo block node remove-external`
* add the missing functionality to `solo block node destroy`
now they both update the `block-nodes.json` correctly when being removed

### Related Issues

* Closes # https://github.com/hiero-ledger/solo/issues/3168

### DIFF
--- a/src/commands/flags.ts
+++ b/src/commands/flags.ts
@@ -21,7 +21,7 @@ import {PathEx} from '../business/utils/path-ex.js';
 import validator from 'validator';
 
 export class Flags {
-  public static KEY_COMMON = '_COMMON_';
+  public static KEY_COMMON: string = '_COMMON_';
 
   private static async prompt(
     type: 'toggle' | 'input' | 'number',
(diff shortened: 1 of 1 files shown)


src/commands/block-node.ts
```diff
@@ -121,6 +152,10 @@ export class BlockNodeCommand extends BaseCommand {
 
   private static readonly UPGRADE_CONFIGS_NAME: string = 'upgradeConfigs';
 
+  private static readonly ADD_EXTERNAL_CONFIGS_NAME: string = 'addExternalConfigs';
+
+  private static readonly REMOVE_CONFIGS_NAME: string = 'removeExternalConfigs';
```
jeromy-cannon: if we rename to delete external, this should also be updated

src/commands/block-node.ts
```diff
@@ -139,6 +174,16 @@ export class BlockNodeCommand extends BaseCommand {
     ],
   };
 
+  public static readonly ADD_EXTERNAL_FLAGS_LIST: CommandFlags = {
+    required: [flags.deployment, flags.externalBlockNodeAddress],
+    optional: [flags.clusterRef, flags.devMode, flags.quiet, flags.priorityMapping],
+  };
+
+  public static readonly REMOVE_EXTERNAL_FLAGS_LIST: CommandFlags = {
```
jeromy-cannon: if we rename to delete external, this should also be updated

PR #3266, opened 2026-02-05
fix: version flag detection regardless of argv position

Olexandr88:
Fix version flag detection when not in argv[2]

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -15,7 +15,7 @@ import {InjectTokens} from './core/dependency-injection/inject-tokens.js';
 import {SoloError} from './core/errors/solo-error.js';
 import {SilentBreak} from './core/errors/silent-break.js';
 import {ArgumentProcessor} from './argument-processor.js';
-import { getSoloVersion } from './version.js';
+import {getSoloVersion} from '../version.js';
 
 if (!process.stdout.isTTY) {
   chalk.level = 0;


jeromy-cannon (changes requested):

src/index.ts
```diff
@@ -51,23 +48,25 @@ export async function main(argv: string[], context?: {logger: SoloLogger}): Prom
...
+  if (argv.some((argument): boolean => ['-version', '--version', '-v', '--v'].includes(argument))) {
     // Check for --output flag (K8s ecosystem standard)
-    const outputFlagIndex: number = argv.findIndex(
-      (argument): boolean => argument.startsWith('--output=') || argument === '--output' || argument === '-o',
+    const outputFlagIndex = argv.findIndex(
+      argument => argument.startsWith('--output=') || argument === '--output' || argument === '-o',
     );
-    let outputFormat: string = '';
+
+    let outputFormat = '';
```
jeromy-cannon:
```suggestion
    let outputFormat: string = '';
```

src/index.ts
```diff
@@ -51,23 +48,25 @@ export async function main(argv: string[], context?: {logger: SoloLogger}): Prom
...
 
     if (outputFlagIndex !== -1) {
-      const outputArgument: string = argv[outputFlagIndex];
+      const outputArgument = argv[outputFlagIndex];
```
jeromy-cannon:
```suggestion
      const outputArgument: string = argv[outputFlagIndex];
```

src/index.ts
```diff
@@ -51,23 +48,25 @@ export async function main(argv: string[], context?: {logger: SoloLogger}): Prom
...
       if (outputArgument.startsWith('--output=')) {
-        outputFormat = outputArgument.split('=')[1];
+        outputFormat = outputArgument.split('=')[1] ?? '';
       } else if (outputFlagIndex + 1 < argv.length) {
         outputFormat = argv[outputFlagIndex + 1];
       }
     }
 
-    const version: string = getSoloVersion();
+    const version = process.env.npm_package_version ?? '';
```
jeromy-cannon:
```suggestion
    const version: string = getSoloVersion();
```

getSoloVersion() has additional logic.  If it returning something incorrect, perhaps we should add the default empty string to it: https://github.com/hiero-ledger/solo/blob/main/version.ts#L51-L51

Olexandr88: Thanks! Applied all suggestions.

jeromy-cannon (changes requested):

PR #3333, opened 2026-02-10
perf: optimize one-shot single deploy performance

mgarbs:
## Summary

Performance optimizations for `solo one-shot single deploy` to reduce deployment time:

- **Replace unconditional 40s relay sleep with conditional 5s wait** — The relay deploy had a hardcoded 40-second sleep after every Helm install. This is unnecessary on fresh installs since there are no old pods to wait for. Now only sleeps 5s on upgrades, skips entirely on fresh installs.

- **Pre-download platform build to host cache** — Downloads the consensus node build zip and checksum to `$SOLO_CACHE_DIR/platform/` on the host before pod setup begins. During `fetchPlatform()`, cached files are copied directly into pods, allowing `extract-platform.sh` to skip its in-pod download. Especially beneficial on repeat deploys.

- **Parallelize cluster sub-chart installations** — MinIO operator, Prometheus stack, and Grafana agent are independent Helm charts. Install them concurrently after the prerequisite pod-monitor-role ClusterRole is created.

- **Parallelize node proxy readiness checks** — Proxy readiness checks for different consensus nodes are independent. Run them concurrently instead of sequentially.

- **Add exponential backoff to pod polling loops** — Replace fixed-interval polling in `waitForRunningPhase` with exponential backoff (1.5x growth, capped at 5x base delay) to reduce unnecessary Kubernetes API calls during long waits.

- **Pre-warm Helm chart repos during one-shot initialization** — Add an early `chartManager.setup()` call so Helm repos are ready before any chart install steps run, avoiding redundant repo setup during individual installations.

## Test plan

- [ ] Run `solo one-shot single deploy` on a fresh cluster and verify all steps complete successfully
- [ ] Run `solo one-shot single deploy` a second time on same cluster to verify repeat deploy works
- [ ] Verify platform build cache is created at `~/.solo/cache/platform/`
- [ ] Verify relay deploys without the 40s wait on fresh install
- [ ] Run `solo one-shot single destroy` and verify clean teardown

### DIFF
--- a/src/commands/cluster/tasks.ts
+++ b/src/commands/cluster/tasks.ts
@@ -408,14 +408,17 @@ export class ClusterCommandTasks {
           console.log('Skipping Grafana Agent chart installation');
         }
 
-        const result = await task.newListr([
-          this.installPodMonitorRole(argv),
-          {
-            title: 'Install Helm charts',
-            task: (_, subTask) => subTask.newListr(chartSubtasks, {concurrent: true}),
-            skip: () => chartSubtasks.length === 0,
-          },
-        ], {concurrent: false});
+        const result = await task.newListr(
+          [
+            this.installPodMonitorRole(argv),
+            {
+              title: 'Install Helm charts',
+              task: (_, subTask) => subTask.newListr(chartSubtasks, {concurrent: true}),
+              skip: () => chartSubtasks.length === 0,
+            },
+          ],
+          {concurrent: false},
+        );
 
         if (argv.dev) {
           await this.showInstalledChartList(context_.config.clusterSetupNamespace, context_.config.context);
--- a/src/core/platform-installer.ts
+++ b/src/core/platform-installer.ts
@@ -13,8 +13,6 @@ import {Templates} from './templates.js';
 import {Flags as flags} from '../commands/flags.js';
 import * as Base64 from 'js-base64';
 import chalk from 'chalk';
-import {pipeline} from 'node:stream/promises';
-import {Readable} from 'node:stream';
 
 import {type SoloLogger} from './logging/solo-logger.js';
 import {type NodeAlias} from '../types/aliases.js';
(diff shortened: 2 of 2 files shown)


src/commands/node/tasks.ts
```diff
@@ -597,9 +597,9 @@ export class NodeCommandTasks {
       });
     }
 
-    // set up the sub-tasks
+    // Proxy checks for different nodes are independent, run in parallel
```
jeromy-cannon: Yes, I had issues with this previously.  However, the nodes are actually being deployed in parallel already.  This is actually only checking that they have completed.  They tend to become all running at the same time, so when we run the checks in parallel, it actually increases the traffic/load on the system and slightly increases the amount of time it takes, as well as increases log white noise.

In single node deployments, such as one-shot single, this has no impact.  But, when using multiple nodes, this can become more obvious.

Please revert or let me know if you disagree.

src/commands/one-shot/default-one-shot.ts
```diff
@@ -190,11 +191,21 @@ export class DefaultOneShotCommand extends BaseCommand implements OneShotCommand
 
               context_.createdAccounts = [];
 
+              // Pre-warm Helm chart repos so they're ready when needed
+              await this.chartManager.setup(constants.DEFAULT_CHART_REPO, true);
```
jeromy-cannon: We will be implementing a caching feature which will makes this redundant and obsolete.  Also, this doesn't actually download anything.  It is only making a connection to the repository if needed.  It doesn't download the charts and it doesn't do a helm dependency update.  Also, this is only for the solo chart repo, it does not include the other helm chart repositories.  And, since this is inside of one-shot, would still be included in the one-shot deployment time.

src/commands/one-shot/default-one-shot.ts
```diff
@@ -190,11 +191,21 @@ export class DefaultOneShotCommand extends BaseCommand implements OneShotCommand
...
               return;
             },
           },
+          {
+            title: 'Pre-download platform build to host cache',
+            task: async (): Promise<void> => {
+              const platformInstaller = diContainer.resolve<PlatformInstaller>(InjectTokens.PlatformInstaller);
+              await platformInstaller.preDownloadPlatformBuild(version.HEDERA_PLATFORM_VERSION);
+            },
+          },
```
jeromy-cannon: good intention, but this would still count towards one-shot deployment time, and we have another caching feature that would be superior to this.  Additionally, if one-shot is not using the default platform version via env variable or falcon, then this would actually cause us to download two different copies, thus increasing total time.

src/integration/kube/k8-client/resources/pod/k8-client-pods.ts
```diff
@@ -221,7 +221,9 @@ export class K8ClientPods extends K8ClientBase implements Pods {
         }
 
         if (++attempts < maxAttempts) {
-          setTimeout((): Promise<void> => check(resolve, reject), delay);
+          // Exponential backoff: start at base delay, grow by 1.5x each attempt, cap at 5x base delay
+          const backoffDelay = Math.min(delay * Math.pow(1.5, Math.min(attempts, 10)), delay * 5);
```
jeromy-cannon:
```suggestion
          const backoffDelay = Math.min(delay * Math.pow(1.5, Math.min(attempts, 10)), delay * 5);
          this.logger.info(
            `Pod not in desired phase(s) yet, retrying... [attempt: ${attempts}/${maxAttempts}, next attempt in ${backoffDelay}ms]`,
          );
```

jeromy-cannon (changes requested):

PR #3434, opened 2026-02-24
chore: test with platform version `v0.72.0-alpha.3`

jan-milenkov:


jan-milenkov:
PR used for testing no longer needed

PR #3818, opened 2026-03-27
chore: bump consensus node to v0.72.0 and block node to v0.30.2

JeffreyDallas:
## Summary
- Bump default consensus node version from `v0.71.0` → `v0.72.0` in `version.ts`
- Bump default block node version from `v0.28.1` → `v0.29.0` in `version.ts`
- Skip dual cluster test and version upgrade due to pending https://github.com/hiero-ledger/hiero-consensus-node/issues/24642
- Modify performance test to use stream mode due to mirror node issue #3972
- Add one copy block-nodes.json step because oneshot parallel mode have race condition that the copy step of block node and copy step of network deploy both not satisfied and skipped
- change performance test to use falcon command to be able to use customized properties
- add a waitForMirrorOperatorAccountReady step so relay node continues until mirror database have 0.0.2 account information for relay to query (less than 1 second delay)


test/e2e/commands/performance.test.ts
```diff
@@ -273,11 +289,16 @@ export function soloOneShotDeploy(testName: string, deployment: string): string[
...
     OneShotCommandDefinition.SINGLE_DEPLOY,
   );
   argvPushGlobalFlags(argv, testName);
-  argv.push(optionFromFlag(Flags.deployment), deployment);
+  argv.push(
+    optionFromFlag(Flags.deployment),
+    deployment,
+    optionFromFlag(Flags.valuesFile),
+    'test/data/performance-test-values.yaml',
+  );
```
jeromy-cannon, line 302: I'm concerned that with these configurations we will no longer match the users experience.  What are these configurations doing?  What is the reason they cannot become the defaults for `solo one-shot single deploy`?

JeffreyDallas: Still the exactly same one node solo, network. The same test results.
But `single` does not support customized value file.
If we do not do this, we will stuck at CN 0.71 and BN 0.28, while user kept complaining about various failed case when using CN 0.72/0.73 , or BN 0.29/0.30, since we never been able to get chance to test and verify those combinations

jeromy-cannon: so what if we just conditionally add these new configurations automatically gated based on the CN and BN versions so that it can just work in `solo one-shot single deploy`?

jan-milenkov:
- Block node version in PR title, as well as created issues to other team seems to be wrong, update all to the same either `v0.30.2` or `v0.29.0`

- Now the performance test also uses block node, check if we are not missing logic to catch it failing

.github/workflows/flow-examples-test.yaml
```diff
@@ -340,7 +344,27 @@ jobs:
           echo "Running example: ${{ matrix.example.name }}"
           echo "Resolved example source: ${{ steps.prepare_example.outputs.display_dir }}"
           echo "Current directory: $(pwd)"
-          
+
+          # Skip version-upgrade-test if consensus version >= v0.72.0
```
jan-milenkov, line 348: Check if logic can be simplified, bitwise operators and version comparisons have code smell

Add `TODO` comment for removing this logic once it's redundant

.github/workflows/zxc-e2e-test.yaml
```diff
@@ -440,6 +441,32 @@ jobs:
         run: |
           export PATH=~/.solo/bin:${PATH}
           echo "PATH: $PATH"
+
+          # Skip test-e2e-dual-cluster-full if consensus version >= v0.72.0
```
jan-milenkov, line 445: same as the one above, check if logic can be simplified or extracted to a separate script, especially now that this piece of code is duplicated

Add `TODO` comment for removing this logic once it's redundant

version.ts
```diff
@@ -18,15 +18,15 @@ export const VFKIT_VERSION: string = 'v0.6.1';
 export const GVPROXY_VERSION: string = 'v0.8.7';
 export const KUBECTL_VERSION: string = 'v1.32.2';
 export const SOLO_CHART_VERSION: string = constants.getEnvironmentVariable('SOLO_CHART_VERSION') || '0.63.3';
-export const HEDERA_PLATFORM_VERSION: string = constants.getEnvironmentVariable('CONSENSUS_NODE_VERSION') || 'v0.71.0';
```
jan-milenkov, line 21: Mainnet is at version `v0.71.3` so why are we switching to `v0.72`

src/commands/one-shot/default-one-shot.ts
```diff
@@ -958,6 +967,77 @@ export class DefaultOneShotCommand extends BaseCommand implements OneShotCommand
...
+      mirrorNode?.metadata?.portForwardConfigs?.[0];
+
+    if (!portForwardConfig) {
+      throw new SoloError('Mirror ingress port-forward is not configured');
+    }
+
+    return portForwardConfig.localPort;
+  }
+
+  private async waitForMirrorOperatorAccountReady(
```
jan-milenkov, line 993: seem to be unused, if this is needed or will be used, switch it off to use the newly added `DefaultEventBus`

JeffreyDallas:
Marked as blocked to wait for MN 0.152.0 changes merged first

JeffreyDallas:
Replaced with PR #3988

Issue #3931, opened 2026-04-07
Create a working implementation for pulling/saving/loading images for Docker and create the supporting cache commands


jan-milenkov:
### Description

Add the ability for solo to handle the lifecycle of images, this includes pulling them from the repositories, storing them inside the cache folder and loading them inside of the cluster.

PR #3956, opened 2026-04-09
feat(cache): Implement cache command and feature for docker images

jan-milenkov:
### Envs

- `ENABLE_IMAGE_CACHE` - Used to enable the image cache feature during one-shot deployments.

> this one is currently enabled for the standard e2e-one-shot-deployment test 

### Commands

Add new command group `cache` together with a new subcommand group `image`, which under there are:
- `pull` - pull and caches docker images used by solo, prerequisite for `solo cache image load`.
- `load` - loads the images archive into a cluster. Pulling the images with `solo cache images pull` is a prerequisite.
- `list` - lists all cached image archives.
- `clear` - clears the image archives.
- `status` - lists all images, displays data about them and displays all missing images.

The image cache feature is currently working only with `docker`

### Dependencies

- Adds new dependency for `crane` - https://michaelsauter.github.io/crane/docs.html
- Crane is required only for the new `solo image cache`, and is optional if this feature is not used

### Related Issues

* Closes # https://github.com/hiero-ledger/solo/issues/3931


src/core/dependency-injection/container-init.ts
```diff
@@ -192,6 +202,13 @@ export class Container {
       new SingletonContainer(InjectTokens.PostgresSharedResource, PostgresSharedResource),
       new SingletonContainer(InjectTokens.SharedResourceManager, SharedResourceManager),
 
+      // Cache
+      new SingletonContainer(InjectTokens.CacheHandlerRegistry, DefaultCacheHandlerRegistry),
+      new SingletonContainer(InjectTokens.FileSystemCacheCatalogStore, FileSystemCacheCatalogStore),
+      new SingletonContainer(InjectTokens.CacheHealthInspector, DefaultCacheHealthInspector),
+      new SingletonContainer(InjectTokens.ImageCacheHandlerBuilder, ImageCacheHandlerBuilder),
+      new SingletonContainer(InjectTokens.DockerClient, DockerClient),
```
jeromy-cannon:
```suggestion
      new SingletonContainer(InjectTokens.ContainerEngineClient, DockerClient),
```

src/core/dependency-injection/container-init.ts
```diff
@@ -192,6 +202,13 @@ export class Container {
       new SingletonContainer(InjectTokens.PostgresSharedResource, PostgresSharedResource),
       new SingletonContainer(InjectTokens.SharedResourceManager, SharedResourceManager),
 
+      // Cache
+      new SingletonContainer(InjectTokens.CacheHandlerRegistry, DefaultCacheHandlerRegistry),
+      new SingletonContainer(InjectTokens.FileSystemCacheCatalogStore, FileSystemCacheCatalogStore),
```
jeromy-cannon:
```suggestion
      new SingletonContainer(InjectTokens.CacheCatalogStore, FileSystemCacheCatalogStore),
```

src/core/dependency-injection/container-init.ts
```diff
@@ -192,6 +202,13 @@ export class Container {
...
+      new SingletonContainer(InjectTokens.CacheHealthInspector, DefaultCacheHealthInspector),
+      new SingletonContainer(InjectTokens.ImageCacheHandlerBuilder, ImageCacheHandlerBuilder),
```
jeromy-cannon, line 209: So, if we want a singleton, then it would be:

```suggestion
      new SingletonContainer(InjectTokens.ImageCacheHandler, DefaultImageCacheHandlerBuilder),
```
However, if you are wanting the builder to be able to use the build from yaml, then you might want to use the factorySupplier or a factory

src/integration/cache/api/cache-operation-handler.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Returns the cache items this handler manages.
+   *
+   * This is a logical listing of the artifacts the handler expects in the cache.
+   * It may be used by callers to inspect or report cache contents.
+   */
+  list(): Promise<readonly CachedItem[]>;
+
+  resolveRequiredArtifacts(): Promise<readonly CacheTargetStructure[]>;
```
jeromy-cannon, line 70: It is good to have the TS docs on all of the interface methods and then not need them on the implementations

src/integration/cache/impl/helm-chart-cache-handler.ts
```diff
@@ -0,0 +1,79 @@
...
+import {type CacheHealthInspector} from '../api/cache-health-inspector.js';
+import {type HelmClient} from '../../helm/helm-client.js';
+import {Chart} from '../../helm/model/chart.js';
+import {InjectTokens} from '../../../core/dependency-injection/inject-tokens.js';
+import {patchInject} from '../../../core/dependency-injection/container-helper.js';
+import {inject, injectable} from 'tsyringe-neo';
+import {CacheTarget} from '../models/impl/cache-target.js';
+
+@injectable()
+// export class HelmChartCacheHandler implements CacheOperationHandler {
```
jeromy-cannon: ?

src/integration/cache/impl/image-cache-handler.ts
```diff
@@ -0,0 +1,149 @@
...
+import {type AnyListrContext} from '../../../types/aliases.js';
+import chalk from 'chalk';
+import {type SoloLogger} from '../../../core/logging/solo-logger.js';
+import {type FileSystemCacheCatalogStore} from './file-system-cache-catalog-store.js';
+
+export class ImageCacheHandler implements CacheOperationHandler {
+  public constructor(
+    private readonly engine: ContainerEngineClient,
+    private readonly provider: CacheTargetProvider,
+    @inject(InjectTokens.FileSystemCacheCatalogStore) public readonly store?: FileSystemCacheCatalogStore,
```
jeromy-cannon:
```suggestion
    @inject(InjectTokens.CacheCatalogStore) public readonly store?: CacheCatalogStore,
```

src/integration/container-engine/docker-client.ts
```diff
@@ -0,0 +1,84 @@
...
+import {type SoloLogger} from '../../core/logging/solo-logger.js';
+import {DefaultKindClientBuilder} from '../kind/impl/default-kind-client-builder.js';
+import {DependencyManager} from '../../core/dependency-managers/index.js';
+import * as constants from '../../core/constants.js';
+import {LoadImageArchiveOptionsBuilder} from '../kind/model/load-image-archive/load-image-archive-options-builder.js';
+import {LoadImageArchiveOptions} from '../kind/model/load-image-archive/load-image-archive-options.js';
+
+@injectable()
+export class DockerClient implements ContainerEngineClient {
+  private readonly sh: ShellRunner;
```
jeromy-cannon:
```suggestion
  private readonly shellRunner: ShellRunner;
```

src/integration/container-engine/docker-client.ts
```diff
@@ -0,0 +1,84 @@
...
+      .build();
+
+    const kindExecutable: string = await this.dependencyManager.getExecutable(constants.KIND);
+    const kindClient: KindClient = await this.kindBuilder.executable(kindExecutable).build(true);
+
+    await kindClient.loadImageArchive(archivePath, options);
+  }
+
+  public async removeImage(image: string): Promise<void> {
+    await this.sh.run(`docker image rm ${this.quote(image)}`);
```
jeromy-cannon: during the security audit we determined that we need to set `shell: false` for all or nearly all shellRunner calls.  This means that we should always be using the array structured version instead of running it like a command line on the shell.

See: https://github.com/hiero-ledger/solo/issues/4005

src/integration/container-engine/docker-client.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  private defaultLinuxPlatform(): string {
+    switch (process.arch) {
+      case 'arm64': {
+        return 'linux/arm64';
+      }
+      case 'x64': {
+        return 'linux/amd64';
+      }
```
jeromy-cannon, line 74: At some point, not in this PR, or maybe you could introduce it, but have a GHI to update the codebase to use it, do like I did with `operating-system.ts`, and create an `architecture.ts`, then we can update the entire codebase to use that instead of having all of these architecture strings all over the place that could be error prone.

test/unit/core/kind/default-kind-cluster/load-image-archive.test.ts
```diff
@@ -1,196 +1,185 @@
...
   let client: DefaultKindClient;
-  let executionBuilderStub: sinon.SinonStubbedInstance<KindExecutionBuilder>;
   let executionStub: sinon.SinonStubbedInstance<KindExecution>;
   let builderArguments: Map<string, string>;
   let builderSubcommands: string[];
-
-  beforeEach(() => {
+  let builderPositionals: string[];
+  let buildStub: SinonStub;
+  // eslint-disable-next-line unused-imports/no-unused-vars
```
jeromy-cannon, line 22: why do you need this?  why disable the linter?

If you are never using it, then instead of:
```
buildStub = sinon.stub(KindExecutionBuilder.prototype, 'build').returns(executionStub as never);
```
you can do this:
```
sinon.stub(KindExecutionBuilder.prototype, 'build').returns(executionStub as never);
```

.github/workflows/zxc-e2e-test.yaml
```diff
@@ -469,6 +469,10 @@ jobs:
             echo "FORCE_PODMAN=true" >> .env
           fi
           
+          if [[ "${{ inputs.custom-job-label }}" == "One Shot Single" ]]; then
+            echo "ENABLE_IMAGE_CACHE=true" >> .env
+          fi
```
jeromy-cannon, line 475: we should always enable the image cache and always load the contents of the image cache into the cluster.  Perhaps default this ENABLE_IMAGE_CACHE to true, and only needs to be set to false in a rare condition.

if no image cache is found to be loaded, then make an info log message and proceed.
if anything is wrong with the image cache, we can load what we do have that is correct, and log warnings to the log files about anything else, and proceed without causing user interruption.

let me know if you have thoughts, concerns, or other suggestions.

resources/config/solo-cache-images-target.yaml
jeromy-cannon, line 1: instead of checking this in, and having to maintain it, it should be a build artifact and a release artifact.  The first version can be manually uploaded to the latest release, after that the tag and release should include it in the release artifacts to upload.

create a GHI, and link it.  we can set it as P2 to work in Q3.

I was also considering bundling both defaults and edge.  However, in Q3, edge will become a moving target once we have all components doing a TCK PR check, and update edge to calculate the component latest to pull...

src/commands/command-definitions/one-shot-command-definition.ts
```diff
@@ -60,7 +60,7 @@ export class OneShotCommandDefinition extends BaseCommandDefinition {
               this.oneShotCommand,
               this.oneShotCommand.deploy,
               DefaultOneShotCommand.DEPLOY_FLAGS_LIST,
-              [...constants.BASE_DEPENDENCIES],
+              [...constants.BASE_DEPENDENCIES, ...(constants.CONFIG.ENABLE_IMAGE_CACHE ? [constants.CRANE] : [])],
```
jeromy-cannon, line 63: maybe just move this into base dependencies since we will always want to use the cache.

src/commands/cache.ts
```diff
@@ -0,0 +1,503 @@
...
+        throw new SoloError(`Error pulling cache: ${error.message}`, error);
+      }
+    } else {
+      this.taskList.registerCloseFunction(async (): Promise<void> => {});
+    }
+
+    return true;
+  }
+
+  public async load(argv: ArgvStruct): Promise<boolean> {
```
jeromy-cannon, line 171: this is fine, but, I'm thinking that we should just automatically check to see if the cache has already been loaded into the cluster, if it hasn't then load it, unless the env var is set to false.

We could take the SHA256 of the cache manifest and load it as a variable value in a resource into the cluster.  Perhaps under a solo-cache namespace and solo-cache configmap.

Then we could do a quick check against the cluster to determine if we should load the cache or not.

We could be selective about which commands automatically perform this command, as it doesn't make sense if it is a command that would never cause a cache item to be used.

We should probably also classify all of our commands with certain characteristics to make it easier to determine which activities should be applied.  

```
commandMetaData: {
  attributes: [
    'CLUSTER_MUTATION',
    'LOCAL_CONFIG_MUTATION',
    'REMOTE_CONFIG_MUTATION',
    'LEDGER_MUTATION',
    'READ_ONLY',
...
  ]
}
```
just a thought for future design

I know that several of these comments are scope creep.  So, let us create GHIs where applicable, and link them and get this merged.

src/integration/cache/api/cache-operation-handler.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Returns the cache items this handler manages.
+   *
+   * This is a logical listing of the artifacts the handler expects in the cache.
+   * It may be used by callers to inspect or report cache contents.
+   */
+  list(): Promise<readonly CachedItemStructure[]>;
+
+  resolveRequiredArtifacts(): Promise<readonly CacheTargetStructure[]>;
```
jeromy-cannon: we should have TSDoc on all of the interface methods perferrably.

src/commands/one-shot/default-one-shot.ts
```diff
@@ -480,6 +481,36 @@ export class DefaultOneShotCommand extends BaseCommand implements OneShotCommand
             skip: (context_: OneShotSingleDeployContext): boolean =>
               context_.config.force === true || context_.config.quiet === true,
           },
+          invokeSoloCommand(
+            `solo ${CacheCommandDefinition.IMAGE_PULL_COMMAND}`,
```
jeromy-cannon, line 495: 1. does the pull command check to see if anything needs to be pulled and complete successfully if no work is to be done?
2. we want the pull to happen during the brew install process by default.
3. the npm install process should kick out a message at the end to suggest to the user to do a cache pull command

again, create GHIs as necessary so we can get this merged.

.github/workflows/zxc-e2e-test.yaml
jan-milenkov: I think we should gate this behind an environment variable for now, until it’s been more thoroughly tested and refined.

This PR introduces the initial implementation, but it’s not meant to be complete. It should be iterated on and optimized in follow-ups. For that reason, I’d recommend keeping it disabled by default and enabling it explicitly via env while we validate the behavior.

src/commands/cache.ts
jan-milenkov: I think we should avoid overworking this PR, we have a future epic for further enhancement on the cache, maybe this would be a better fit there (as an issue)

src/commands/one-shot/default-one-shot.ts
jan-milenkov: 1. yes

jan-milenkov: all issues are linked to the epic https://github.com/hiero-ledger/solo/issues/4141

jan-milenkov: I believe the scope of this PR is complete. Any additional enhancements or follow-ups have been captured in GitHub issue #4141.

src/commands/cache.ts
jeromy-cannon: agreed, add as a bullet to epic or issue, and link.

resources/config/solo-cache-images-target.yaml
jeromy-cannon: add as a bullet to epic or issue, and link.

jeromy-cannon (dismissed):

jeromy-cannon (dismissed):

jeromy-cannon (approved):

PR #4286, opened 2026-05-08
chore: improve gRPC service healthcheck for node commands

### LINKED ISSUE #4016: Gate Node Readiness on JVM Stability, Not Just Platform Status
### Feature to Enhance

The consensus node readiness check in `checkNetworkNodeActiveness()` (`src/commands/node/tasks.ts`). This function polls `http://localhost:9999/metrics` inside the node pod and waits for `platform_PlatformStatus` to reach `ACTIVE` (status code `2`). Once `ACTIVE` is reached, it applies a fixed stabilization delay (`NETWORK_NODE_ACTIVE_EXTRA_DELAY_MS`, default: 2000ms) before declaring the node ready to accept traffic.

### Requested Enhancement

After a node reaches `ACTIVE` platform status, add a secondary readiness gate that verifies JVM stability before declaring the node ready.

### Business Need

Solo v0.64 parallelized its startup sequence, which means consensus node JVMs may still be warming up (and under heavy GC pressure) by the time Solo signals readiness to callers. The consensus nodes run on the G1GC collector. While each individual pause may be short, cumulative GC activity during JVM warmup can cause a node to be unresponsive to gRPC traffic for 20–30 seconds even after it has reported `ACTIVE` platform status.

From a client's perspective, the node briefly "disappears" - the gRPC connection pool times out waiting for a response, and the request fails with `DEADLINE_EXCEEDED` or a connection pool error rather than being retried cleanly. The existing 2-second fixed delay does not cover a worst-case 20–30 second GC pause.

The comment in the code itself acknowledges the current approach is a workaround:

```typescript
if (constants.NETWORK_NODE_ACTIVE_EXTRA_DELAY_MS > 0) {
  await sleep(Duration.ofMillis(constants.NETWORK_NODE_ACTIVE_EXTRA_DELAY_MS)); // delaying prevents - gRPC service error
}
```

### Additional Context

- [hiero-ledger/hiero-sdk-swift#593](https://github.com/hiero-ledger/hiero-sdk-swift/pull/593) - direct consequence of this behavior; the Swift SDK had to add an explicit catch for `GRPCConnectionPoolError` because Solo was advertising nodes as ready while they were still GC-pausing

### LINKED ISSUE #4315: Intermittent 400 Bad Request / 502 during contract deployment right after Solo startup despite readiness checks
### Describe the bug

Solo generally works well for me. On both my local machine and CI, it starts in around 3 minutes, which is acceptable for my use case.

However, I am seeing an intermittent issue when starting Solo and then deploying contracts for later tests. Sometimes, right after Solo appears to be ready, contract deployment fails with a JSON-RPC error.

I suspect this may be some kind of race condition where Solo reports readiness, but one of the underlying services is not fully ready to handle contract deployment or transaction simulation yet.
```
Error: server response 400 Bad Request (request={  }, response={  }, error=null, info={ "requestUrl": "http://127.0.0.1:37546", "responseBody": "{\"error\":{\"code\":-32000,\"message\":\"[Request ID: e58df071-741d-4520-b8ab-02790c76766c] Error occurred during transaction simulation: Request failed with status code 502\"},\"jsonrpc\":\"2.0\",\"id\":2}", "responseStatus": "400 Bad Request" }, code=SERVER_ERROR, version=6.15.0)
    at makeError (/home/runner/_work/hedera-agent-kit-js/hedera-agent-kit-js/node_modules/.pnpm/ethers@6.15.0/node_modules/ethers/src.ts/utils/errors.ts:698:21)
    at assert (/home/runner/_work/hedera-agent-kit-js/hedera-agent-kit-js/node_modules/.pnpm/ethers@6.15.0/node_modules/ethers/src.ts/utils/errors.ts:719:25)
    at FetchResponse.assertOk (/home/runner/_work/hedera-agent-kit-js/hedera-agent-kit-js/node_modules/.pnpm/ethers@6.15.0/node_modules/ethers/src.ts/utils/fetch.ts:950:9)
    at JsonRpcProvider._send (/home/runner/_work/hedera-agent-kit-js/hedera-agent-kit-js/node_modules/.pnpm/ethers@6.15.0/node_modules/ethers/src.ts/providers/provider-jsonrpc.ts:1272:18)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
    at async <anonymous> (/home/runner/_work/hedera-agent-kit-js/hedera-agent-kit-js/node_modules/.pnpm/ethers@6.15.0/node_modules/ethers/src.ts/providers/provider-jsonrpc.ts:545:40) {
  code: 'SERVER_ERROR',
  request: FetchRequest {},
  response: FetchResponse {},
  error: undefined,
  info: {
    requestUrl: 'http://127.0.0.1:37546',
    responseBody: '{"error":{"code":-32000,"message":"[Request ID: e58df071-741d-4520-b8ab-02790c76766c] Error occurred during transaction simulation: Request failed with status code 502"},"jsonrpc":"2.0","id":2}',
    responseStatus: '400 Bad Request'
  },
  shortMessage: 'server response 400 Bad Request'
}
```

To avoid deploying contracts too early, I added a script that checks whether all required endpoints are ready before contract deployment starts.

The script reports success before the deployment attempt:
```
✓ Consensus node (gRPC) : 127.0.0.1:35211 (0.0.3)  (listening)
· Mirror node    (gRPC) : 127.0.0.1:5600  (cluster-internal, not probed)
✓ Mirror node    (REST) : http://127.0.0.1:38081/api/v1/network/nodes  (ready)
✓ Mirror node    (Web3) : http://127.0.0.1:38081/api/v1/contracts/call  (ready)
✓ JSON-RPC Relay        : http://127.0.0.1:37546  (eth_blockNumber: ready (block=0x12))
```

Despite these checks passing, the deployment still occasionally fails with the error above.

Example failing workflow:
https://github.com/hashgraph/hedera-agent-kit-js/actions/runs/25561808085/job/75035326318?pr=773

Full logs from workflow:
[failing-smart-contract-deployment.log](https://github.com/user-attachments/files/27636105/failing-smart-contract-deployment.log)

After all readiness checks pass, contract deployment should work reliably.

### To Reproduce

1. Deploy Solo `solo one-shot single deploy`
2. Deploy Smart Contract immediately after Solo is ready

### Additional Context

_No response_

jan-milenkov:
## Description

Upgrade the logic around the consensus node activeness checks, by adding logic that requires for the gRPC service to be pinged with successfully 

### Envs

- Removed
  - `NETWORK_NODE_ACTIVE_EXTRA_DELAY_MS` 
- Added
  - `NETWORK_NODE_GRPC_READINESS_MAX_ATTEMPTS` - The maximum number of attempts to check if gRPC service is ready.
  - `NETWORK_NODE_GRPC_READINESS_DELAY` - The interval between attempts to check if gRPC service is ready, in the unit of milliseconds.
  - `NETWORK_NODE_GRPC_READINESS_REQUIRED_SUCCESSES` - The successful passes needed for the healthcheck to pass       

### Related Issues

* Closes #4016
* Closes #4315


src/core/account-manager.ts
```diff
@@ -1201,4 +1201,50 @@ export class AccountManager {
     }
     return accountMap;
   }
+
+  public async loadNodeClientForNodeAlias(
```
JeffreyDallas: This new function is 90% like existing refreshNodeClient function.
Difference is: one excludes a node vs only one node
We should be able to merge the two with one node selection logic.

JeffreyDallas (commented):
see comment

src/core/account-manager.ts
jan-milenkov: updated

JeffreyDallas (dismissed):

src/commands/node/tasks.ts
```diff
@@ -661,13 +663,69 @@ export class NodeCommandTasks {
...
+    this.logger.showUser(
+      `node '${nodeAlias}' failed gRPC readiness check ` +
+        `[ attempt = ${chalk.blueBright(`${attempt}/${constants.NETWORK_NODE_GRPC_READINESS_MAX_ATTEMPTS}`)} ]`,
+    );
+
+    // Erroring disabled to prevent the additional logic from blocking deployments
+    // throw new SoloError(
+    //   `node '${nodeAlias}' failed gRPC readiness check ` +
+    //     `[ attempt = ${chalk.blueBright(`${attempt}/${constants.NETWORK_NODE_GRPC_READINESS_MAX_ATTEMPTS}`)} ]`,
+    // );
```
jeromy-cannon: can we just delete this?

jeromy-cannon (dismissed):

src/commands/node/tasks.ts
jan-milenkov: updated

JeffreyDallas (approved):

PR #4386, opened 2026-05-15
fix: Reduce CN memory footprint

timo0:
Reduce Memory Limits back


JeffreyDallas (changes requested):
These values are gated, only being used when CN >= 0.74
but failed due to OOM

https://github.com/hiero-ledger/solo/actions/runs/25897540404/job/76115179623#step:12:1267

JeffreyDallas (approved):

jeromy-cannon (approved):

PR #4410, opened 2026-05-16
fix: migration test intermittently failure

### LINKED ISSUE #4365: Intermittently migration test failure on main branch
### Describe the bug

https://github.com/hiero-ledger/solo/actions/runs/25872578729/job/76031434282

```
wallet1 = 0xeC8641b927447e971D28c66f16ffee7A701D217E
wallet2 = 0x21D68c33322E7597258fc803044B95153e98c76c
Error in before hook (attempt 1/5): [Request ID: da424349-d273-43c7-a2ff-9b0d8386d483] Requested resource not found. address '0xeC8641b927447e971D28c66f16ffee7A701D217E'. ProviderError: [Request ID: da424349-d273-43c7-a2ff-9b0d8386d483] Requested resource not found. address '0xeC8641b927447e971D28c66f16ffee7A701D217E'.
    at HttpProvider.request (/home/runner/_work/solo/hedera-smart-contracts/node_modules/hardhat/src/internal/core/providers/http.ts:116:21)
    at processTicksAndRejections (node:internal/process/task_queues:103:5)
    at HardhatEthersSigner.sendTransaction (/home/runner/_work/solo/hedera-smart-contracts/node_modules/@nomicfoundation/hardhat-ethers/src/signers.ts:125:18)
    at ContractFactory.deploy (/home/runner/_work/solo/hedera-smart-contracts/node_modules/ethers/src.ts/contract/factory.ts:111:24)
    at Context.<anonymous> (/home/runner/_work/solo/hedera-smart-contracts/test/openzeppelin/ERC-20/ERC20.js:39:25)
    1) "before all" hook for "should be able to execute name()"


  0 passing (568ms)
  1 failing

  1) @OZERC20 Test Suite
       "before all" hook for "should be able to execute name()":
     ProviderError: [Request ID: da424349-d273-43c7-a2ff-9b0d8386d483] Requested resource not found. address '0xeC8641b927447e971D28c66f16ffee7A701D217E'.
      at HttpProvider.request (/home/runner/_work/solo/hedera-smart-contracts/node_modules/hardhat/src/internal/core/providers/http.ts:116:21)
      at processTicksAndRejections (node:internal/process/task_queues:103:5)
      at HardhatEthersSigner.sendTransaction (/home/runner/_work/solo/hedera-smart-contracts/node_modules/@nomicfoundation/hardhat-ethers/src/signers.ts:125:18)
      at ContractFactory.deploy (/home/runner/_work/solo/hedera-smart-contracts/node_modules/ethers/src.ts/contract/factory.ts:111:24)
      at Context.<anonymous> (/home/runner/_work/solo/hedera-smart-contracts/test/openzeppelin/ERC-20/ERC20.js:39:25)


```

should not fail

### To Reproduce

PR checks

### Additional Context

_No response_

JeffreyDallas:
## Description

migration test intermittently failure

* Simplified the first run with one-shot falcon command
* Removed many old kubectl port forward logic since we currently have build-in port forward refresh command
* Addressed a running hash mismatch issue (https://github.com/hiero-ledger/hiero-consensus-node/issues/25486) 
* CN would sometimes continue to generate empty round after freeze, so although no new record streams are pushed to importer, but hash chain keeps running and prev-hash keep changing
* Once CN restart from saved state, it will generate new record stream based on new/changes previous hash, while importer still expecting record steam signature calculated based on old previous hash from freeze step.
* So some new steps added to migration test to detect the gap, and manually implant the new prev-hash into importer
* Update port forward refresh to make sure the port forward process always attached to the new pod not the terminating old ones
* Update address/gossip generation function to always respect the `gossipFqdnRestricted` set in configMap or `application.properties` file

### Related Issues

* Closes #4365
* Related to #

### Testing

* TBD

The following was not tested:

* TBD

<details>
<summary>
Commit message guidelines
</summary>

</details>


src/core/helpers.ts
```diff
@@ -1,12 +1,78 @@
...
+      const repoProperties: string = fs.readFileSync(repoPath, 'utf8');
+      const parsed: boolean | undefined = parseGossipFqdnRestricted(repoProperties);
+      if (parsed !== undefined) {
+        return parsed;
+      }
+    }
+  }
+  return true;
+}
+// SPDX-License-Identifier: Apache-2.0
```
jeromy-cannon: The file now has **two** `// SPDX-License-Identifier: Apache-2.0` headers (lines 1 and 67) and `resolveGossipFqdnRestricted` is placed *above* all imports. Function declarations are hoisted so this happens to run, but the structure is broken — one SPDX header per file, and exported helpers belong below the imports next to the other gossip helpers (`parseGossipFqdnRestricted`, `readGossipFqdnRestrictedFromFile`).

```suggestion
```

Delete this duplicate header and move the function below the imports.

src/core/helpers.ts
```diff
@@ -1,12 +1,78 @@
 // SPDX-License-Identifier: Apache-2.0
 
+export async function resolveGossipFqdnRestricted(options: {
+  k8?: {
+    configMaps: () => {
+      read: (namespace: unknown, configMapName: string) => Promise<{data?: Record<string, string>}>;
+    };
+  };
+  namespace?: unknown;
```
jeromy-cannon: `k8?: { configMaps: () => ... }` and `namespace?: unknown` sidestep the real `K8` / `NamespaceName` types from `src/integration/kube/`. We've taken care to give those domain types throughout the rest of `core/` — please use them here too, otherwise callers can pass anything that has a `configMaps()` method and the type system can't catch a mistake.

.github/workflows/script/launch_network.sh
```diff
@@ -116,14 +443,27 @@ then
...
+  installedSoloVersion="${installedSoloVersion#v}"
+fi
+
+if [[ -n "${installedSoloVersion}" && "${installedSoloVersion}" == "${expectedSoloVersion}" ]]; then
+  echo "Solo version ${installedSoloVersion} already installed, skipping npm install"
+else
+  npm install -g @hashgraph/solo@"${fromSoloVersion}" --force
+fi
+
 SOLO_NO_CACHE=true npm install -g @hashgraph/solo@"${fromSoloVersion}" --force
```
jeromy-cannon: The `if/else` block above (lines 454-458) checks whether the requested Solo version is already installed and prints `skipping npm install` — but **this line always runs**, so the optimization has no effect. Either gate this install on the same condition, or drop the `if/else` entirely.

```suggestion
if [[ -n "${installedSoloVersion}" && "${installedSoloVersion}" == "${expectedSoloVersion}" ]]; then
  echo "Solo version ${installedSoloVersion} already installed, skipping npm install"
else
  SOLO_NO_CACHE=true npm install -g @hashgraph/solo@"${fromSoloVersion}" --force
fi
```

.github/workflows/script/launch_network.sh
```diff
@@ -34,10 +35,13 @@ on_exit() {
...
 
+  if [[ -n "${TEMP_ONE_SHOT_VALUES_FILE:-}" && -f "${TEMP_ONE_SHOT_VALUES_FILE}" ]]; then
+    rm -f "${TEMP_ONE_SHOT_VALUES_FILE}"
+  fi
+
   if [[ ${rc} -ne 0 ]]; then
     echo "Test failed, current port forward process: "
     ps -ef | grep port-forward
-    collect_failure_diagnostics "${rc}"
   fi
```
jeromy-cannon, line 46: The `collect_failure_diagnostics` call was deleted here but the function definition (line 10) is still in the file. Failure runs now collect a `ps -ef | grep port-forward` snapshot and nothing else. Was this intentional? If so, please remove the function too; if not, restore the call.

.github/workflows/script/launch_network.sh
```diff
@@ -101,6 +105,329 @@ show_service_ips() {
   kubectl get svc -n "${namespace}" network-node1-svc network-node2-svc -o custom-columns=NAME:.metadata.name,CLUSTER-IP:.spec.clusterIP,CREATED:.metadata.creationTimestamp
 }
 
+# Restart relay after upgrade, wait for rollout, and ensure localhost JSON-RPC
+# forwarding is stable before smoke tests.
+refresh_relay_network_config() {
```
jeromy-cannon, line 110: This is a workaround pattern — 18 × 5 s polling around `solo deployment refresh port-forwards` to confirm a single port-forward is healthy. If the command itself doesn't guarantee a working forward after returning, fix the command. We've already pushed the port-forward refresh into the Solo CLI so this kind of CI-side babysitting shouldn't be needed.

.github/workflows/script/launch_network.sh
```diff
@@ -101,6 +105,329 @@ show_service_ips() {
...
+
+  if [[ -z "${pgPassword}" ]]; then
+    echo "[IMPORTER_RESET] Could not get postgres password from secret solo-shared-resources-passwords, skipping"
+    return 0
+  fi
+
+  # Show current state for diagnostics
+  local currentHash
+  currentHash=$(kubectl exec -n "${namespace}" "${postgresPod}" -c postgresql -- \
+    sh -c "PGPASSWORD='${pgPassword}' psql -U postgres -d mirror_node -t -A -c \
```
jeromy-cannon: Two problems on this line:

1. **Password leak via argv** — `kubectl exec ... sh -c "PGPASSWORD='${pgPassword}' psql ..."` embeds the postgres password into the shell command and the container's argv. It will show up in `ps -ef` on the node and in any audit log that captures `kubectl exec`. Use `kubectl exec --env PGPASSWORD=...` (or pipe via stdin) instead.
2. **SQL string interpolation** — `UPDATE record_file SET hash = '${mismatchActualHash}'` builds SQL by concatenation. `mismatchActualHash` is regex-parsed from importer logs and is *expected* to be hex, but it isn't validated before substitution. If the upstream log format ever changes this becomes arbitrary SQL execution against mirror-node Postgres.

Applies in the same way at line 362.

.github/workflows/script/launch_network.sh
```diff
@@ -101,6 +105,329 @@ show_service_ips() {
...
+  fi
+}
+
+# Function to realign mirror importer hash chain in postgres to the first mismatched
+# record-file previous hash reported by importer logs.
+#
+# We must not set hash='' because newer importer parser paths still validate and will fail
+# with "Expected = , Actual = ...". Instead we set the last DB hash to the mismatch line's
+# "Actual" running-hash value to bridge the restart boundary deterministically.
+reset_importer_hash_chain_for_upgrade() {
```
jeromy-cannon, line 282: This whole block (and `auto_recover_importer_hash_chain` below) is a workaround for hiero-consensus-node#25486 — CN generating empty rounds after freeze that break the importer's hash chain after restart. Solo shouldn't know about the importer's signature-chain invariants, and CI shouldn't rewrite `record_file.hash` in mirror-node Postgres to paper over an upstream issue. This would be hard for our SREs and end-users to intuitively grasp.

Is there a consensus-node or solo-containers PR tracking the actual fix (suppress empty rounds after freeze, or expose an importer flag to skip the gap)? If so let's link to it and scope this PR down. If not, please open one and treat the SQL rewrite as a tracked temporary workaround with a clear removal plan, not a permanent diagnostic helper in the migration script.

src/core/account-manager.ts
```diff
@@ -630,6 +654,54 @@ export class AccountManager {
     }
   }
 
+  private async getGossipFqdnRestricted(
```
jeromy-cannon: This duplicates `helpers.resolveGossipFqdnRestricted` — same K8s ConfigMap → cache template → repo template cascade (just missing the staging-dir step because AccountManager doesn't have one). `commands/node/tasks.ts` was correctly refactored to delegate to the helper; please do the same here and pass `{k8: this.k8Factory.getK8(context), namespace, cacheDir: constants.SOLO_CACHE_DIR, resourcesDir: constants.RESOURCES_DIR}`. One implementation is much easier to keep aligned than three.

src/core/account-manager.ts
```diff
@@ -210,6 +211,7 @@ export class AccountManager {
     clusterReferences: ClusterReferences,
     deployment: DeploymentName,
     forcePortForward?: boolean,
+    gossipFqdnRestricted?: boolean,
```
jeromy-cannon: `gossipFqdnRestricted?: boolean` is threaded through `getNodeClient` → `refreshNodeClient` → `_getNodeClient` → `loadNodeServices`, but no caller in the diff passes a value, so `loadNodeServices` always falls through to its own `getGossipFqdnRestricted()` (line 515). Either remove the parameter from all four signatures or wire it from the actual call sites — right now it's dead code that gives the false impression of an injection point.

src/core/profile-manager.ts
```diff
@@ -922,12 +912,38 @@ export class ProfileManager {
...
       const applicationPropertiesContent: string = fs.readFileSync(applicationPropertiesPath, 'utf8');
       const parsedFromApplicationProperties: boolean | undefined =
-        this.parseGossipFqdnRestricted(applicationPropertiesContent);
+        helpers.parseGossipFqdnRestricted(applicationPropertiesContent);
       if (parsedFromApplicationProperties !== undefined) {
         return parsedFromApplicationProperties;
       }
     }
 
+    const cacheApplicationPropertiesPath: string = PathEx.join(
```
jeromy-cannon: This now implements the same K8s → staging → cache → repo cascade as `helpers.resolveGossipFqdnRestricted` and `account-manager.ts::getGossipFqdnRestricted`. Please call the shared helper instead of growing a third near-identical implementation.

test/unit/core/helpers.test.ts
```diff
@@ -250,4 +252,337 @@ describe('Helpers', (): void => {
...
+      const content: string = `
+# Configuration file
+some.other.property=value
+nodes.gossipFqdnRestricted=true
+another.property=123
+`;
+      expect(helpers.parseGossipFqdnRestricted(content)).to.equal(true);
+    });
+
+    it('is case-insensitive for true value', (): void => {
```
jeromy-cannon: Test name says "is case-**insensitive**" but the assertions show the regex matches lowercase only (`TRUE` → undefined, `True` → undefined, `true` → true). Rename to `is case-sensitive for true value` so the title matches the behavior. Same on line 294.

jeromy-cannon (dismissed):
## Intent

Fix flaky migration tests by simplifying first-run setup with `solo one-shot falcon deploy`, replacing kubectl port-forward chasing with the built-in `deployment refresh port-forwards`, and bridging a CN-after-freeze hash-chain mismatch in mirror-importer Postgres. Also adds a cluster-config → cache → repo template fallback for resolving `nodes.gossipFqdnRestricted` during address/gossip generation.

## Verdict

Request changes — there are two correctness bugs and several structural issues (DRY, dead-code parameter chain, workaround-shaped code paths) that should be sorted before merging.

## Critical

- **`src/core/helpers.ts:1-67`** — file now has two SPDX headers and the new function sits above all imports. See inline.
- **`.github/workflows/script/launch_network.sh:454-460`** — the new "skip npm install if already installed" branch is immediately followed by an *unconditional* `SOLO_NO_CACHE=true npm install` on line 460. The branch is dead. See inline.

## Major

- **DRY — three near-identical implementations of `gossipFqdnRestricted` resolution.** `helpers.ts::resolveGossipFqdnRestricted` (new), `account-manager.ts::getGossipFqdnRestricted` (new private), and `profile-manager.ts::getGossipFqdnRestricted` (extended) all walk K8s ConfigMap → staged file → cache template → repo template. `commands/node/tasks.ts` was correctly refactored to call the helper — please do the same in `AccountManager` and `ProfileManager` and delete the duplicates.
- **`gossipFqdnRestricted?: boolean` parameter chain is dead** — `getNodeClient`, `refreshNodeClient`, `_getNodeClient`, `loadNodeServices` all gained the parameter but no caller in the diff passes it. The resolution always falls back to `loadNodeServices`'s internal `getGossipFqdnRestricted`. Either remove the parameter or wire it from the call sites that actually have the value.
- **Importer hash-chain rewrite is a workaround for an upstream CN bug** (hiero-consensus-node#25486). Mutating `record_file.hash` from a CI shell script to bridge the freeze/restart boundary will surprise SREs and end-users — Solo shouldn't know about the importer's signature-chain invariants. Push the fix to consensus-node (suppress the post-freeze empty rounds) or surface it as an importer flag, and keep Solo out of the database.
- **Postgres password through `kubectl exec ... sh -c "PGPASSWORD='...' psql ..."`** (lines 336, 362) — embedding the password inside a `-c` string puts it on the container's argv, visible in `ps -ef` and any audit log that captures kubectl exec. Use `kubectl exec --env PGPASSWORD=... -- psql ...` or pipe SQL through stdin.
- **SQL string interpolation** — `UPDATE record_file SET hash = '${mismatchActualHash}' ...` builds SQL by concatenation. The hash value comes from regex-parsing importer logs and isn't validated before substitution. If the log format ever drifts, this becomes an arbitrary SQL execution against mirror-node Postgres.
- **Removed `collect_failure_diagnostics` call from `on_exit`** (line 45) while leaving the function definition in place. Failure runs now collect a `ps -ef | grep port-forward` snapshot and nothing else — diagnostics on CI failure regress. Either restore the call or delete the dead function.
- **Removed schemaVersion verification block.** The old verification (`local-config-after.yaml` contains `schemaVersion: 2`, `remote-config-after.yaml` contains `schemaVersion: 7`) was a hard gate on migration correctness. Removing it silently weakens the test. If `one-shot falcon` makes those assertions redundant, please say so; otherwise restore them.
- **`refresh_relay_network_config` polling loop** — 18×5s = up to 90 s of polling for a port-forward to "stabilize" after a relay rollout. If `solo deployment refresh port-forwards` doesn't guarantee a working forward, fix that command — adding retry probes around it just hides the real bug.
- **Loose typing on `resolveGossipFqdnRestricted` options** — `k8?: { configMaps: () => ... }` and `namespace?: unknown` sidestep the real `K8` and `NamespaceName` types we have in `src/integration/kube/`. See inline.

## Minor

- Three new exported functions in `helpers.ts` (`parseGossipFqdnRestricted`, `readGossipFqdnRestrictedFromFile`, `resolveGossipFqdnRestricted`). `helpers.ts` already has a lot of exported functions so this matches the file, but per checklist §2 a `GossipFqdnRestrictedHelper` class with static methods would be more consistent with the rest of `core/`.
- `helpers.test.ts:287` and `:294` — describe names say "is case-insensitive" but the assertions test that the regex is case-**sensitive**. See inline.
- `test/unit/commands/node/local-build-path-pvc-validation.test.ts` — gossip resolution tests don't belong in a PVC validation file. The helper-level tests already live in `helpers.test.ts`; the remaining `invokeGetGossipFqdnRestricted` tests can move to `tasks.test.ts`.
- `# Function to ...` block comments above each new bash function (lines 110, 144, 225, 299, 378, 414, 417) re-state the function name. Drop or fold into a one-line purpose comment.
- `grep -E '...37546:7546' | grep -v grep | wc -l` checks rely on the literal port number — if the port changes the check passes-by-accident as 0. `pgrep -af` or capturing the PID at start would be steadier.
- `SOLO_NAMESPACE` / `SOLO_DEPLOYMENT` renamed from `solo-e2e` → `one-shot`. CI-only, but the change touches multiple files — worth a heads-up in the PR description.

## Questions for the author

- Is there a tracked solo-containers / consensus-node PR that *removes* the post-freeze empty rounds? If so, that's the real fix and this PR can be reduced to the gossipFqdnRestricted resolution change.
- Why was `collect_failure_diagnostics` left in the file but unwired? Intentional partial revert?
- Was the `schemaVersion` verification block intentionally dropped or accidental? It's the only assertion that the migration actually advanced the schema.

## Positive

- Refactoring `parseGossipFqdnRestricted` out of `NodeCommandTasks` into shared helpers is the right call — three classes were going to need it. The unit-test coverage on `parseGossipFqdnRestricted` is genuinely thorough (whitespace, comments, similar property names, multiline).
- Replacing the brittle `kubectl get pods | grep relay | awk` style with label selectors + sorted creation timestamps in `get_active_importer_pod` is a real improvement.
- Generating the one-shot values file via `mktemp` with a paired `on_exit` cleanup is the right shape.

## Companion work

- solo-docs PR: N/A (no docs changes)
- solo-containers PR: **missing** — the hash-chain workaround belongs upstream. Please link a consensus-node / containers issue or PR that addresses the empty-rounds-after-freeze behavior, or scope this PR down.

src/core/helpers.ts
JeffreyDallas: Removed

src/core/helpers.ts
JeffreyDallas: Updated. resolveGossipFqdnRestricted now uses typed options (k8: K8, namespace: NamespaceName) instead of structural/unknown types.

.github/workflows/script/launch_network.sh
JeffreyDallas: Fixed. The unconditional install was removed; install now runs only in the else branch when the requested Solo version is not already installed.

.github/workflows/script/launch_network.sh
JeffreyDallas: Restored. collect_failure_diagnostics is called again from on_exit on failure.

.github/workflows/script/launch_network.sh
JeffreyDallas: Addressed by removing the extra CI polling loop. refresh_relay_network_config now does restart + rollout status + deployment refresh port-forwards without the 18x retry wrapper.

.github/workflows/script/launch_network.sh
JeffreyDallas: Hardened this flow: validated mismatchActualHash as 96 hex chars, stopped putting PG password on argv, and switched SQL to psql variable substitution instead of string interpolation.

.github/workflows/script/launch_network.sh
JeffreyDallas: Acknowledged. Added explicit tracking/removal notes in-script and linked this temporary workaround to solo#4492 and upstream hiero-consensus-node#25486.

src/core/account-manager.ts
JeffreyDallas: Refactored. account-manager now delegates to helpers.resolveGossipFqdnRestricted and the duplicate private implementation was removed.

src/core/account-manager.ts
JeffreyDallas: Removed the dead parameter chain. gossipFqdnRestricted was dropped from the loadNodeClient/refreshNodeClient path since it was not wired from call sites.

src/core/profile-manager.ts
JeffreyDallas: Refactored. profile-manager now uses the shared helpers.resolveGossipFqdnRestricted helper instead of a third near-identical cascade.

test/unit/core/helpers.test.ts
JeffreyDallas: Updated test names to match behavior: is case-sensitive for true value and is case-sensitive for false value.

jan-milenkov (commented):
lgtm, waiting for an approval from @jeromy-cannon

src/core/helpers.ts
```diff
@@ -85,6 +85,93 @@ export function splitFlagInput(input: string, separator: string = ','): string[]
...
+  applicationPropertiesPath?: string;
+}
+
+export async function resolveGossipFqdnRestricted(options: ResolveGossipFqdnRestrictedOptions): Promise<boolean> {
+  const {k8, namespace, stagingDir, cacheDir, resourcesDir, applicationPropertiesPath} = options;
+
+  // 1. K8s configMap
+  if (k8 && namespace) {
+    try {
+      const configMap: {data?: Record<string, string>} = await k8
```
jeromy-cannon: The pre-refactor code read this into the imported `ConfigMap` type; the inline `{data?: Record<string, string>}` loosens the typing. Prefer the real type:

```suggestion
      const configMap: ConfigMap = await k8
```
(re-add the `ConfigMap` import that `tasks.ts`/`profile-manager.ts` dropped).

src/core/account-manager.ts
```diff
@@ -491,6 +491,15 @@ export class AccountManager {
         );
       }
 
+      // Resolve once at this boundary so children inherit a single explicit value.
+      const k8Context: Optional<string> = clusterReferences.values().next().value;
```
jeromy-cannon: This resolves `gossipFqdnRestricted` from the *first* cluster reference and applies it to every node network-wide. For a dual-cluster deployment, is reading one cluster's configMap and applying it to all nodes correct?

.github/workflows/script/launch_network.sh
```diff
@@ -101,6 +106,316 @@ show_service_ips() {
...
+
+  if [[ -z "${pgPassword}" ]]; then
+    echo "[IMPORTER_RESET] Could not get postgres password from secret solo-shared-resources-passwords, skipping"
+    return 0
+  fi
+
+  # Show current state for diagnostics
+  local currentHash
+  currentHash=$(kubectl exec -n "${namespace}" "${postgresPod}" -c postgresql -- \
+    sh -c "PGPASSWORD='${pgPassword}' psql -U postgres -d mirror_node -t -A -c \
```
jeromy-cannon: This diagnostic query inlines `PGPASSWORD='${pgPassword}'` on the command line, so the password lands in the pod's process args. The actual `UPDATE` below already uses a `PGPASSFILE` heredoc — please use the same approach here for consistency and to keep the secret off the command line.

.github/workflows/script/launch_network.sh
```diff
@@ -175,71 +502,32 @@ fi
...
-solo consensus network deploy --deployment "${SOLO_DEPLOYMENT}" --pvcs --release-tag "${FROM_CONSENSUS_NODE_VERSION}" -q --dev
-solo consensus node setup --deployment "${SOLO_DEPLOYMENT}" --release-tag "${FROM_CONSENSUS_NODE_VERSION}" -q --dev
-solo consensus node start --deployment "${SOLO_DEPLOYMENT}" -q --dev
-solo ledger account create --deployment "${SOLO_DEPLOYMENT}" --hbar-amount 100 --dev
-
-solo mirror node add --deployment "${SOLO_DEPLOYMENT}" --cluster-ref ${SOLO_CLUSTER_NAME} --enable-ingress --pinger -q --mirror-node-version ${MIRROR_NODE_VERSION_PRIOR_TO_UPGRADE} --dev
-solo explorer node add --deployment "${SOLO_DEPLOYMENT}" --cluster-ref ${SOLO_CLUSTER_NAME} --explorer-version ${PREV_EXPLORER_VERSION} -q --dev
-solo relay node add -i node1,node2 --deployment "${SOLO_DEPLOYMENT}" --cluster-ref ${SOLO_CLUSTER_NAME} --relay-release ${PREV_RELAY_VERSION} --dev
+export ONE_SHOT_WITH_BLOCK_NODE=true
+solo one-shot falcon deploy \
```
jeromy-cannon, line 527: Switching the first run to `one-shot falcon deploy` is a nice simplification, but it removed the migration-verification block that used to live here — the `schemaVersion: 2` (local) and `schemaVersion: 7` (remote) assertions, plus `deployment config list`. This job is the *migration* test (closes #4365); without those assertions it can go green even if the schema migration silently doesn't run. Was the removal intentional, and if so what now guarantees the schema actually migrated?

JeffreyDallas: With current stabilized design, we do not expect significant change of local configy yaml and remote config yam. So no need to test it any more

Also if upgrade of solo itself and all components are working, then config file upgrade should also be working  as part of the system workflow

src/core/helpers.ts
JeffreyDallas: Updated. This now uses the concrete ConfigMap type again in helpers, matching the pre-refactor typing.

src/core/account-manager.ts
JeffreyDallas: Changed account-manager to resolve gossipFqdnRestricted per cluster context and cache it in a context->value map, so dual-cluster deployments no longer inherit the first cluster's setting across the whole network.

.github/workflows/script/launch_network.sh
JeffreyDallas: Fixed. The diagnostic SELECT now uses the same PGPASSFILE heredoc pattern as the UPDATE path, so the postgres password no longer lands in the pod process args.

jan-milenkov (approved):

PR #4443, opened 2026-05-21
feat: provision RSA bootstrap roster file when adding block nodes

### LINKED ISSUE #4426: Provision RSA bootstrap roster file when adding block nodes
### Feature to Enhance

When a block node starts up, it looks for an RSA bootstrap file at `app.state.rsaBootstrapFilePath` (default:
  `/opt/hiero/block-node/node/rsa-bootstrap-roster.json`). This file contains the RSA public keys of the consensus nodes that the block node
  trusts for block verification. If the file is absent at startup, the RsaRosterBootstrapPlugin falls back to fetching node public keys from a
  Hedera Mirror Node REST API.

  In Solo-managed deployments — especially private networks with no public Mirror Node — that fallback is unavailable. The operator must be
  able to supply the bootstrap roster at deploy time so the block node starts up with the correct trust roots.

Solo already generates and holds the exact key material required — the RSA gossip signing certificate for each consensus
  node is stored in `network-<nodeAlias>-keys-secrets` — but does not use it to populate the block node's bootstrap file.

### Requested Enhancement

Proposed Solution

  During `solo block-node add` or any solo process/command which provisions a block node, after consensus node keys are available, Solo should automatically:

  1. Enumerate the consensus nodes in the deployment from the remote config.
  2. For each node, read the gossip signing public key from its Kubernetes secret.
  3. Export the key as hex-encoded DER SPKI (the same extraction already done in `AccountManager.buildAddressBookBase64()`).
  4. Construct a `NodeAddressBook` JSON and store it in a Kubernetes ConfigMap.
  5. Mount the ConfigMap as a file at `/opt/hiero/block-node/node/rsa-bootstrap-roster.json` via the Helm chart before the JVM
  starts.

Bootstrap File Format
```
  {
    "nodeAddress": [
      { "RSAPubKey": "<hex-encoded DER SPKI RSA public key>", "nodeId": 4 },
      { "RSAPubKey": "<hex-encoded DER SPKI RSA public key>", "nodeId": 5 }
    ]
  }
```

  The RSAPubKey field is identical to the hex DER value already produced by `AccountManager.buildAddressBookBase64()`  from
  `s-public-<nodeAlias>.pem.`

  Acceptance Criteria

  - `solo block-node add` automatically constructs and mounts the bootstrap roster when consensus node keys exist in the
  namespace.
  - The block node's `RsaRosterBootstrapPlugin` skips the Mirror Node fetch at startup (no Mirror Node HTTP calls in pod logs).
  - When consensus node keys are not yet available (block node deployed pre-genesis), a warning is logged and bootstrapping
  is skipped; Mirror Node fallback applies.
  - `--block-node-rsa-bootstrap-file` optionally overrides the auto-generated file.
  - `solo block-node destroy` results in clean removal of the ConfigMap.
  - Unit tests cover roster construction and key extraction.
  - E2E test confirms the file reaches the pod and the node starts without Mirror Node calls.

### Business Need

Needed for CN to stream WRBs to BN and then BN to MN

jan-milenkov:
## Description

Adds RSA bootstrap roster support for block node deployments.

Solo now attempts to build an RSA bootstrap roster from consensus node gossip signing certificates when running `block node add`. If the roster can be generated, Solo passes it to the block-node Helm chart through the new chart-native `blockNode.rsaBootstrap` values. The chart then writes the roster file into the block-node pod before the JVM starts and exposes the configured file path through block-node config.

This also adds support for `--block-node-rsa-bootstrap-file`, allowing users to provide a roster JSON file directly instead of generating one from consensus node secrets.

If consensus node gossip signing certificates are not available yet, Solo skips RSA bootstrap roster generation and continues with the existing behavior.

### Related Issues

* Closes # https://github.com/hiero-ledger/solo/issues/4426
* Depends on # https://github.com/hiero-ledger/hiero-block-node/pull/2863

### DIFF
--- a/src/business/utils/k8-helper.ts
+++ b/src/business/utils/k8-helper.ts
@@ -46,4 +46,10 @@ export class K8Helper {
       .list(namespace, Templates.renderBlockNodeLabels(id))
       .then((pods: Pod[]): Pod => pods[0]);
   }
+
+  public async getBlockNodeContainer(namespace: NamespaceName, id: ComponentId): Promise<Container> {
+    return await this.getBlockNodePod(namespace, id)
+      .then((pod): ContainerReference => ContainerReference.of(pod.podReference, constants.BLOCK_NODE_CONTAINER_NAME))
+      .then((containerReference): Container => this.k8.containers().readByRef(containerReference));
+  }
 }
--- a/src/business/utils/k8-helper.ts
+++ b/src/business/utils/k8-helper.ts
@@ -46,10 +46,4 @@ export class K8Helper {
       .list(namespace, Templates.renderBlockNodeLabels(id))
       .then((pods: Pod[]): Pod => pods[0]);
   }
-
-  public async getBlockNodeContainer(namespace: NamespaceName, id: ComponentId): Promise<Container> {
-    return await this.getBlockNodePod(namespace, id)
-      .then((pod): ContainerReference => ContainerReference.of(pod.podReference, constants.BLOCK_NODE_CONTAINER_NAME))
-      .then((containerReference): Container => this.k8.containers().readByRef(containerReference));
-  }
 }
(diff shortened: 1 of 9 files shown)


jan-milenkov:
not needed anymore https://github.com/hiero-ledger/solo/issues/4426

PR #4532, opened 2026-06-02
fix: collect local logs when cluster or context is not reachable

### LINKED ISSUE #4505: Diagnostics tip is broken: 'diagnostics logs' fails without a kube context and 'diagnostics report' is not a valid command
### Describe the bug

When a command fails, Solo prints a tip suggesting two ways to collect diagnostics / auto-create a GitHub issue:

```
solo deployment diagnostics logs
solo deployment diagnostics report
```

Neither works in the situation where it's actually needed (a failed `one-shot` before any cluster exists):

1. `solo deployment diagnostics logs` errors with:
   ```
   No active kubernetes context found. Please set current kubernetes context.
   ```
   So when cluster creation fails (no kube context yet), the diagnostics command the tip points you to cannot run.

2. `solo deployment diagnostics report` — the advertised "create a GitHub issue in one step" command — is not actually a recognized command:
   ```
   Unknown arguments: deployment, diagnostics, report
   ```
   (`solo --help` lists no `diagnostics` subcommand under `deployment`.)

- The suggested diagnostics commands should exist and run successfully.
- `diagnostics logs` should degrade gracefully (collect local logs / config) when no active kube context is present, rather than hard-failing — since the most common time a user needs it is right after a setup/cluster failure.
- If `diagnostics report` / auto-issue creation isn't implemented, the tip should not advertise it.

### To Reproduce

1. Trigger any command failure so the diagnostics tip is shown (e.g. a failing `solo one-shot single deploy`)
2. Run `solo deployment diagnostics logs` → fails with "No active kubernetes context found"
3. Run `solo deployment diagnostics report` → "Unknown arguments: deployment, diagnostics, report"

### Additional Context

- Version: 0.74.0
- OS: macOS (Darwin 24.6.0), arm64
- Discovered while trying to report the `one-shot single deploy` kind-config failure.

JeffreyDallas:
## Description

* Makes `solo deployment diagnostics logs` / `all` / `report` **degrade gracefully** when the Kubernetes cluster cannot be reached, instead of hard-failing. Previously the diagnostics tip recommended these commands, but they failed with `No active kubernetes context found` (no context) or `ECONNREFUSED` (a stale context still pointing at a torn-down cluster) — exactly when a user needs diagnostics most, right after a setup/cluster failure.
* Adds `DiagnosticsCollector` (`src/commands/util/diagnostics-collector.ts`) with two static helpers:
  * `isKubeClusterReachable(k8Factory)` — verifies both that a current context exists **and** that the cluster actually answers an API call (`namespaces().list()`), so a stale context for a deleted cluster is correctly treated as unreachable.
  * `collectLocalDiagnostics(logger, outputDirectory)` — collects the artifacts available without a cluster (`solo.log`, `solo.ndjson`, `local-config.yaml`).
* When the cluster is unreachable, the diagnostics commands now collect local logs/config, run the existing analyzer over them, print a clear warning, and exit `0`.

### Related Issues

* Closes #4505
* Related to #

### Testing

* [ ] Anything not tested is documented

* `deployment diagnostics logs` with an empty `KUBECONFIG` (no context): degrades to local collection, exits `0`.
* `deployment diagnostics logs` with a kubeconfig whose `current-context` points at a dead cluster (`https://127.0.0.1:1`, connection refused): degrades to local collection, exits `0`, prints the "No reachable Kubernetes cluster" warning.
* Added `test/unit/commands/node/handlers-diagnostics-local.test.ts` (4 cases) covering both the no-context and stale-context/dead-cluster paths for `logs` and `all`; existing `handlers-report.test.ts` still passes.

The following was not tested:

* End-to-end against a live multi-node deployment (existing behavior with a reachable cluster is unchanged).


src/commands/node/handlers.ts
```diff
@@ -106,6 +107,61 @@ export class NodeCommandHandlers extends CommandHandler {
     }
   }
 
+  /**
+   * Resolves a deployment name using only locally available information (the
+   * deployment flag and local config) without contacting any cluster. Returns an
+   * empty string when no unambiguous deployment can be determined locally.
+   */
+  private async resolveDeploymentNameWithoutCluster(argv: ArgvStruct): Promise<string> {
```
jeromy-cannon, line 131: This seems like a duplicate of `resolveDeploymentForLogs` — the flag check and the valid-deployments filter loop are copy-pasted from lines 722-734. Consider extracting a shared private helper and calling it from both, e.g.:

```suggestion
  private async collectValidLocalDeployments(): Promise<Deployment[]> {
    await this.localConfig.load();
    const validDeployments: Deployment[] = [];
    for (const deployment of this.localConfig.configuration.deployments) {
      if (deployment?.name && deployment.name.trim().length > 0) {
        validDeployments.push(deployment);
      }
    }
    return validDeployments;
  }

  private async resolveDeploymentNameWithoutCluster(argv: ArgvStruct): Promise<string> {
```

…then have `resolveDeploymentForLogs` use the same helper for its local-config branch. Checklist §1 / DRY.

src/commands/util/diagnostics-collector.ts
```diff
@@ -0,0 +1,100 @@
...
+  public static async isKubeClusterReachable(k8Factory: K8Factory): Promise<boolean> {
+    try {
+      const k8: K8 = k8Factory.default();
+      if (!k8.contexts().readCurrent()) {
+        return false;
+      }
+
+      // Confirm the cluster actually answers; a stale context for a deleted cluster
+      // throws here (e.g. connection refused) and must be treated as unreachable.
+      await k8.namespaces().list();
```
jeromy-cannon, line 71: The bare `catch` below treats *any* error here as "unreachable" — including an RBAC/authorization failure. A user with a reachable cluster who simply can't list namespaces cluster-wide would silently fall back to local-only collection and never learn that remote logs were skipped. Is that the intended behavior, or should we distinguish a connectivity failure (ECONNREFUSED) from an authz failure so the latter still surfaces? A lighter probe scoped to the deployment namespace might also avoid the cluster-wide list permission entirely.

src/commands/node/handlers.ts
```diff
@@ -670,6 +726,11 @@ export class NodeCommandHandlers extends CommandHandler {
 
   public async logs(argv: ArgvStruct): Promise<boolean> {
     argv = helpers.addFlagsToArgv(argv, NodeFlags.LOGS_FLAGS);
+
+    if (!(await DiagnosticsCollector.isKubeClusterReachable(this.k8Factory))) {
+      return await this.collectLocalDiagnosticsOnly(argv);
```
jeromy-cannon: Backwards-compat question: previously these commands exited non-zero when the cluster was unreachable, and now they exit 0. Scripts/CI that branch on the exit code to detect "diagnostics couldn't reach the cluster" will now see success. Is exit 0 the contract we want here, or should the degraded path still signal a distinct exit code? Checklist §7.

src/commands/node/handlers.ts
```diff
@@ -879,11 +945,18 @@ export class NodeCommandHandlers extends CommandHandler {
...
+    // Resolve deployment before calling collectDebug() so it's available for the issue title/body.
+    // Without an active kube context, resolve from local information only so the command still
+    // produces a report (collectDebug() degrades to local-only diagnostics in that case).
+    let deployment: string;
+    if (argv[flags.deployment.name]) {
+      deployment = String(argv[flags.deployment.name]);
+    } else if (await DiagnosticsCollector.isKubeClusterReachable(this.k8Factory)) {
+      deployment = await this.resolveDeploymentForLogs(argv);
+    } else {
+      deployment = await this.resolveDeploymentNameWithoutCluster(argv);
```
jeromy-cannon: `resolveDeploymentNameWithoutCluster` was added specifically for this `report` path, but no unit test exercises `report()` with an unreachable cluster — the two new tests only cover `logs` and `all`. Please add a `report`-with-unreachable-cluster case so this branch is covered. Checklist §15.

Also minor: `report` probes `isKubeClusterReachable` here, then `debug → all` probes it again — two live `namespaces().list()` round-trips for one command. Could thread the result through if it's easy.

src/commands/node/handlers.ts
JeffreyDallas: Updated with shared function `collectValidLocalDeployments()`

src/commands/util/diagnostics-collector.ts
JeffreyDallas: Updated

src/commands/node/handlers.ts
JeffreyDallas: Exit 0 happens **only** when the cluster is genuinely unreachable (no context, or a connection-level failure) — which is exactly the scenario #4505 asked us to handle gracefully, and where the command still did useful work (collected + analyzed local logs). Any case where the API server actually responds (authz, 5xx, etc.) is no longer swallowed: it flows through the normal path and exits non-zero as before.

So the only behavior change vs. `main` is: `unreachable cluster` went from non-zero → 0. I believe exit 0 is the right contract there (it matches the issue intent, and the degraded path is clearly announced on stderr with the reason), but happy to gate it behind a flag / distinct exit code if you'd prefer to keep a machine-detectable "couldn't reach cluster" signal — let me know.

src/commands/node/handlers.ts
JeffreyDallas: Added the test in ea7aeb48c: `report resolves the deployment from local config when the cluster is unreachable` exercises the `else` branch — it asserts `resolveDeploymentForLogs` (the cluster path) is **not** called, that the deployment is resolved locally (`solo-deployment`), and that `runDiagnosticsReport` is still driven with it.

On the double probe: I left it as-is for now. Threading the result through `runDiagnosticsReport → collectDebug → debug → all` would couple the reporter/debug/all chain to a pre-computed reachability flag, and `all`/`debug` are also standalone command entry points that must probe on their own. The probe is a single `namespaces().list()` and returns immediately on `ECONNREFUSED`, so the duplicate cost is negligible. Open to threading it if you feel strongly, but it didn't seem worth the added coupling.

test/unit/commands/node/handlers-diagnostics-local.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    expect(collectLocalDiagnosticsStub).to.have.been.calledOnce;
+    expect(analyzeStub).to.have.been.calledOnce;
+    expect(resolveDeploymentForLogsStub).to.not.have.been.called;
+    expect(initializeStub).to.not.have.been.called;
+  });
+
+  it('all collects local diagnostics when the context is stale and the cluster is unreachable', async (): Promise<void> => {
+    defaultK8Stub.returns(makeK8WithListError(connectionRefusedError()));
+
+    const argv: ArgvStruct = {_: ['deployment', 'diagnostics', 'all']} as unknown as ArgvStruct;
```
jan-milenkov: dont hardcode the command use the `DeploymentCommandDefintion`

test/unit/commands/node/handlers-diagnostics-local.test.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    expect(result).to.equal(true);
+    expect(collectLocalDiagnosticsStub).to.have.been.calledOnce;
+    expect(analyzeStub).to.have.been.calledOnce;
+    expect(resolveDeploymentForLogsStub).to.not.have.been.called;
+    expect(initializeStub).to.not.have.been.called;
+  });
+
+  it('all collects local diagnostics when no active kube context is present', async (): Promise<void> => {
+    const argv: ArgvStruct = {_: ['deployment', 'diagnostics', 'all']} as unknown as ArgvStruct;
```
jan-milenkov: dont hardcode the command use the `DeploymentCommandDefintion`

test/unit/commands/node/handlers-diagnostics-local.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    // Must not attempt cluster-dependent deployment resolution or the initialize task.
+    expect(resolveDeploymentForLogsStub).to.not.have.been.called;
+    expect(initializeStub).to.not.have.been.called;
+  });
+
+  it('logs collects local diagnostics when the context is stale and the cluster is unreachable', async (): Promise<void> => {
+    // Context exists (kubeconfig entry survives the cluster) but the API call is refused.
+    defaultK8Stub.returns(makeK8WithListError(connectionRefusedError()));
+
+    const argv: ArgvStruct = {_: ['deployment', 'diagnostics', 'logs']} as unknown as ArgvStruct;
```
jan-milenkov: dont hardcode the command use the `DeploymentCommandDefintion`

test/unit/commands/node/handlers-diagnostics-local.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    commandActionStub = sinon.stub(NodeCommandHandlers.prototype, 'commandAction').resolves();
+    runDiagnosticsReportStub = sinon.stub(DiagnosticsReporter, 'runDiagnosticsReport').resolves();
+  });
+
+  afterEach((): void => {
+    sinon.restore();
+  });
+
+  it('logs collects local diagnostics when no active kube context is present', async (): Promise<void> => {
+    const argv: ArgvStruct = {_: ['deployment', 'diagnostics', 'logs']} as unknown as ArgvStruct;
```
jan-milenkov: dont hardcode the command use the `DeploymentCommandDefintion`

test/unit/commands/node/handlers-diagnostics-local.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    expect(resolveDeploymentForLogsStub).to.not.have.been.called;
+    expect(initializeStub).to.not.have.been.called;
+  });
+
+  it('logs does NOT degrade when the cluster responds with an authorization error', async (): Promise<void> => {
+    // The server answered (HTTP 403) -> the cluster is reachable; the real error must surface
+    // through the normal collection path instead of being hidden by a local-only fallback.
+    defaultK8Stub.returns(makeK8WithListError(forbiddenError()));
+
+    const argv: ArgvStruct = {_: ['deployment', 'diagnostics', 'logs']} as unknown as ArgvStruct;
```
jan-milenkov: dont hardcode the command use the `DeploymentCommandDefintion`

test/unit/commands/node/handlers-diagnostics-local.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    expect(result).to.equal(true);
+    expect(collectLocalDiagnosticsStub).to.not.have.been.called;
+    expect(resolveDeploymentForLogsStub).to.have.been.calledOnce;
+    expect(initializeStub).to.have.been.calledOnce;
+  });
+
+  it('report resolves the deployment from local config when the cluster is unreachable', async (): Promise<void> => {
+    defaultK8Stub.returns(makeK8WithListError(connectionRefusedError()));
+
+    const argv: ArgvStruct = {_: ['deployment', 'diagnostics', 'report'], quiet: true} as unknown as ArgvStruct;
```
jan-milenkov: dont hardcode the command use the `DeploymentCommandDefintion`

src/commands/util/diagnostics-collector.ts
```diff
@@ -0,0 +1,155 @@
...
+   * The diagnostics commands use this to degrade to local-only collection instead of
+   * hard-failing, which is the common situation right after a cluster is deleted.
+   */
+  public static async isKubeClusterReachable(k8Factory: K8Factory): Promise<ClusterReachability> {
+    let k8: K8;
+    try {
+      k8 = k8Factory.default();
+    } catch {
+      // k8Factory.default() throws when there is no active context/cluster in the kubeconfig.
+      return {reachable: false, reason: 'no active Kubernetes context'};
```
jan-milenkov: you have no guarantees that if k8Factory.default() throws, this is caused by no active Kubernetes context

JeffreyDallas: Updated

test/unit/commands/node/handlers-diagnostics-local.test.ts
JeffreyDallas: Updated

test/unit/commands/node/handlers-diagnostics-local.test.ts
JeffreyDallas: Updated

test/unit/commands/node/handlers-diagnostics-local.test.ts
JeffreyDallas: Updated

test/unit/commands/node/handlers-diagnostics-local.test.ts
JeffreyDallas: Updated

test/unit/commands/node/handlers-diagnostics-local.test.ts
JeffreyDallas: Updated

test/unit/commands/node/handlers-diagnostics-local.test.ts
JeffreyDallas: Updated

jan-milenkov (approved):

jan-milenkov (approved):

PR #4599, opened 2026-06-09
fix(remote-config): make component creation idempotent on existing component

### LINKED ISSUE #4560: fix(remote-config): make component creation idempotent on existing component
## Summary
  
  The remote-config component creation step currently throws when a component of the requested type already exists. Phase 1 changes this to log and return successfully; the phase-gated guards on later steps handle whether subsequent work
  needs to run.
  
  ## Scope
  
  - Detect "component already exists" and short-circuit with a log line.
  - Preserve existing validation for malformed inputs.
  
  ## Files to touch
  
  - `src/core/config/remote/components-data-wrapper.ts`
  - `src/business/runtime-state/config/remote/remote-config-runtime-state.ts` (caller)
  - Unit tests for the existing-component short-circuit.
  
  ## Acceptance criteria
  
  - Re-creating a component of an existing type returns success with an `INFO` log.
  - Component arrays are not duplicated or corrupted on re-run. 
  - Phase-gated guards on consensus/block/mirror/explorer/relay steps (Issues 5, 6) are unaffected.

jan-milenkov:
## Description

- Updated remote config component creation to be idempotent.
- The component wrapper now skips existing components instead of throwing, returns whether a component was actually added, and avoids incrementing component counters when nothing was added. - Callers can now log and continue safely on reruns, while malformed input validation remains unchanged.

### Related Issues

* Closes # https://github.com/hiero-ledger/solo/issues/4560

### DIFF
--- a/src/business/runtime-state/config/remote/remote-config-runtime-state.ts
+++ b/src/business/runtime-state/config/remote/remote-config-runtime-state.ts
@@ -270,15 +270,23 @@ export class RemoteConfigRuntimeState implements RemoteConfigRuntimeStateApi {
 
     //* add the new nodes to components
     for (const nodeAlias of nodeAliases) {
-      this.configuration.components.addNewComponent(
-        componentFactory.createNewConsensusNodeComponent(
-          Templates.renderComponentIdFromNodeAlias(nodeAlias),
-          clusterReference,
-          namespace,
-          DeploymentPhase.REQUESTED,
-        ),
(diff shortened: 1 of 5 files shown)


JeffreyDallas (commented):
If interrupt a `block node add` solo command,
then run `block node add` again, does it add new block node 2 to the network or retry finishing add block node 1.
There seems many corner cases here depends on which step of block node add being interrupted

JeffreyDallas (approved):

PR #4647, opened 2026-06-12
fix: fix empty `block-nodes.json` race in parallel deploy

### LINKED ISSUE #4644: oneshot failed get relay pod ready
### Describe the bug

Since we enabled the solo cache image pull function, Intermittently one shot would fail, waiting for the relay part to get ready. 
There seems to be a timing issue. 

wait for relay pod ready

### To Reproduce

Should not fail

### Additional Context

_No response_

JeffreyDallas:
## Summary

Fixes #4644

When `parallelDeploy: true`, the one-shot orchestrator runs `Deploy block node` and `Deploy network node` (including `node setup`) concurrently. `node setup`'s `updateBlockNodesJson()` was reading `blockNodeMap` before `block-node add`'s `updateConsensusNodesInRemoteConfig()` had populated it, so `block-nodes.json` landed on the consensus node as `{"nodes":[]}`. The node started without any block node connections.

**Root cause (confirmed from `solo.log`):**
- `block-node add` helm-installs the block node (~20–30 s) and only calls `updateConsensusNodesInRemoteConfig()` after the pod is healthy
- Concurrently, `node setup` starts, loads `remoteConfig`, and runs `updateBlockNodesJson()` ~8 s before the block node pod is even ready
- `block-node add` skips re-copying `block-nodes.json` because `ledgerPhase === UNINITIALIZED`
- Result: permanent empty `block-nodes.json` on the node

**Fix — event-gate approach (no change to `parallelDeploy`):**
- New `BlockNodeDeployed` event type + `BlockNodeDeployedEvent` class
- `block-node add` emits the event after `handleConsensusNodeUpdating()` finishes (blockNodeMap persisted)
- The skip callback in the one-shot orchestrator emits the event immediately when block-node deploy is skipped (not needed / already deployed), so the gate resolves from history and there is **zero extra latency** on sequential or retry paths
- `withWaitCondition(SoloEventType.BlockNodeDeployed, 10 min)` added to the **"Deploy consensus node"** phase — `consensus network deploy` has its own `createAndCopyBlockNodeJsonFileForConsensusNode` call, so gating that phase ensures `blockNodeMap` is fully populated before either `node setup` or `network deploy` copy step runs

Also adds `BlockNodesJsonEmptySoloError` (SOLO-5074) as an early-fail guard in `createAndCopyBlockNodeJsonFileForConsensusNode`: if the generated `nodes` array is empty the function throws immediately with actionable troubleshooting steps instead of silently deploying broken config.

**Independent postgres fixes (found while debugging, unrelated to the race):**
- `obj_description` → `shobj_description` for the DB sentinel check (correct system catalog function for schemas)
- Added `mirror_rest` to the roles dropped during partial initialization cleanup

## Test plan

- [ ] Run `task test-e2e-standard` — existing one-shot test suite
- [ ] Run one-shot with `parallelDeploy: true` and confirm `block-nodes.json` contains node entries on the consensus node
- [ ] Run one-shot with `parallelDeploy: false` (sequential) and confirm no regression / no extra wait time
- [ ] Retry scenario: run deploy → destroy → deploy again and confirm block-nodes.json is correct on second deploy

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jan-milenkov (dismissed):
If this has negative effect on performance, inform managers before proceeding with merge

JeffreyDallas:
> If this has negative effect on performance, inform managers before proceeding with merge

no negative effect on oneshot timing or performance memory foot print

jan-milenkov (dismissed):

JeffreyDallas:
Wait #4798 to fix migration test

src/commands/one-shot/orchestrator/deploy/default-one-shot-deploy-orchestrator.ts
```diff
@@ -610,20 +611,30 @@ export class DefaultOneShotDeployOrchestrator implements OneShotDeployOrchestrat
...
-                (): string[] => DeployArgvBuilders.buildBlockNodeArgv(getConfig()),
-                this.taskList,
-                (): boolean =>
-                  !DeployArgvBuilders.shouldDeployBlockNode(getConfig()) ||
+            asListrTask: (
+              getConfig: () => OneShotSingleDeployConfigClass,
+            ): SoloListrTask<OneShotSingleDeployContext> => {
+              const skipAndNotify: () => boolean = (): boolean => {
+                const shouldSkip: boolean =
+                  constants.ONE_SHOT_WITH_BLOCK_NODE.toLowerCase() !== 'true' ||
```
jeromy-cannon: this uses `constants.ONE_SHOT_WITH_BLOCK_NODE` directly, but the original skip used `!DeployArgvBuilders.shouldDeployBlockNode(getConfig())` — the established API for this check. They are equivalent today since `shouldDeployBlockNode` ignores config entirely, but now there are two parallel paths for the same predicate.

consider:
```suggestion
                  !DeployArgvBuilders.shouldDeployBlockNode(getConfig()) ||
```
or alternatively update `isBlockNodeEnvironmentEnabled()` inside `DeployArgvBuilders` to delegate to `constants.ONE_SHOT_WITH_BLOCK_NODE` and call `shouldDeployBlockNode` here.

src/commands/one-shot/orchestrator/deploy/default-one-shot-deploy-orchestrator.ts
```diff
@@ -638,7 +649,10 @@ export class DefaultOneShotDeployOrchestrator implements OneShotDeployOrchestrat
                   // Idempotency guard: skip if the consensus node is already deployed or the network Helm release exists.
                   (): boolean => this.isConsensusDeployStepComplete(deploymentStateSnapshot),
                 ),
-            }),
+              // consensus network deploy has a "Copy block-nodes.json" step that reads blockNodeMap.
+              // Gate it on BlockNodeDeployed so block-node add has fully populated blockNodeMap before
+              // consensus network deploy runs.
+            }).withWaitCondition(SoloEventType.BlockNodeDeployed, Duration.ofMinutes(10)),
```
jeromy-cannon, line 655: the PR description says this `withWaitCondition` is on the **"Setup consensus node"** phase, but it's on **"Deploy consensus node"**. the placement here is actually correct — `network deploy` also calls `createAndCopyBlockNodeJsonFileForConsensusNode` (see the `false` added in `network.ts`), and gating the first phase in the composite transitively covers setup too. the code comment is accurate. please update the PR description to match.

src/core/helpers.ts
```diff
@@ -740,10 +741,15 @@ export class Helpers {
 
     const blockNodesJsonData: string = new BlockNodesJsonWrapper(blockNodeMap, externalBlockNodeMap).toJSON();
 
+    const parsedBlockNodesJson: {nodes: unknown[]} = JSON.parse(blockNodesJsonData) as {nodes: unknown[]};
```
jeromy-cannon, line 744: the new guard is independently testable — construct a `ConsensusNode` stub with empty `blockNodeMap` / `externalBlockNodeMap`, call `createAndCopyBlockNodeJsonFileForConsensusNode(..., false, ...)`, and assert `BlockNodesJsonEmptySoloError` is thrown. a unit test would catch regressions much faster than the E2E suite. consider adding one alongside the existing `postgres.test.ts` coverage in this PR.

src/core/shared-resources/postgres.ts
```diff
@@ -207,7 +207,7 @@ export class PostgresSharedResource {
           '# Check for the sentinel comment that marks a fully completed initialization.',
           '# Using a DB comment means the sentinel survives pod restarts and is only written',
           '# after init-postgres.sh completes successfully (see end of this script).',
-          `SENTINEL=$(psql -tc "SELECT obj_description(oid, 'pg_database') FROM pg_database WHERE datname = '${databaseName}'" 2>/dev/null | tr -d '[:space:]')`,
+          `SENTINEL=$(psql -tc "SELECT shobj_description(oid, 'pg_database') FROM pg_database WHERE datname = '${databaseName}'" 2>/dev/null | tr -d '[:space:]')`,
```
jeromy-cannon, line 210: these postgres changes (`shobj_description` and the `mirror_rest` addition at line 224) look unrelated to the block-nodes.json race fix. were they found while debugging this issue? if so, worth calling them out separately in the PR description so reviewers know they're intentional independent fixes.

src/commands/one-shot/orchestrator/deploy/default-one-shot-deploy-orchestrator.ts
JeffreyDallas: updated

src/commands/one-shot/orchestrator/deploy/default-one-shot-deploy-orchestrator.ts
JeffreyDallas: PR description updated

src/core/helpers.ts
JeffreyDallas: unit test added

src/core/shared-resources/postgres.ts
JeffreyDallas: PR description updated

jan-milenkov (approved):

PR #4650, opened 2026-06-13
fix: propagate shared resource values and rerun mirror postgres init

### LINKED ISSUE #4649: Mirror node add rerun skips shared Postgres initialization after partial install failure
## Problem

When `solo mirror node add` fails after installing the shared resources chart but before mirror-node Postgres initialization completes, a retry can skip the internal Postgres setup tasks.

The retry sees the existing `solo-shared-resources` Helm release and sets `installSharedResources=false`. The mirror-node flow currently uses that flag to skip both:

- `Initialize Postgres pod`
- `Prime mirror-node postgres secret`

That leaves the shared Postgres pod running, but without the mirror-node database roles/users initialized. The later mirror-node pods can then fail readiness because the expected DB users were never created.

## Root Cause

`installSharedResources` means "the shared-resources Helm release needs to be installed in this run". It does not mean "the internal shared Postgres is already initialized for mirror-node".

The Postgres initialization path is already designed to be idempotent through the `solo-initialized` database comment sentinel, so it should run for internal database deployments even when the shared-resources chart already existed before the current run.

## Expected Behavior

For internal database deployments, mirror-node add/upgrade should always prime the mirror password secret and invoke shared Postgres initialization. The initialization code should decide whether work is needed.

For external database deployments, these shared Postgres tasks should still be skipped.

JeffreyDallas:
## Summary

- Propagate mirror-node values into the shared-resources Helm chart so Postgres/Redis inherit scheduling values such as tolerations and node selectors.
- Add a Redis role scheduling fallback so shared Redis inherits `solo.hashgraph.io/role` scheduling from mirror-node component values when Redis does not define it explicitly.
- Propagate top-level scheduling values into mirror pinger and the separate mirror/explorer HAProxy ingress controller charts.
- Apply explorer `--ingress-controller-value-file` consistently for explorer ingress controller installs.
- Run internal mirror-node Postgres secret priming and database initialization on reruns even when the shared-resources Helm release already exists.
- Add unit coverage for shared-resource scheduling propagation, Helm scheduling propagation, and rerun skip behavior.

## Root Cause

There were four related failure modes:

1. Mirror-node values were not propagated to the shared-resources chart, so shared Postgres/Redis pods could miss cluster scheduling values and remain Pending on tainted performance clusters.
2. The performance cluster also taints nodes by `solo.hashgraph.io/role`. The mirror-node values put that role scheduling on mirror-node components, but Redis did not define role scheduling directly. Redis therefore still lacked the role toleration after owner/network propagation was fixed.
3. Mirror pinger is a mirror chart component that does not inherit top-level `nodeSelector`/`tolerations`, and mirror/explorer HAProxy ingress controllers are installed as separate Helm releases. Those pods could remain Pending even after shared resources were fixed.
4. After a partial failure, a retry set `installSharedResources=false` because the shared-resources release already existed. The mirror-node command used that flag to skip Postgres initialization, even though the mirror-node DB roles/users may not have been created yet.

`installSharedResources` only tells us whether the shared-resources chart needs to be installed in this run. It is not a valid signal that mirror-node Postgres has already been initialized. The existing initialization code is idempotent through the `solo-initialized` database comment sentinel, so it is safe to invoke for every internal DB deployment.

Closes #4649.

## Validation

- `npx mocha 'test/unit/core/shared-resources/shared-resource-manager.test.ts'`
- `MOCHA_SUITE_NAME="MirrorNodeCommand" npx mocha 'test/unit/commands/mirror-node.test.ts'`
- `npx mocha test/unit/core/util/helm-scheduling-values.test.ts test/unit/commands/mirror-node.test.ts test/unit/commands/explorer.test.ts`
- `npx tsc --noEmit --pretty false`
- `npx eslint src/core/shared-resources/shared-resource-manager.ts test/unit/core/shared-resources/shared-resource-manager.test.ts`
- `npx eslint src/commands/mirror-node.ts test/unit/commands/mirror-node.test.ts`
- `task build`
- `task format`
- `task check`

`task build`, `task format`, and `task check` completed successfully with the repository's existing lint/typedoc warnings.


JeffreyDallas:
Migration CI follow-up: the failure was caused by the Postgres init wrapper reading the database sentinel with `obj_description(oid, 'pg_database')`. Database comments are shared-object comments in PostgreSQL, so this returned empty even after Solo wrote `COMMENT ON DATABASE mirror_node IS 'solo-initialized'`. During `mirror node upgrade`, Solo treated the initialized DB as partial, dropped/recreated the DB, left the legacy `mirror_rest` role behind, and the importer later failed Flyway V1 with `ERROR: role "mirror_rest" already exists`.

Commit `d730739a2` fixes this by reading the sentinel with `shobj_description(oid, 'pg_database')` and including `mirror_rest` in partial-cleanup role handling. Validation passed in a clean PR worktree: focused Postgres shared-resource unit test, `npx tsc --noEmit --pretty false`, `task build`, `task format`, and `task check` all completed successfully with existing repository warnings.

src/core/shared-resources/shared-resource-manager.ts
jan-milenkov, line 1: Avoid creation of a bunch of rogue function exports

JeffreyDallas: updated

jan-milenkov, line 1: Cleanup rogue function exports, look for opportunity to make the code more readable, avoid unnecessary complexion

JeffreyDallas: updated

src/core/util/helm-scheduling-values.ts
jan-milenkov, line 1: Files exporting rogue functions are unwanted, if you need help with figuring out how to where/and how to refactor this, feel free to reach-out to teammembers

JeffreyDallas: updated.
Added new rule to coding style to enforce the rogue function requirment

JeffreyDallas:
Wait #4798 to fix migration test

src/core/shared-resources/shared-resource-manager.ts
```diff
@@ -9,19 +9,22 @@ import {type ChartManager} from '../chart-manager.js';
...
-
-interface SharedResourceSchedulingValues {
-  postgresNodeSelector: HelmMapValue;
-  postgresTolerations: HelmToleration[];
-  redisNodeSelector: HelmMapValue;
-  redisTolerations: HelmToleration[];
-}
+import {HelmSchedulingValues} from '../util/helm-scheduling-values.js';
+
+const ROLE_SCHEDULING_KEY: string = 'solo.hashgraph.io/role';
```
jeromy-cannon: These constants are only consumed by `buildSchedulingChartValues` below. When that function becomes a `private static` method of `SharedResourceManager` (see comment at line 131), these should become `private static readonly` class members too.

src/core/shared-resources/shared-resource-manager.ts
```diff
@@ -126,152 +129,18 @@ export class SharedResourceManager {
 }
 
 function buildSchedulingChartValues(sourceChartValues: HelmChartValues): HelmChartValues {
```
jeromy-cannon: Module-scope helper used only by `SharedResourceManager` — consider moving to a `private static` method of the class, consistent with how the extracted helpers in `HelmSchedulingValues` are structured. (Review checklist §3)

src/core/util/helm-scheduling-values.ts
```diff
@@ -0,0 +1,270 @@
...
+type HelmValuesObject = Record<string, unknown>;
+type HelmMapValue = SchedulingValues['nodeSelector'];
+type HelmToleration = SchedulingValues['tolerations'][number];
+
+interface SchedulingValues {
+  nodeSelector: Record<string, HelmChartValue>;
+  tolerations: Record<string, HelmChartValue>[];
+}
+
+interface HelmSchedulingValueMapping {
```
jeromy-cannon: `HelmSchedulingValueMapping` (and `HelmSchedulingValueFallback` at line 23) appear in the public signature of `buildMappedSchedulingChartValues` but are not exported. Callers who want to build the `mappings` array in a named variable can't type it. Consider exporting both interfaces.  And, exported interfaces should be in their own files kebab naming style matching the interface name.

src/commands/explorer.ts
```diff
@@ -439,9 +445,7 @@ export class ExplorerCommand extends BaseCommand {
           'controller.extraArgs.controller-class',
           config.ingressReleaseName,
         );
-        if (config.tlsClusterIssuerType === 'self-signed') {
-          explorerIngressControllerChartValues.filesFromCommaSeparatedInput(config.ingressControllerValueFile);
-        }
+        explorerIngressControllerChartValues.filesFromCommaSeparatedInput(config.ingressControllerValueFile);
```
jeromy-cannon, line 448: Just confirming: removing the `tlsClusterIssuerType === 'self-signed'` guard means `ingressControllerValueFile` is now applied for every TLS type. Are there any non-self-signed TLS deployments that previously relied on the file being silently skipped here?

docs/contributing/typescript-code-style.md
```diff
@@ -138,6 +138,10 @@ Do not use default exports.
 
 Only export symbols used outside of the module. Minimize exported API surface.
 
+Files exporting rogue functions are unwanted. Do not export implementation helpers, parser details, or
```
jeromy-cannon: "rogue functions" reads a bit informal for style-guide prose. Consider "module-scope behavior functions."

src/commands/explorer.ts
JeffreyDallas: Yes, intentional — `--ingress-controller-value-file` is a general-purpose values override for the HAProxy ingress chart, independent of TLS type. The original guard was a mistake that silently dropped user-provided values for any non-self-signed TLS deployment. The fix removes the guard so the file is always applied when provided.

src/core/shared-resources/shared-resource-manager.ts
JeffreyDallas: updated

src/core/shared-resources/shared-resource-manager.ts
JeffreyDallas: updated

src/core/util/helm-scheduling-values.ts
JeffreyDallas: updated

docs/contributing/typescript-code-style.md
JeffreyDallas: updated

alex-kuzmin-hg (approved):
Tested, it works well

jan-milenkov (approved):

PR #4773, opened 2026-06-23
fix: select the native package manager per Linux distribution

### LINKED ISSUE #4676: brew install hiero-ledger/tools/solo fails on Fedora Linux (no bottle, source build also fails)
**Reporters:** Henry Tong

`brew install hiero-ledger/tools/solo` fails on Fedora Linux because no pre-built bottle is available for the platform and the source build requires a C compiler that is also difficult to install via Homebrew on Fedora. This is the first Linux (non-Ubuntu) submission in the Solo user testing program.

**Error observed**
```
Error: The following formula cannot be installed from bottle and must be
built from source.
  solo
Install Clang or run `brew install gcc`.
```

Attempting `brew install gcc` also fails:
```
Warning: The post-install step did not complete successfully
You can try again using:
  brew postinstall gcc
```

**Affects:** Linux (Fedora), brew install path

**Note:** Henry eventually succeeded on his second attempt via the npm install path and a destroy+redeploy cycle, but the brew path is not usable on Fedora without manual intervention not covered in docs or the install flow.

**Suggested fix**
Either:
- Publish a Fedora-compatible bottle for the Solo brew formula, or
- Add a pre-install check that detects non-Ubuntu Linux and redirects to the npm install path with a clear message, or
- Document the npm fallback path prominently for non-Ubuntu Linux users (see related docs issue #215).

ddeskov-limechain:
## Description

This pull request makes Solo's Linux install flow distribution-aware instead of assuming `apt-get` + Homebrew on every Linux host. Previously `solo init` / one-shot on Linux hard-coded `apt-get` and bootstrapped linuxbrew to run `brew install podman` — on Fedora (and any non-Debian distro) this failed immediately (`apt-get` absent) and the linuxbrew bootstrap required a C compiler ("Install Clang or run `brew install gcc`"), which is the exact failure reported in #4676.

* Detects the Linux distribution from `/etc/os-release` (`ID` / `ID_LIKE`) and routes to the correct native package manager.
* Adds a shared `LinuxPackageManager` base plus `dnf`, `zypper`, `pacman`, and `apk` managers alongside the existing `apt-get` one; refactors `apt-get` onto the base and makes every install/remove/update/upgrade command non-interactive (e.g. `-y`, `--noconfirm`, `--non-interactive`).
* `OsPackageManager` now resolves the manager per-distro via `/etc/os-release`, with a binary-probe fallback (`FALLBACK_PROBE_ORDER`), and throws a clear `UnsupportedLinuxDistributionSoloError` (`SOLO-5074`) when nothing matches.
* Lifts the sudo-callback setters onto the `PackageManager` interface, removing the `getPackageManager() as AptGetPackageManager` down-cast in `cluster-task-manager.ts`.
* Installs `git`, `iptables`, and `podman` on Linux via the native package manager and **drops the linuxbrew bootstrap on Linux** — removing the C-compiler/Homebrew dependency. macOS (VM-mode podman binary download) is unchanged.
* Removes now-dead `brewPackageManager` / `osPackageManager` injections from `init`.
* Adds unit tests for distribution detection and per-manager command building.

**Behavior change to be aware of (not a contract break):** on Linux the mechanism that installs podman changes from linuxbrew to the native package manager. The end state (podman installed, cluster created) is unchanged, and on the distros tested the native podman is current (see versions in the test table). On very old Ubuntu LTS the apt-provided podman could be older than what linuxbrew shipped — called out under "not tested" below.

### Related Issues

* Closes #4676
* Related to #215

* [ ] This PR updated documentation — no in-repo docs change required; the runtime now emits a clear `UnsupportedLinuxDistributionSoloError` for unsupported distros, and the original `brew install` formula concern lives in `hiero-ledger/homebrew-tools` (tracked by #215)
* [x] This PR has no breaking changes (no CLI flag/output/contract changes; Linux install mechanism change documented above)
* [x] Any technical debt has been documented as a separate issue and linked to this PR — N/A, none introduced
* [x] Any `package.json` changes have been explained to and approved by a repository manager — N/A, no `package.json` changes

### Testing

**1. Unit tests (real detection + command-builder code) — 33 passing, 0 failing**

```bash
npx mocha 'test/unit/core/package-managers/os-package-manager.test.ts' 'test/unit/core/package-managers/linux-package-managers.test.ts'
```

* Detection: `fedora→dnf`, `ubuntu→apt-get`, `debian→apt-get`, `opensuse-leap→zypper`, `arch→pacman`, `alpine→apk`, `rocky` matched via `ID_LIKE`→`dnf`, and an unsupported distro throws `UnsupportedLinuxDistributionSoloError`.
* Command builders: every manager asserts the exact string emitted for install / remove / update / upgrade, and that the availability check (`--version`) runs **without** sudo.
* The existing `init` and `cluster-task-manager` unit tests still pass with the slimmed constructors.

**2. Static checks**

* `npx tsc` — clean (no type errors).
* `npx eslint` on all changed files — 0 errors.
* `npx dpdm … circular:1` — no circular dependencies introduced.

**3. Cross-platform verification on real distributions (Docker)**

For each target distro I ran an ephemeral container (`docker run --rm --pull=missing -v verify.sh:/verify.sh:ro <image> sh /verify.sh`). The in-container script:
1. parses the real `/etc/os-release` (`ID` / `ID_LIKE`),
2. selects the package manager using the same mapping + binary-probe fallback as `OsPackageManager`,
3. confirms `apt-get` is **absent** on non-Debian distros (proving the old "apt-get everywhere" path would have failed there),
4. confirms no C compiler is present or needed,
5. runs the **exact** command Solo emits — e.g. `dnf install -y git iptables podman` — and
6. verifies `git`, `iptables`, and `podman` are installed and prints their versions.

| Distro | Image | Expected → selected manager | `apt-get` absent (old bug) | Install RC | git / iptables / podman | podman version | Result |
|---|---|---|---|---|---|---|---|
| Fedora | `fedora:latest` | dnf → **dnf** | yes | 0 | all present | 5.8.3 | **PASS** |
| Ubuntu | `ubuntu:latest` | apt-get → **apt-get** | n/a (Debian) | 0 | all present | 5.7.0 | **PASS** |
| openSUSE Leap | `opensuse/leap:latest` | zypper → **zypper** | yes | 0 | all present | 5.4.2 | **PASS** |
| Alpine | `alpine:latest` | apk → **apk** | yes | 0 | all present | 5.8.3 | **PASS** |
| Arch | `menci/archlinuxarm:base` (arm64) | pacman → **pacman** | yes | 0 | all present | 5.8.3 | **PASS** |

* All five select the correct native package manager from their real `/etc/os-release`, install `git` + `iptables` + `podman` natively, and report **no C compiler present or required** — the gcc/clang + Homebrew failure from #4676 is gone.
* **Fedora — the exact OS in the bug report — passes cleanly.**
* The Arch image reports `ID=archarm` (Arch Linux ARM), which is not in the os-release map; Solo still correctly selected `pacman` via the stage-2 binary-probe fallback, exercising that path on a real system.
* Arch was first attempted on `archlinux:latest` under x86-64 qemu emulation (Apple-Silicon host) and failed there due to pacman 6.1's internal seccomp sandbox being incompatible with qemu-user emulation — an environment artifact, not a Solo defect. Re-running on an arm64-native image gave a genuine PASS. The arm64 Arch community image needed base prep (`pacman-key --init/--populate`, plus `pacman -Syu` to reconcile a stale `gcc-libs`→`libgcc`/`libstdc++` split); this prep does **not** alter the command Solo emits.

The following was not tested:

* The full rootful `solo init` end-to-end flow (kind cluster creation, sudo elevation, rootful podman) on a real non-Ubuntu Linux host — the Docker matrix validates distro detection and the exact native package-install commands, but not the surrounding Listr task orchestration or the `sudo` wrapper (containers run as root).
* Podman version provided by `apt` on older Ubuntu LTS releases — on the tested `ubuntu:latest` it is current (5.7.0), but older LTS apt repos may ship an older podman than linuxbrew did.
* Windows (WSL2) — unchanged; still routes to `apt-get`.

<details>
<summary>
Commit message guidelines
</summary>

</details>


src/core/dependency-injection/container-init.ts
```diff
@@ -96,6 +96,10 @@ import {BackupRestoreCommandDefinition} from '../../commands/command-definitions
 import {BrewPackageManager} from '../package-managers/brew-package-manager.js';
 import {OsPackageManager} from '../package-managers/os-package-manager.js';
 import {AptGetPackageManager} from '../package-managers/apt-get-package-manager.js';
+import {DnfPackageManager} from '../package-managers/dnf-package-manager.js';
+import {ZypperPackageManager} from '../package-managers/zypper-package-manager.js';
+import {PacmanPackageManager} from '../package-managers/pacman-package-manager.js';
+import {ApkPackageManager} from '../package-managers/apk-package-manager.js';
```
jeromy-cannon: Consider that we could remove these from dependency injection and instead only instantiate the one that is necessary in OsPackageManager.  Anything using OsPackageManager would just be working with the PackageManager type.  Only one of these instances would need to be created making it more efficient.

src/core/cluster-task-manager.ts
```diff
@@ -106,24 +103,11 @@ export class ClusterTaskManager extends ShellRunner {
...
-          if (!brewInstalled) {
-            this.logger.info('Homebrew not found, installing Homebrew...');
-            if (!(await this.brewPackageManager.install())) {
-              throw new SoloErrors.system.homebrewInstallFailed();
-            }
+            const packageManager: PackageManager = this.osPackageManager.getPackageManager();
+            packageManager.setOnSudoGranted(onSudoGranted);
+            packageManager.setOnSudoRequested(onSudoRequested);
+            await packageManager.update();
+            await packageManager.installPackages(['git', 'iptables']);
```
jeromy-cannon, line 115: are the packages named `git` and `iptables` in all of the other distros?

src/core/cluster-task-manager.ts
```diff
@@ -135,9 +119,13 @@ export class ClusterTaskManager extends ShellRunner {
             this.logger.info(`Podman already installed: ${podmanVersion}`);
           } catch {
             this.logger.info('Podman not found, installing Podman...');
-            await this.brewPackageManager.installPackages(['podman']);
```
jeromy-cannon, line 138: There were certain things that the brew installer was handling for us.  It reduced the complexity of the installation.  I don't think we have test cases covering those scenarios.  Please follow up with @Ivo-Yankov to ensure we can remove this.

src/core/cluster-task-manager.ts
```diff
@@ -135,9 +119,13 @@ export class ClusterTaskManager extends ShellRunner {
...
-            const brewBin: string[] = await this.run('which podman');
-            process.env.PATH = `${process.env.PATH}:${brewBin.join('').replace('/podman', '')}`;
+            const {onSudoGranted, onSudoRequested} = this.sudoCallbacks(parentTask);
+            const packageManager: PackageManager = this.osPackageManager.getPackageManager();
+            packageManager.setOnSudoGranted(onSudoGranted);
+            packageManager.setOnSudoRequested(onSudoRequested);
+            await packageManager.installPackages(['podman']);
```
jeromy-cannon: does the podman package exist on all of the other distros?

src/core/cluster-task-manager.ts
ddeskov-limechain: Verified against each distro's repos: `git` is `git` everywhere. `iptables` installs cleanly on all of them (Debian/Ubuntu, openSUSE Leap 15.x/16.0/Tumbleweed, Arch, Alpine, RHEL 7/CentOS 7). The only naming nuance is Fedora/RHEL 8+, where there's no package literally named `iptables` — but `iptables-nft` carries Provides: `iptables`, so `dnf/yum install iptables` resolves to it. Do you want to have an explicit `dnf/ym iptables-nft` on Fedora/RHEL 8+ ?

src/core/cluster-task-manager.ts
ddeskov-limechain: Verified against each distro's actual repo contents: `podman` exists as a package named exactly `podman` on all of them — Debian (5.4.2), Ubuntu 24.04 (4.9.3), Fedora (5.8.3), openSUSE Leap 16.0 (5.4.2), Arch (5.8.3), Alpine (5.8.3, in community), and CentOS/RHEL 7 (1.6.4, in extras). No naming divergence. Only operational notes: Alpine's community repo and RHEL 7's extras channel must be enabled (both default-on for CentOS 7 / standard Alpine), but no code changes needed as I see it.

src/core/cluster-task-manager.ts
ddeskov-limechain: Followed up with @Ivo-Yankov on this.

He confirmed the brew installer was load-bearing specifically for podman: the native package (notably Ubuntu/apt) lags behind, and podman also has rootful/rootless variants where the OS-appropriate version matters. The old Kubic repo that used to backport a current podman to Debian/Ubuntu was discontinued, so brew was the reliable cross-distro way to get a recent-enough podman for kind (kind requires podman ≥ 3.0).

So I've restored the Homebrew bootstrap and brew-based podman install for the rootful Linux path — it's  in commit 503a4f60.

On the test-coverage point — you're right that we don't exercise the actual per-distro install. That's a pre-existing gap (we only have package-selection unit tests, nothing that runs a real install podman). Do we have an issue to track per-distro install validation?

src/core/cluster-task-manager.ts
jeromy-cannon: yes, that sounds appropriate

src/core/cluster-task-manager.ts
jeromy-cannon: No, would you create one?  Just one is enough and make it an epic.  It probably will be a low priority and sit in the backlog for a while.  

My Mac is a M1.  I ran into issues with having a Linux VM and nested virtualization.  It seems it is supported with M3+.  https://apple.stackexchange.com/questions/466761/is-nested-virtualization-is-supported-by-apple-silicon-chips-m2-m3

src/core/dependency-injection/container-init.ts
ddeskov-limechain: Done — the per-distro package managers aren't registered anymore

src/core/cluster-task-manager.ts
ddeskov-limechain: Created the issue to track per-distro install validation: https://github.com/hiero-ledger/solo/issues/4888

jeromy-cannon (approved):

Issue #4005, opened 2026-04-14
security: convert shell exec calls to array form with shell:false



jeromy-cannon:
## Overview
All shell execution calls should be converted to use array arguments with `shell: false` to eliminate shell injection vulnerabilities. Where environment variables are needed, build a minimal set of required env vars and pass them explicitly rather than relying on shell interpolation. Cases requiring more complex logic changes will be tracked in separate linked issues. 

## Sub-Tasks
- [ ] Audit all shell exec / child_process calls across the codebase
- [ ] Convert each call to use array form (e.g. `execFile` or `spawn` with array args)
- [ ] Set `shell: false` on each converted call
- [ ] Identify minimal required environment variables for each call and pass them explicitly
- [ ] For calls requiring significant logic changes, open a linked follow-up issue
- [ ] Verify no regressions in unit and E2E tests

## Prerequisites
- [ ] #4004 (related shell security hardening) — coordinate to avoid conflicts

## References
- https://docs.google.com/spreadsheets/d/1d6fssx4wpSfiph00DQJN644dylXZqFsCOXTsBJRZLfo — SA4, SA5, SA6, SA22
- relates to #4004

jeromy-cannon:
@jan-milenkov , what is the status on this?

jan-milenkov:
@jeromy-cannon I've set `shell: false` for almost all calls, except for some `kubectl` calls that were causing too much trouble.

Helm values got a rework, so all the Helm calls are complete. 

`Identify minimal required environment variables for each call and pass them explicitly` is partially done, as it needed too much time to investigate what all the envs are.

The rest is dependent on the `Prerequisites` issue.

jeromy-cannon:
@jan-milenkov , would you create some sub-issues for the items remaining? then we can continue finishing this in Q3.

jan-milenkov:
> [@jan-milenkov](https://github.com/jan-milenkov) , would you create some sub-issues for the items remaining? then we can continue finishing this in Q3.

Created and linked

jeromy-cannon:
@jan-milenkov , can this be closed?

PR #4804, opened 2026-06-24
chore: convert shell exec calls to array form with shell false

jan-milenkov:
## Description
                                                                                                                                                                                
Changed how external commands runs
  
### Related Issues

* Related to # https://github.com/hiero-ledger/solo/issues/4005


test/e2e/integration/core/dependency-managers/podman-dependency-manager.test.ts
```diff
@@ -427,7 +427,7 @@ describe('PodmanDependencyManager', (): void => {
         }
         throw Object.assign(new Error('ENOENT'), {code: 'ENOENT'});
       });
-      runStub.withArgs(`"${fakeGlobalPodmanPath}" --version`).resolves([`podman version ${PODMAN_LOW_VERSION}`]);
+      runStub.withArgs(fakeGlobalPodmanPath, ['--version']).resolves([`podman version ${PODMAN_LOW_VERSION}`]);
       runStub
         .withArgs(`"${PathEx.join(temporaryDirectory, 'podman')}" --version`)
```
jeromy-cannon: This stub was not updated to match the new `run(executableWithPath, ['--version'])` call signature. The sibling stub at line 430 (`fakeGlobalPodmanPath`) was updated; this one for the local path was missed.

```suggestion
      runStub.withArgs(PathEx.join(temporaryDirectory, 'podman'), ['--version']).resolves([`podman version ${PODMAN_LOW_VERSION}`]);
```

test/e2e/integration/core/dependency-managers/crane-dependency-manager.test.ts
```diff
@@ -363,8 +363,8 @@ describe('CraneDependencyManager', (): void => {
 
     it('should install crane locally if the global installation does not meet the requirements', async (): Promise<void> => {
       runStub.withArgs('which crane').resolves(['/usr/local/bin/crane']);
```
jeromy-cannon: `'which crane'` is the old string-form signature. If `BaseDependencyManager` still routes global executable detection through `this.run()`, update to `('which', ['crane'])`. If it no longer does (per the comment at base-dependency-manager.ts line 113 — "This avoids spawning a shell subprocess"), this stub is dead code and should be removed.

src/core/package-managers/brew-package-manager.ts
```diff
@@ -1,48 +1,88 @@
 // SPDX-License-Identifier: Apache-2.0
 
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
```
jeromy-cannon: New `node:path` import — per TS style guide §3.3.5, new `node:path` imports should use `PathEx` from `src/business/utils/path-ex.ts` instead. Affects `path.join()` in `runHomebrewScript` and `path.delimiter` in `install()` / `applyShellEnvironment()`.

```suggestion
import {PathEx} from '../../business/utils/path-ex.js';
```

(Remove the `import path` line; `os.tmpdir()` can stay as-is since `PathEx` wraps path ops, not temp-dir queries.)

src/core/package-managers/brew-package-manager.ts
```diff
@@ -1,48 +1,88 @@
...
+   * Downloads a Homebrew install/uninstall script and runs it with bash, without invoking a shell to
+   * perform command substitution (replaces the previous `bash -c "$(curl …)"` pattern).
+   */
+  private async runHomebrewScript(scriptUrl: string): Promise<void> {
+    const scriptPath: string = PathEx.join(fs.mkdtempSync(PathEx.join(os.tmpdir(), 'solo-brew-')), 'homebrew.sh');
+    try {
+      await this.run('curl', ['-fsSL', scriptUrl, '-o', scriptPath]);
+      await this.run('bash', [scriptPath], true, false, {NONINTERACTIVE: '1'});
+    } finally {
+      fs.rmSync(scriptPath, {force: true});
```
jeromy-cannon: `mkdtempSync` creates a temp directory, but `rmSync` here only removes the script file — the parent directory is leaked on every call. Consider:

```suggestion
      fs.rmSync(path.dirname(scriptPath), {recursive: true, force: true});
```
(import `path` from `node:path`, or use `PathEx` and adjust to grab the parent dir)

src/commands/backup-restore.ts
```diff
@@ -322,9 +322,20 @@ export class BackupRestoreCommand extends BaseCommand {
...
-            await shellRunner.run(compressionCommand, [], true, false);
+            // Run zip from the output directory (cwd) with an explicit argument array and no shell, so the
+            // password and file names cannot be interpreted by a shell.
+            await shellRunner.run(
+              'zip',
+              ['-rX', '-P', zipPassword, zipFile, '.'],
+              true,
+              false,
+              {},
+              undefined,
```
jeromy-cannon: two explicit `undefined` values just to reach `workingDirectory` as the 9th positional parameter — this is the fragility that comes from adding another positional param to an already-8-deep signature. Would it make sense to make `workingDirectory` part of an options object (similar to Node's `SpawnOptions`) so callers do not have to thread through positional `undefined`s?

src/commands/backup-restore.ts
```diff
@@ -1397,9 +1408,24 @@ export class BackupRestoreCommand extends BaseCommand {
           this.logger.info(`Multiple clusters detected (${context_.clusters.length}), creating Kind Docker network...`);
           try {
             const shellRunner: ShellRunner = new ShellRunner(this.logger);
-            await shellRunner.run(
-              'docker network rm -f kind || true && docker network create kind --scope local --subnet 172.19.0.0/16 --driver bridge',
-            );
+            // Remove any pre-existing network (ignoring failure, replacing the previous `|| true`), then
```
jeromy-cannon: stale comment — "replacing the previous `|| true`" and "instead of a shell `||`/`&&` pipeline" reference the old code. Per checklist §10, keep only the load-bearing WHY (that failure here is safe to ignore).

```suggestion
            // Ignore failure — the network may not exist yet.
```
(remove the second line entirely)

src/core/cluster-task-manager.ts
```diff
@@ -145,47 +145,69 @@ export class ClusterTaskManager extends ShellRunner {
...
-                `${this.podmanInstallationDirectory}${path.delimiter}` +
-                `${this.kindInstallationDirectory}${path.delimiter}${process.env.PATH}`,
-            },
-          ];
+          const sudoEnvironment: Record<string, string> = {
+            PATH:
+              `${this.podmanInstallationDirectory}${path.delimiter}` +
+              `${this.kindInstallationDirectory}${path.delimiter}${process.env.PATH}`,
+          };
+          // PATH passed to kind (via `env`) replicates the previous inline `PATH="$PATH:${podmanPath}"`,
```
jeromy-cannon: stale comment — "replicates the previous inline PATH=..." references old code that won't mean anything to a future reader. The second sentence is the load-bearing part; drop the first.

```suggestion
          // PATH must include both kindInstallationDirectory (for kind) and podmanPath (for podman).
```

src/core/cluster-task-manager.ts
```diff
@@ -145,47 +145,69 @@ export class ClusterTaskManager extends ShellRunner {
...
-            `chmod 755 ${temporaryDirectory}/kube-config-root`,
-            ...sudoRunOptions,
+            'chmod',
+            ['755', rootKubeConfigPath],
+            false,
+            false,
+            sudoEnvironment,
           );
 
           const rootYamlData: string = fs.readFileSync(`${temporaryDirectory}/kube-config-root`, 'utf8');
```
jeromy-cannon: duplicate literal — `rootKubeConfigPath` was introduced a few lines above for exactly this string; use it here too.

```suggestion
          const rootYamlData: string = fs.readFileSync(rootKubeConfigPath, 'utf8');
```

src/core/package-managers/brew-package-manager.ts
```diff
@@ -1,48 +1,88 @@
...
+    }
+  }
+
+  /**
+   * Applies the environment that `brew shellenv` would export, without a shell `eval`. Parses the
+   * `export KEY="VALUE";` lines and expands references to the current PATH so the values are correct.
+   */
+  private async applyShellEnvironment(): Promise<void> {
+    const output: string[] = await this.run(`${BrewPackageManager.LINUXBREW_BIN}/brew`, ['shellenv']);
```
jeromy-cannon, line 76: does `brew shellenv` ever emit other shell variable references beyond `${PATH+:$PATH}` and `$PATH` (e.g. `$HOMEBREW_PREFIX` in MANPATH or INFOPATH values)? If so, those would pass through unexpanded. Is this tested against actual `brew shellenv` output on a Linux machine?

src/integration/kind/execution/kind-execution.ts
```diff
@@ -100,6 +116,11 @@ export class KindExecution {
     });
 
     const successPromise: Promise<boolean> = new Promise((resolve): void => {
+      if (this.spawnError) {
+        resolve(false);
+        return;
+      }
+      this.process.on('error', (): void => resolve(false));
```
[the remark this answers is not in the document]
jan-milenkov: fkxed

src/core/package-managers/brew-package-manager.ts
```diff
@@ -1,48 +1,88 @@
...
+   * perform command substitution (replaces the previous `bash -c "$(curl …)"` pattern).
+   */
+  private async runHomebrewScript(scriptUrl: string): Promise<void> {
+    const scriptPath: string = PathEx.join(fs.mkdtempSync(PathEx.join(os.tmpdir(), 'solo-brew-')), 'homebrew.sh');
+    try {
+      await this.run('curl', ['-fsSL', scriptUrl, '-o', scriptPath]);
+      await this.run('bash', [scriptPath], true, false, {NONINTERACTIVE: '1'});
+    } finally {
+      fs.rmSync(scriptPath, {force: true});
+    }
```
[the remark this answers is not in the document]
jan-milenkov: fixed

src/core/package-managers/brew-package-manager.ts
jan-milenkov: updated

test/e2e/integration/core/dependency-managers/crane-dependency-manager.test.ts
```diff
@@ -439,7 +438,7 @@ describe('CraneDependencyManager', (): void => {
             return [executablePath];
           });
 
-        sandbox.stub(ShellRunner.prototype, 'run').withArgs(`which ${constants.CRANE}`).alwaysReturned(false);
+        sandbox.stub(ShellRunner.prototype, 'run');
```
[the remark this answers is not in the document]
jan-milenkov: updated

src/commands/backup-restore.ts
jan-milenkov: updated

src/core/cluster-task-manager.ts
jan-milenkov: updated

src/core/package-managers/brew-package-manager.ts
jan-milenkov: i've tested it locally and updated it, i've not tested on linux

jan-milenkov: brew shellenv emits more than PATH (the HOMEBREW_* vars, plus INFOPATH with ${INFOPATH:-} and MANPATH as a conditional line). I reworked the parser to expand the general forms (${VAR+…}, ${VAR:-…}, $VAR) instead of hardcoding PATH. The MANPATH conditional just gets skipped, which is fine since we don't use i

jeromy-cannon (approved):

PR #4879, opened 2026-06-26
ci: add share solo image cache in github workflow

### LINKED ISSUE #4803: Use shared GitHub cache to warm Solo image cache in CI
## Problem

Many CI workflows repeatedly run `solo cache image pull`, downloading the same Solo dependency images from Hashgraph, GHCR, Docker Hub, Quay, and related registries. This adds latency to jobs and increases exposure to registry throttling or transient registry failures.

## Proposal

Add a shared GitHub Actions cache layer that stores Solo image cache artifacts and uses it as a buffer before falling back to registry pulls. Workflows should restore the cache early, run `solo cache image pull` only for cache misses or changed versions, then save/update the cache for later jobs.

## Goals

- Reduce elapsed time spent in `solo cache image pull` across CI workflows.
- Reduce repeated pulls from external registries for identical Solo/CN/MN/BN/relay/explorer versions.
- Keep cache keys version-aware so stale images are not reused across incompatible version changes.
- Provide a reusable workflow action or shared workflow snippet so all relevant workflows use the same cache behavior.

## Acceptance Criteria

- A shared cache key strategy covers Solo version, consensus node version, mirror node version, block node version, relay version, explorer version, and any registry mirror inputs that affect image selection.
- CI workflows restore the shared image cache before `solo cache image pull`.
- `solo cache image pull` can reuse restored artifacts and only pull missing images.
- Workflows save the warmed cache after successful image pull/cache population.
- Migration, one-shot, performance, examples, and nightly workflows use the common cache path instead of duplicating logic.
- Cache fallback remains safe: if GitHub cache is missing, stale, or corrupt, workflows still pull from registries and continue normally.

## Notes

This should be tracked separately from migration test fixes. The implementation can build on the `ci-share-solo-image-cache` branch/PR work if that branch is still available.

JeffreyDallas:
## Description

* Added a shared GitHub Actions cache layer that stores Solo image cache artifacts and uses it as a buffer before falling back to registry pulls, to lower exposure to registry throttling or transient registry failures.
* Restored cache image env var for standard runner and windows runner.
* **Fix:** explicitly set `blockStream.writerMode=FILE` when no block node is deployed to prevent the CN from stalling record file production (see Root Cause Analysis below).
* **Fix:** corrected three stale/wrong image entries in `solo-cache-images-target.yaml` that were causing `ImagePullBackOff` on subsequent CI runs (see Image Cache Fix below).
* **Enhancement:** `solo cache load` now throws an error if any expected image is missing from the cluster after the load completes, catching future yaml drift early.
* **Fix (follow-up):** `blockStream.writerMode` was incorrectly included in `soloManagedKeys` in the one-shot orchestrator, causing it to be stripped from the merged `application.properties` before staging. The CN pod therefore started without an explicit `writerMode`, defaulting to `FILE_AND_GRPC` and stalling. Removed from `soloManagedKeys` so the base template's `writerMode=FILE` passes through directly (see Windows Stall Follow-up below).

### Root Cause Analysis — Windows runner pinger failure

**Symptom:** pinger pod exits with code 1 after ~30 minutes on the Windows runner one-shot workflow. The test had been passing before image caching was added.

**Root cause:** CN defaults `blockStream.writerMode=FILE_AND_GRPC` (present since v0.72.0). Without a block node to drain the gRPC side, the 5-block buffer (`blockStream.buffer.maxBlocks=5` in the small-memory profile) fills up during account-setup transactions and **permanently stalls record file production**. The mirror importer never receives new `.rcd.gz` files, pinger transactions never appear in the mirror REST API, and the pinger readiness probe eventually fails.

This is deterministic — not a race condition. The buffer always fills after ~5 blocks, which account setup consumes before the pinger even starts.

**Why image caching made it surface consistently:** without caching, image pulls took 30–60 s on WSL2, creating enough timing slack to obscure the stall. With caching all pods start immediately, account setup transactions hit the CN at full speed, and the buffer fills every run.

**Fix — three files:**

1. `resources/templates/application.properties` — adds `blockStream.writerMode=FILE` as the explicit default so non-one-shot deploys are also covered.

2. `src/core/profile-manager.ts` — removes the early return in `updateApplicationPropertiesForBlockNode` that skipped the no-block-node case. Now explicitly writes `writerMode=FILE` when no block nodes are deployed (handles one-shot, where the template value is stripped during staging).

3. `src/core/helpers.ts` — changes `createAndCopyBlockNodeJsonFileForConsensusNode` to **replace** an existing `writerMode=FILE` entry with `FILE_AND_GRPC` (instead of only appending), so when a block node is present the value is always upgraded correctly regardless of deploy order.

**Version compatibility:** `blockStream.writerMode` is a valid CN config property with the same `FILE_AND_GRPC` default from v0.72.0 onward. No version gating is needed. When no block node is deployed, `FILE` is safe on all supported CN versions. When a block node is present, `createAndCopyBlockNodeJsonFileForConsensusNode` upgrades it to `FILE_AND_GRPC`.

### Windows Stall Follow-up — writerMode stripped by soloManagedKeys

**Symptom:** After the `blockStream.writerMode=FILE` fix in the base template and profile-manager, the Windows runner one-shot still stalled with the same 5-block buffer fill pattern. Inspection of the actual CN pod `application.properties` from CI confirmed `blockStream.writerMode` was **absent** from the file — despite the fix in `resources/templates/application.properties`.

**Root cause:** `DefaultOneShotDeployOrchestrator.deploy()` builds a merged `application.properties` by concatenating the base template with the small-memory overlay, then filters out a `soloManagedKeys` set of properties (realm, shard, chainId, streamMode) before passing the result to `network deploy` as `--application-properties`. `blockStream.writerMode` was mistakenly included in `soloManagedKeys`, so it was stripped from the merged file. The intent was that `prepareStagingDirectory`'s `updateApplicationPropertiesForBlockNode` call would re-add it, but the re-add evidently does not survive to the final ConfigMap in the Windows path, leaving the CN without an explicit `writerMode` and defaulting to `FILE_AND_GRPC`.

**Evidence:** the 1405-byte `application.properties` extracted from the failing CN pod (jobs 83835673825/83836526440, commit `5bf31fde`) contained `blockStream.streamMode=RECORDS` (correctly re-added by staging) but was completely missing `blockStream.writerMode` and the surrounding comment block — exactly matching the shape of a merged file that had `writerMode=FILE` filtered out.

**Fix:** removed `'blockStream.writerMode'` from `soloManagedKeys` in `default-one-shot-deploy-orchestrator.ts`. The base template already carries `writerMode=FILE`; keeping it in the merged file means it passes through to staging directly without depending on the re-add step. `updateApplicationPropertiesForBlockNode` still overrides it to `FILE_AND_GRPC` for block-node deployments as before.

### Image Cache Fix — standard runner `ImagePullBackOff`

**Symptom:** after the `blockStream.writerMode` fix, standard runner CI still failed with `ImagePullBackOff` on `mirror-1-web3` and `network-node1-0`.

**Root cause:** three entries in `solo-cache-images-target.yaml` were wrong or stale:

| Entry | Problem | Correct value |
|---|---|---|
| `gcr.io/mirrornode/hedera-mirror-web3` | Wrong registry — `mirror-node-values.yaml` overrides web3 to `ghcr.io/hiero-ledger/hiero-mirror-node/web3` | `ghcr.io/hiero-ledger/hiero-mirror-node/web3` |
| `ghcr.io/hashgraph/solo-containers/ubi8-init-java21:0.41.2` | Wrong image name and hardcoded stale version — solo-deployment chart uses `debian-s6-java25` | `ghcr.io/hashgraph/solo-containers/debian-s6-java25:SOLO_CONTAINERS_VERSION` |
| `ghcr.io/hashgraph/solo-cheetah/cheetah:0.3.4` | Hardcoded stale version — chart uses `0.4.5` | `ghcr.io/hashgraph/solo-cheetah/cheetah:SOLO_CHEETAH_VERSION` |

**Why tests passed most of the time:** a wrong entry in the cache yaml means zero cache benefit for that image — kubelet still pulls from the live registry, which usually succeeds. Failures only appeared on transient registry errors (EOF), which are common on shared runners under load.

**Fix:**

* Corrected all three entries in `solo-cache-images-target.yaml`.
* Added `SOLO_CHEETAH_VERSION` and `SOLO_CONTAINERS_VERSION` template variables (backed by `version.ts` constants) so these versions stay in sync with `charts/solo-deployment/values.yaml` whenever `SOLO_CHART_VERSION` bumps.
* Added a `verifyImagesLoadedIntoCluster` task that compares expected images against `kind get nodes` loaded images after the cache load completes; throws a descriptive `SoloError` listing missing images if any are not found.
* **Fix (follow-up):** the verification task used exact-string comparison but containerd's `ctr images ls -q` omits the `docker.io/` registry prefix for Docker Hub images (e.g. returns `library/busybox:latest`, not `docker.io/library/busybox:latest`). This caused all 11 Docker Hub images to appear missing even though they were loaded correctly. Fixed by normalizing both sides — stripping `docker.io/` before the set comparison.

**Note on GHCR vs GCR for mirror node:** `mirror-node-values.yaml` deliberately overrides `grpc`, `importer`, `rest-java`, and `web3` to `ghcr.io/hiero-ledger/hiero-mirror-node/*` (new Hiero branding). The chart defaults use `gcr.io/mirrornode/hedera-mirror-*` (old Hedera branding). GCR images for those 4 components do exist at the current version, but the overrides are intentional — the project is migrating to GHCR and GCR will eventually be deprecated. The cache yaml now correctly reflects the GHCR registry for those 4 components and GCR for `rest`, `pinger`, and `monitor` (which have no overrides).

### Related Issues

* Closes #4803
* Related to #

### Testing

* Diagnosed failing Windows runner run ([job 83750594142](https://github.com/hiero-ledger/solo/actions/runs/28259165845/job/83750594142)): confirmed last record file at `01:12:48`, mirror importer stuck retrying from that timestamp for 30 minutes, block-node-comms.log shows no block node connection, application.properties has no explicit `writerMode`.
* Confirmed `blockStream.writerMode` property and `FILE_AND_GRPC` default exist in CN v0.72.0, v0.73.0, and v0.74.0 source.
* Diagnosed standard runner `ImagePullBackOff` ([job 83819001096](https://github.com/hiero-ledger/solo/actions/runs/28289502544/job/83819001096)): traced `mirror-1-web3` to wrong GCR entry in cache yaml and `network-node1-0` to stale `ubi8-init-java21` image name.
* Verified GCR images for all mirror node components exist at current version with both AMD64 and ARM64 manifests.
* Verified correct GHCR image names by inspecting the live mirror node Helm chart subchart values.
* Diagnosed follow-up failure ([runs 28290875191](https://github.com/hiero-ledger/solo/actions/runs/28290875191) / [28290875183](https://github.com/hiero-ledger/solo/actions/runs/28290875183)): verification task false-negatives for all 11 Docker Hub images due to missing `docker.io/` prefix normalization.
* Extracted CN pod `application.properties` from failing Windows CI run (jobs 83835673825/83836526440, commit `5bf31fde`): confirmed `blockStream.writerMode` entirely absent despite base template fix, identifying `soloManagedKeys` as the root cause.

The following was not tested:

* End-to-end CI run with the `soloManagedKeys` fix applied (pending CI).

<details>
<summary>
Commit message guidelines
</summary>

</details>


JeffreyDallas:
Split to smaller PR #4886

PR #4894, opened 2026-06-29
feat: complete new idempotency strategy for one-shot and strengthen the one-shot destroy command

### LINKED ISSUE #4812: feat: enhance destroy command to properly handle local and remote config
## Problem

  When `solo one-shot single deploy` fails mid-flight, local config (`~/.solo/local-config.yaml`) and the remote ConfigMap (`solo-remote-config`) are left behind. Phase 1 of the
  idempotency epic (#4423) made the **re-deploy** path recover from this state gracefully — guards detect the stale entries and skip past them.

  But the **destroy** path doesn't have the same treatment. Operators who instead try to clean up via `solo one-shot single destroy` (per the error message's own suggestion: `→ If
  rollback was skipped, clean up partial resources: solo one-shot single destroy`) hit cases where destroy itself fails, leaving the configs stuck in place. The only remaining
  recovery is the manual `kind delete cluster && rm -rf ~/.solo` that #4348, #3965, and #3968 all describe.
  
  Destroy needs to be a true symmetric counterpart of deploy: same idempotency guarantees, same best-effort recovery, same "always converges to clean state" promise.

  ## Concrete failure modes destroy should handle
  
  1. **Local config has deployment, remote ConfigMap is missing.** Destroy should remove the local entry and exit cleanly. Currently throws because it expects the ConfigMap to
  exist.
  2. **Local config has deployment, remote ConfigMap exists, Helm releases partially installed.** Destroy should uninstall what's there (skip-if-missing per release), delete the
  ConfigMap, delete the namespace, then clean local config.
  3. **Kind cluster is gone but local config still references it.** Destroy should detect the unreachable cluster context, skip the cluster-side cleanup with a `WARN`, and still
  purge local config.
  4. **Cluster-ref entries left behind after deployment is removed.** Destroy should drop cluster-refs that no remaining deployment uses.
  5. **`solo deployment delete` invoked on a deployment that was never fully created.** Should be idempotent — succeeds if there's nothing to delete.

  ## Expected behavior

  `solo one-shot single destroy` (and the underlying `solo deployment delete`) becomes:

  - **Idempotent** — running it on already-clean state is a no-op, not an error.
  - **Best-effort cleanup, never blocking** — each sub-step either succeeds or logs a `WARN` and continues. The final local-config cleanup always runs.
  - **Symmetric with deploy** — emits `Step 'X' skipped: ...` log lines (same pattern as Phase 1's deploy guards) when a sub-step finds nothing to do.
  - **Converges to clean state** — after destroy, `~/.solo/local-config.yaml` has no `deployments` or `clusterRefs` entries for the destroyed deployment, and the remote ConfigMap
  is gone (or unreachable but recorded as such).

  ## Acceptance criteria

  - Destroy succeeds on fully-deployed state (no regression).
  - Destroy succeeds when local config has the deployment but the remote ConfigMap is missing.
  - Destroy succeeds when the kind cluster is unreachable; local config is still cleaned.
  - Destroy succeeds when invoked on already-clean state (idempotent no-op).
  - After destroy, `~/.solo/local-config.yaml` contains no stale `deployments` or unused `clusterRefs` entries for the destroyed deployment.
  - New E2E test covering the failed-deploy → destroy → re-deploy cycle without manual cleanup.

  ## Related work

  - Phase 1 idempotency epic: #4423
  - #3965, #3968 — deploy-side fixes for the same stale-state problem (closed)
  - #4348 — explicit user report of the recovery gap
  - #2053 — broader self-healing for `~/.solo`

### LINKED ISSUE #4813: feat(one-shot): auto-clean stale resources at the start of deploy when none can be verified healthy
## Problem

  `solo one-shot single deploy` already runs an early "Check existing deployment state" step (Phase 1 of the idempotency epic, #4423) that pre-fetches a `DeploymentStateSnapshot`
  to decide whether each subsequent orchestration step is "already done" and can be skipped.

  The snapshot only checks **existence**, not **health.** A Helm release showing as installed counts as "done" regardless of whether its pods are running, healthy, or crashing. The
   Phase 1 RFC §3 explicitly defers health-aware skipping to Phase 3:

  > Detecting unhealthy-but-installed resources (e.g., pods in CrashLoopBackOff will be treated as installed → skipped). Deferred to Phase 3.

  Until Phase 3 lands, the snapshot can't tell a usable prior deployment from a broken one. The safer behavior in the meantime: when the snapshot finds **any** existing one-shot
  resources, auto-destroy them at the start of deploy and proceed with a fresh install. Skipping past unhealthy state is worse than starting over — operators get stuck with deploys
   that "succeed" against half-broken state.
  
  ## Expected behavior

  A new pre-deploy step (or extension of the existing "Check existing deployment state") that:

  1. **Reads the snapshot** (already produced today).
  2. **If any one-shot resource is detected** — remote ConfigMap, any component Helm release, any per-component phase ≥ DEPLOYED, accounts.json on disk, etc. — invokes destroy
  logic to remove it.
  3. **Logs an explicit banner** so the operator sees what was found and why we're cleaning it up: `Existing one-shot deployment state detected; cleaning up before proceeding`.
  4. **Proceeds with a fresh deploy** after cleanup completes.

  This is a temporary safety net for the period between Phase 1 (already shipped) and Phase 3 (health-aware guards, not yet designed). Once health-aware guards land, deploy can
  intelligently skip healthy prior state and only auto-clean the unhealthy parts. For now, "clean everything we find" is the correct default.

  ## Behavior matrix

  | Snapshot finds | Action |
  |---|---|
  | Nothing | Fresh deploy as today (no-op for this feature) |
  | Any one-shot resource | Auto-clean, then fresh deploy |

  ## Why not just use Phase 1's skip-guards?

  Because they're existence-only. Without health, "skip because it's there" sometimes means "skip because it's broken" — and the deploy then fails downstream on a half-working
  consensus node or a stuck Helm release. The auto-clean approach removes that ambiguity until Phase 3 lands.

  ## Why not just rely on the user running destroy manually?

  Two reasons:

  1. The reported user friction in #4348 and similar issues is precisely that operators don't realize they need to run destroy first, or that destroy itself sometimes can't recover
   from partial state (a sibling destroy-robustness issue tracks that side).
  2. The whole point of one-shot is "one command and it works." Forcing operators to run destroy + deploy is two commands.

  ## Related work
  
  - Phase 1 idempotency epic: #4423
  - Phase 1 RFC §3 (Non-Goals — explicit deferral of health-aware skipping): `docs/design/architecture/idempotent-one-shot.md`
  - Destroy-handles-stale-config sibling issue (the destroy-side counterpart)
  - #4348, #3965, #3968 — user-visible reports of partial-state recovery friction
  - Phase 3 design (when written): this feature can be removed or scoped down once health-aware guards exist

jan-milenkov:
## Description

Make `one-shot single deploy`/`destroy` always converge to a clean state.
- **Destroy is now best-effort and never blocking.** It cleans up local config even when the cluster is unreachable or the remote config/ConfigMap is missing, and prunes unused cluster-refs. `deployment delete` warns instead of      
  throwing on leftover remote resources.
- **Deploy auto-cleans before running.** If any prior one-shot state is detected, deploy destroys it and rebuilds from a clean slate instead of trying to resume.
- **Cleanup always asks first.** A confirmation prompt is shown before destroying prior installs. `--quiet`/`--force` can't show a prompt, so they error (new `ConfirmationRequiredSoloError`, `SOLO-4077`) rather than silently wiping a deployment.
- **Removed the "skip already-deployed components" logic.** Since deploy always rebuilds, the per-step idempotency guards (and the snapshot fields/helpers behind them) are gone.

### Related Issues

* Closes # https://github.com/hiero-ledger/solo/issues/4812
* Closes # https://github.com/hiero-ledger/solo/issues/4813


test/unit/commands/one-shot/orchestrator/deploy/default-one-shot-deploy-orchestrator.test.ts
```diff
@@ -387,554 +354,21 @@ describe('DefaultOneShotDeployOrchestrator buildDeploymentStateSnapshot', (): vo
...
+function createAccountsSkip(config: OneShotSingleDeployConfigClass): boolean {
   const orchestrator: DefaultOneShotDeployOrchestrator = makeOrchestrator();
   // @ts-expect-error - to access private method
-  const task: {skip: (context_: MockType) => boolean} = orchestrator.buildCreateAccountsTask(config);
-  return task.skip({deploymentStateSnapshot: snapshot} as MockType);
+  const task: {skip: () => boolean} = orchestrator.buildCreateAccountsTask(config);
+  return task.skip();
 }
 
 describe('DefaultOneShotDeployOrchestrator Create Accounts skip guard', (): void => {
```
jeromy-cannon, line 376: We should also be looking for opportunities for catching as much as possible inside of unit testing which can run far faster, more frequently, and with less costs.

The 573-line deletion correctly removes tests for deleted code, but the replacement logic is uncovered:
- `hasExistingOneShotState` (the core detection gate) — no tests
- `buildAutoCleanConfirmationMessage` — no tests
- `quiet`/`force` → `ConfirmationRequiredSoloError` path — no tests
- Prompt confirm/deny flow in the "Confirm cleanup" phase — no tests

Please add a `describe('hasExistingOneShotState', ...)` block here covering at minimum: (a) returns false when snapshot is undefined, (b) returns true when ConfigMap exists, (c) returns true when Helm releases present, (d) returns true when accounts file present, (e) returns true when a component phase is at DEPLOYED, (f) returns false when all fields are empty.

src/commands/one-shot/orchestrator/destroy/default-one-shot-destroy-orchestrator.ts
```diff
@@ -332,43 +338,33 @@ export class DefaultOneShotDestroyOrchestrator implements OneShotDestroyOrchestr
...
-                (error as {cause?: unknown}).cause instanceof NoKubeConfigContextError
-              ) {
-                this.logger.showUser(
-                  'Remote config not found. This may indicate that the deployment has already been deleted or there is an issue with the cluster. Proceeding with best effort cleanup.',
+                this.logger.warn(
+                  `Cluster context '${config.context}' is unreachable; skipping cluster-side teardown and cleaning local config only.`,
                 );
-                this.logger.error('Error loading remote config:', error);
-                config.skipAll = true;
+                config.skipClusterCleanup = true;
```
jeromy-cannon, line 349: `skipClusterCleanup` is declared as `boolean` (non-optional) in `OneShotSingleDestroyConfigClass`, but it is only set to `true` in catch blocks — on the happy path it stays `undefined`. Works at runtime since `undefined` is falsy, but violates the declared type. Initialize it to `false` before this block:

```suggestion
                config.skipClusterCleanup = true;
```

Consider adding `config.skipClusterCleanup = false;` at the start of the Initialize task before the cluster-connection try/catch, so the two-state semantics are explicit.

src/commands/one-shot/orchestrator/deploy/default-one-shot-deploy-orchestrator.ts
```diff
@@ -1309,6 +1191,25 @@ export class DefaultOneShotDeployOrchestrator implements OneShotDeployOrchestrat
...
+  private buildAutoCleanConfirmationMessage(snapshot: DeploymentStateSnapshot | undefined): string {
+    const detected: string[] = [];
+    if (snapshot?.remoteConfig.configMapExists) {
+      detected.push('  - remote config (ConfigMap)');
+    }
+    if (snapshot && snapshot.helm.installedReleases.size > 0) {
+      detected.push(`  - Helm releases: ${[...snapshot.helm.installedReleases].join(', ')}`);
+    }
+    if (snapshot?.accounts.accountsFileExists) {
+      detected.push('  - accounts file on disk');
```
jeromy-cannon, line 1197: `hasExistingOneShotState` can return `true` purely because of component phases (no ConfigMap, no Helm releases, no accounts file), but `buildAutoCleanConfirmationMessage` does not include phases in the `detected` list. The confirmation dialog would show the header with a blank bullet list, which is confusing.

Consider appending the detected component phases after line 1203:
```suggestion
      detected.push('  - accounts file on disk');
    }
    for (const [type, phase] of (snapshot?.remoteConfig.componentPhases ?? new Map()).entries()) {
      if (isDeploymentPhaseAtLeast(phase, DeploymentPhase.DEPLOYED)) {
        detected.push(`  - component ${type} in phase ${phase}`);
      }
    }
```

jan-milenkov: updated

src/commands/one-shot/orchestrator/destroy/default-one-shot-destroy-orchestrator.ts
jan-milenkov: updated

test/unit/commands/one-shot/orchestrator/deploy/default-one-shot-deploy-orchestrator.test.ts
jan-milenkov: updated

jeromy-cannon (approved):

Issue #4888, opened 2026-06-29
Automated per-distro install validation for Solo system dependencies


ddeskov-limechain:
## Summary
Solo installs system dependencies (git, iptables, podman) on Linux via per-distro
native package managers (apt-get, dnf, yum, zypper, pacman, apk) plus a Homebrew
bootstrap used to obtain a recent-enough podman for kind. Today the only automated
coverage is unit tests for package-manager *selection* — nothing exercises a real
install on each distribution, so per-distro regressions (package naming, repo
availability, podman version) would go undetected.

## Goal
Validate actual installs on each supported distribution (containers/VMs) and assert
git, iptables, and podman end up installed and usable.

## Scope / coverage
- Distros: Debian/Ubuntu (apt-get), Fedora/RHEL 8+/Rocky/Alma (dnf), RHEL 7/CentOS 7
  (yum), openSUSE/SLES (zypper), Arch (pacman), Alpine (apk).
- Package-naming nuances: Fedora/RHEL 8+ uses `iptables-nft` (provides `iptables`).
- Repo prerequisites: Alpine `community` repo and RHEL 7 `extras` channel must be
  enabled for `podman` (default-on for standard Alpine / CentOS 7, but worth asserting).
- Podman: verify the version meets kind's requirement (≥ 3.0); on Debian/Ubuntu the
  native package lags, so the Homebrew bootstrap path must be exercised.

## Notes
CI runner selection matters: nested virtualization for Linux VMs is not supported on
Apple M1 (appears to require M3+).
Ref: https://apple.stackexchange.com/questions/466761/is-nested-virtualization-is-supported-by-apple-silicon-chips-m2-m3

PR #4911, opened 2026-06-30
test(ci): add per-distro install validation workflow (Ubuntu)

ddeskov-limechain:
## Summary
Foundation for automated per-distro install validation. Adds a GitHub Actions
workflow that runs Solo's **real** native package-manager install path inside a
disposable Ubuntu container and asserts `git` and `iptables` land on the system.

Today the only coverage for the per-distro install logic (#4773) is mocked unit
tests for package-manager *selection* — nothing exercises an actual install, so
per-distro regressions (package naming, repo availability) would go undetected.

This is a thin vertical slice: a single distro wired as a one-row matrix, proving
the harness end to end. Follow-up work fans the matrix out across the remaining
supported distributions by appending entries to `matrix.include`.

## Changes
- `.github/workflows/flow-install-validation.yaml` — runs in a `container: ubuntu:24.04`
  job (no VM / nested virtualization). Bootstraps git/curl/sudo, checks out, `npm ci`,
  then runs the validation test with `SOLO_OS_INSTALL_VALIDATION=true`. Triggers on PRs
  to `main` and `workflow_dispatch`.
- `test/e2e/integration/core/package-managers/os-package-manager-install.test.ts` —
  exercises the real `OsPackageManager` → `AptGetPackageManager` path
  (`update()` + `installPackages(['git', 'iptables'])`) and asserts both binaries
  resolve on PATH. Gated behind `SOLO_OS_INSTALL_VALIDATION` so it only runs inside CI
  distribution containers and never mutates a developer host.

## Scope
- Distro: Ubuntu (apt-get) only.
- Validates: the native package-manager install path for `git` and `iptables`.

## Out of scope (tracked in #4888)
- Remaining distros (Debian, Fedora/Rocky/Alma, CentOS 7, openSUSE, Arch, Alpine) —
  matrix fan-out follow-up.
- Podman + Homebrew bootstrap and `kind` runtime validation — separate follow-up.
  On Linux, Solo obtains podman via the brew bootstrap rather than the native package
  manager, and running podman/kind needs a privileged/VM-capable runner, so it's
  deliberately kept out of this container-based slice.

## Testing
- The workflow is the test — a green job is the validation.
- Locally, the test skips cleanly when `SOLO_OS_INSTALL_VALIDATION` is unset (verified:
  1 pending, no system mutation).
- Container sanity check:
  `docker run --rm ubuntu:24.04 bash -c '<bootstrap + npm ci + mocha>'`.

Part of #4888

### DIFF
--- a/.github/workflows/flow-install-validation.yaml
+++ b/.github/workflows/flow-install-validation.yaml
@@ -65,7 +65,13 @@ jobs:
       - name: Install Dependencies
         run: npm ci
 
+      - name: Install Task
+        uses: arduino/setup-task@b91d5d2c96a56797b48ac1e0e89220bf64044611 # v2.0.0
+        with:
+          version: 3.49.1
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Validate System-Dependency Install
         env:
           SOLO_OS_INSTALL_VALIDATION: 'true'
-        run: npx mocha 'test/e2e/integration/core/package-managers/os-package-manager-install.test.ts'
+        run: task test-install-validation
--- a/.github/workflows/flow-install-validation.yaml
+++ b/.github/workflows/flow-install-validation.yaml
@@ -1,0 +1,71 @@
+##
+# Copyright (C) 2025 Hedera Hashgraph, LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


.github/workflows/flow-install-validation.yaml
```diff
@@ -0,0 +1,71 @@
...
+        with:
+          node-version: 22
+
+      - name: Install Dependencies
+        run: npm ci
+
+      - name: Validate System-Dependency Install
+        env:
+          SOLO_OS_INSTALL_VALIDATION: 'true'
+        run: npx mocha 'test/e2e/integration/core/package-managers/os-package-manager-install.test.ts'
```
jeromy-cannon: we use the pattern of leveraging go:Task with a Taskfile task for running the test.  Please update to leverage this pattern.

example https://github.com/hiero-ledger/solo/blob/main/Taskfile.tests.yml#L103

ddeskov-limechain: Done in commit 7fbdaf9e

jan-milenkov (commented):
lgtm, @jeromy-cannon please review too

jan-milenkov (approved):

PR #4925, opened 2026-07-01
ci: add clean-environment one-shot workflow to verify published Solo package install

### LINKED ISSUE #3937: Add test to run Solo in a clean environment
We need to validate that Solo can successfully install and run in clean environments (no prior setup, no cached images, no residual state). This ensures a reliable first time user experience and helps identify hidden dependencies or environment specific issues.

ddeskov-limechain:
## What

Adds a new GitHub Actions workflow, `flow-clean-environment-one-shot.yaml`, that validates
Solo can **install and run from a clean state** — reproducing a first-time user experience
with no prior setup, no cached images, and no residual state.

Closes #3937.

## Why this approach (and an open question)

After reading the issue, I wasn't certain *where* this test should ultimately live — options ranged from a
minimal-container (Docker-in-Docker) job that most faithfully simulates a bare OS, to a
plain `test/` integration test, to a CI workflow on a standard runner. I went with the
**standard-runner + packed global install** approach as the current solution because it's
the smallest, lowest-maintenance change that actually exercises the real install path,
and it slots cleanly into the existing `flow-*-one-shot` workflow family. **Happy to move
it into a container-based job or a different trigger if reviewers prefer** — flagging this
as the main decision to sanity-check.

## The current solution

The workflow differs from the existing `flow-standard-runner-one-shot.yaml` in one key way:
that workflow builds from source and runs `npm run solo`, so it never exercises the path a
real first-time user hits. This one instead:

1. `task build:compile` → `npm pack` → `npm install -g ./hashgraph-solo-*.tgz` — installs
   the current build **globally**, running the package `postinstall`, and then invokes the
   installed `solo` binary directly (not `npm run solo`). This surfaces packaging, `bin`,
   `postinstall`, and dependency-bootstrap issues that building from source would mask.
2. Runs `solo one-shot single deploy --dev`, verifies port-forwarding and the Mirror Node
   REST API, then always runs `solo one-shot single destroy` and uploads logs.

"Clean environment" is enforced end-to-end:

- **No cached images:** `SOLO_NO_CACHE=true` on install skips postinstall image-cache
  population, and `ENABLE_IMAGE_CACHE=false` on deploy forces a fresh image pull.
- **No residual state / no prior setup:** each run is a fresh `ubuntu-latest` runner, and
  Solo bootstraps Helm/Kind/kubectl into `~/.solo/bin` itself.

## Trigger

Nightly (`0 3 * * *` UTC, offset from the 06:00 extended nightly tests) plus
`workflow_dispatch`. A full deploy takes ~70 min, so it's kept off the per-PR critical
path while still catching regressions daily. Easy to switch to per-PR if we want a
stronger gate.


.github/workflows/flow-clean-environment-one-shot.yaml
```diff
@@ -0,0 +1,146 @@
...
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+env:
+  STEP_TIMEOUT_MINUTES: '5'
+
+jobs:
+  clean-environment-one-shot:
+    name: Clean Environment Deploy and Destroy
+    runs-on: ubuntu-latest
```
nathanklick: Please use our self-hosted runner vs. a github hosted runner.

ddeskov-limechain: Done

JeffreyDallas (commented):
I suggest not adding another new workflow, 
maybe just enhance existing one-shot work flow 
https://github.com/hiero-ledger/solo/blob/main/.github/workflows/flow-one-shot-timing.yaml
or
https://github.com/hiero-ledger/solo/blob/main/.github/workflows/flow-standard-runner-one-shot.yaml

ddeskov-limechain:
Done in 319d115de58b9a5a97cef3829a6924bedf28c1fe — removed the new workflow and folded the test into `flow-standard-runner-one-shot.yaml` instead.

The workflow now runs a two-leg matrix on `install-mode`:

- `source` - the existing job, unchanged: `task build:compile` + `npm run solo`  from the repo root. The job name is still "Standard Runner Deploy and Destroy", so the required status check keeps matching.

- `global-package` - the clean-environment test: pack via the dual-publish release tasks, `npm install -g `the `@hiero-ledger/solo` tarball (with `SOLO_NO_CACHE`), and run every `solo` command from a directory outside the checkout. The deploy/verify/destroy/log-upload steps are shared between both legs.

Two things changed relative to the standalone workflow:

1. It now runs on ubuntu-latest instead of `hiero-solo-linux-large`. Ephemeral GitHub-hosted runners are clean by construction, so the whole "Make Clean" teardown (uninstall Solo, `rm -rf ~/.solo`, remove leftover Kind clusters) is no longer needed — it only existed because the self-hosted runner is reused.

2. The "Assert Clean Environment" step now only checks Solo-managed state (`solo` absent from PATH, no `~/.solo`). It no longer asserts helm/kind/kubectl absence, since the GitHub runner image ships those preinstalled — which is also realistic for a first-time user, and irrelevant since Solo installs its own copies into `~/.solo/bin`.

jan-milenkov (dismissed):

.github/workflows/flow-clean-environment-one-shot.yaml
jeromy-cannon: Nathan meant for you to use:
```
    runs-on: hiero-solo-linux-large
```
but, instead you consolidated your workflow into the existing workflow: `flow-standard-runner-one-shot.yaml` that runs with:
```
    runs-on: ubuntu-latest
```
which is a GitHub hosted runner.

If you want to matrix it because of how much the workflow is redundant, I recommend matrixing the runner also, using the `ubuntu-latest` for `Standard Runner Deploy and Destroy` and `hiero-solo-linux-large` for `Clean Environment Deploy and Destroy`

ddeskov-limechain: Fixed by matrixing the runner

jeromy-cannon (approved):

PR #4926, opened 2026-07-01
chore: improvements to `one-shot falcon prepare` 

### LINKED ISSUE #4043: Improvements to one-shot falcon prepare
The current implementation of `one-shot falcon prepare` looks difficult to maintain. It can be improved by adding a `yaml` configuration file that controls the default overwrites of the flags and the overall prompt workflow.
Also, at this point, it makes sense to split the `one-shot` commands into separate classes.
All of this can be done in a follow-up PR.

_Originally posted by @Ivo-Yankov in https://github.com/hiero-ledger/solo/issues/3879#issuecomment-4288266066_

jan-milenkov:
## Description

Make `one-shot falcon prepare` data-driven: the per-section flag overrides, blocked flags, legacy keys, and the prompt workflow now live in a bundled `resources/one-shot-falcon-prepare.yaml` spec.
                                                                                                                                                                                                                                           > Original behavior is preserved and generated values file are identical                                                                                                                                                     
  
### Related Issues

* Closes # https://github.com/hiero-ledger/solo/issues/4043


src/commands/node/tasks.ts
jan-milenkov, line 1: lint-fixes only

JeffreyDallas:
This feature seems not user friendly and not necessary: user need to prepare a yaml to be able to generate yaml value files used by one-shot falcon. 

Why not just ask user directly provides yaml value file, no need this extra step.

@Reccetech @jeromy-cannon 
Either way , user need to get a familiar with a yaml format and syntax. no reason to add extra step.

jan-milenkov:
> This feature seems not user friendly and not necessary: user need to prepare a yaml to be able to generate yaml value files used by one-shot falcon.
> 
> Why not just ask user directly provides yaml value file, no need this extra step.
> 
> @Reccetech @jeromy-cannon Either way , user need to get a familiar with a yaml format and syntax. no reason to add extra step.

This is for the solo team to control what the user will be prompted via a centralized file

simply put, the file is for us not for the users

(checking the linked issue might help shed some light #4043 )


This is the header on the config file,
https://github.com/hiero-ledger/solo/pull/4926/changes#diff-e6fc0ac61bf11df1ac8d00d04cfceb0adcac7cc2922aed4350bd9a8e0162afdfR3-R15

src/commands/one-shot/default-one-shot.ts
```diff
@@ -746,121 +723,85 @@ export class DefaultOneShotCommand extends BaseCommand implements OneShotCommand
...
+          ? DefaultOneShotCommand.resolveFalconValue(section.overrides[flag.name], flag, config)
+          : '';
     }
-    return section;
+
+    // Keys forced in regardless of the flag-list (legacy version keys kept for backward
+    // compatibility with existing templates, tests, and user-edited values files).
+    for (const [flagName, raw] of Object.entries(section.extraKeys ?? {})) {
+      const flag: CommandFlag | undefined = flags.allFlagsMap.get(flagName);
+      if (flag) {
```
ddeskov-limechain: an unknown flag name in the spec is silently dropped here, and the same happens for `promptFlags` (line 847) and `setConfig` (line 851). That's inconsistent with the unknown-`flagsFrom` path right above, which throws `falconValuesPreparationFailed`.

The whole point of this PR is that editing the YAML changes behavior without touching TypeScript — but the old code had compile-time safety on these names (`flags.releaseTag` couldn't be misspelled), and this trades that away. A typo in `extraKeys` or `promptFlags` now silently removes a key from the generated file or a step from the wizard. Suggest failing fast in all three places:

```suggestion
      if (!flag) {
        throw new SoloErrors.component.falconValuesPreparationFailed(
          new Error(`Unknown flag '${flagName}' in extraKeys for section '${section.name}'`),
        );
      }
```

Same guard applies to the `${config.<key>}` resolution in `resolveFalconValue` — an unknown config key currently resolves to `undefined`.

src/commands/one-shot/falcon-prepare-spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    flagsFrom: string;
+    /** Override values, keyed by flag name, applied only when the flag is in the flag-list. */
+    overrides?: Record<string, FalconOverrideValue>;
+    /** Keys forced into the section regardless of the flag-list, keyed by flag name. */
+    extraKeys?: Record<string, FalconOverrideValue>;
+  }>;
+
+  /** Interactive steps run after the base flag prompts. */
+  prompts: Array<{
+    type: 'confirm';
```
ddeskov-limechain, line 29: `type` is declared as `'confirm'` but `runFalconPreparePrompts` never checks it — a future spec entry with a different type would silently run as a confirm prompt. Either validate it (throw on anything other than `'confirm'`) or drop the field until a second prompt type actually exists.

src/commands/one-shot/falcon-prepare-spec-loader.ts
```diff
@@ -0,0 +1,22 @@
...
+import {type FalconPrepareSpec} from './falcon-prepare-spec.js';
+
+/**
+ * Loads and caches the bundled `one-shot-falcon-prepare.yaml` spec that drives
+ * `solo one-shot falcon prepare`.
+ */
+export class FalconPrepareSpecLoader {
+  private static cached: FalconPrepareSpec | undefined;
+
+  public static load(): FalconPrepareSpec {
```
ddeskov-limechain, line 15: the existing `one-shot-falcon-prepare.test.ts` tests pin the generated output well, but the new machinery introduced here has no direct unit coverage: the unknown-`flagsFrom` error path, the `${default}`/`${config.<key>}`/literal resolution in `resolveFalconValue`, and the quiet-mode branch of `runFalconPreparePrompts`. We should be catching these in unit tests which run far faster and more frequently than the One Shot E2E suites.

src/commands/node/tasks.ts
```diff
@@ -750,7 +750,7 @@ export class NodeCommandTasks {
   private _generateGossipKeys(generateMultiple: boolean): SoloListrTask<NodeKeysContext | NodeAddContext> {
     return {
       title: 'Generate gossip keys',
-      task: ({config}, task): any => {
+      task: ({config}, task): SoloListr<AnyListrContext> => {
```
ddeskov-limechain, line 752: the typing cleanups in this file (this one plus the `: any` removals at lines ~781, ~881, ~1083, ~1233, ~2860, ~4142 and the `addWrapsLib` context union) are nice improvements, but they look unrelated to `one-shot falcon prepare` — intentional to bundle them here rather than a separate chore PR?

ddeskov-limechain (dismissed):
LGTM. Comments are proposals from Claude that are noted but can be ignored according to me

Ivo-Yankov (dismissed):

src/commands/node/tasks.ts
jan-milenkov: as we are not doing PRs specifically for fixing those, there is no problem

src/commands/one-shot/default-one-shot.ts
jan-milenkov: updated

src/commands/one-shot/falcon-prepare-spec.ts
jan-milenkov: updated

src/commands/one-shot/falcon-prepare-spec-loader.ts
jan-milenkov: updated

jeromy-cannon (approved):

PR #4946, opened 2026-07-03
fix: Improve cache speed and reliability

jan-milenkov:
## Description

Improve Solo image cache reliability and visibility:
- More reliable image caching and loading, with handling of any failures (logged and surfaced in an end-of-run summary instead of passing silently).
- In one-shot, cache pull/load are now their own top-level step with per-image progress visible.
- prune is now idempotent, and status handles an empty cache cleanly.

### Related Issues

* Closes # N/A
- PR for documentation update # https://github.com/hiero-ledger/solo-docs/pull/269


src/commands/cache.ts
```diff
@@ -484,10 +486,9 @@ export class CacheCommand extends BaseCommand {
     return {
       title: 'Pull and cache container images',
       task: async ({config: {imageCacheHandler}}, task): Promise<SoloListr<AnyListrContext>> => {
-        return task.newListr(
-          await imageCacheHandler.pull(),
-          constants.LISTR_DEFAULT_OPTIONS.WITH_CONCURRENCY_COLLAPSABLE,
-        );
+        // WITH_CONCURRENCY (not the collapsable variant) so each image's step stays visible
+        // after it completes, for easier inspection of what was cached.
+        return task.newListr(await imageCacheHandler.pull(), constants.LISTR_DEFAULT_OPTIONS.WITH_CONCURRENCY);
```
jeromy-cannon: This pull path stays unbounded (`WITH_CONCURRENCY` = `concurrent: true`), so up to ~30 crane pulls run at once against Docker Hub. Yet the new `CACHE_IMAGE_MAX_CONCURRENCY` bound is applied only to `load` below (line 506), and its comment in `constants.ts` justifies the bound by "amplifying Docker Hub rate limits when pulling ~30 images." The rate-limit pressure is here on pull, not on load (which is a local `ctr import`). If the intent is to throttle Docker Hub, the bound belongs on this `newListr`.  Perhaps we should limit both the pull and the load.

src/commands/cache.ts
```diff
@@ -496,14 +497,14 @@ export class CacheCommand extends BaseCommand {
...
-        subTasks.push(...newTasks);
-
-        return task.newListr(subTasks, constants.LISTR_DEFAULT_OPTIONS.WITH_CONCURRENCY_COLLAPSABLE);
+        const clusterName: string = this.prepareClusterName(this.k8Factory.getK8(context).clusters().readCurrent());
+        const subTasks: SoloListrTask<CacheLoadContext>[] = await imageCacheHandler.load(clusterName);
+
+        // Load images concurrently (bounded) and keep each step visible after it completes.
+        return task.newListr(subTasks, {
+          ...constants.LISTR_DEFAULT_OPTIONS.WITH_CONCURRENCY,
+          concurrent: constants.CACHE_IMAGE_MAX_CONCURRENCY,
```
jeromy-cannon, line 508: `load` bounds concurrency but doesn't touch the network/pull — `loadImageArchiveIntoCluster` is a local `ctr import` of an already-cached archive. Perhaps we should have separate concurrency limiting on both pull and load?

src/commands/cache.ts
```diff
@@ -327,20 +322,17 @@ export class CacheCommand extends BaseCommand {
...
-            const config: CachePruneConfigClass = {
-              imageCacheHandler: await this.buildImageCacheHandlerFromRenderedFile(cacheDirectory),
-            };
-
-            context_.config = config;
-
-            await config.imageCacheHandler.prune();
+            // Wipe the whole cache directory and the rendered targets file. Idempotent: it does not
+            // require the cache to be materialized and never fails when there is nothing to prune.
+            await this.cacheCatalogStore.clear();
```
jeromy-cannon, line 334: This bypasses `ImageCacheHandler.prune()`, which is now left with no callers (and its `CacheOperationHandler.prune()` interface method too). Per CLAUDE.md "Dead Code Removal," that method should be removed — or, to keep the clearing logic in one place, enhance `ImageCacheHandler.prune()` to also remove the rendered targets file and call it here. Right now prune logic is split between the command and an orphaned handler method.

src/core/constants.ts
```diff
@@ -410,6 +410,11 @@ export const LISTR_DEFAULT_OPTIONS: {
   },
 };
 
+// Maximum number of image cache operations (load into engine) that run concurrently.
+// Bounded to avoid saturating the network/disk and amplifying Docker Hub rate limits when pulling ~30 images,
+// while staying wide enough to keep a cold pull responsive. Override with CACHE_IMAGE_MAX_CONCURRENCY.
+export const CACHE_IMAGE_MAX_CONCURRENCY: number = +getEnvironmentVariable('CACHE_IMAGE_MAX_CONCURRENCY') || 12;
```
jeromy-cannon, line 419: `CACHE_IMAGE_MAX_CONCURRENCY` is a new env var read via `getEnvironmentVariable()`, so it needs documenting. `env.md` no longer lives in-repo (moved to solo-docs), so this needs a companion solo-docs PR — CLAUDE.md "Environment Variable Documentation."

Separately, the comment says the bound guards against Docker Hub rate limits "when pulling ~30 images," but the bound is applied to `load` (local import), not `pull`. Either move the bound to the pull path or reword this comment so it matches where it's actually applied.

src/integration/container-engine/docker-client.ts
```diff
@@ -56,6 +57,37 @@ export class DockerClient implements ContainerEngineClient {
     });
   }
 
+  public async saveImageArchive(image: string, archivePath: string): Promise<void> {
```
jeromy-cannon: `saveImageArchive` and `saveImage` (line 47) share the same preamble — `fs.mkdir(path.dirname(...))`, `Architecture.getLinuxPlatform()`, and `dependencyManager.getExecutable(constants.CRANE)`. Two occurrences, so just a note: consider a small `private` helper for the shared setup. (Nice comment on the OCI-layout handling, by the way — it explains the *why* clearly.)

src/commands/cache.ts
jan-milenkov: Made `CACHE_IMAGE_MAX_CONCURRENCY` so both pull  and load use the same it for their limit.

src/commands/cache.ts
jan-milenkov: Both paths are now bounded by `CACHE_IMAGE_MAX_CONCURRENCY`. Load stays bounded to avoid saturating local disk during ctr import, and pull is now bounded for the Docker Hub rate limit. Updated the load comment to say why it is bounded.

src/integration/container-engine/docker-client.ts
jan-milenkov: extracted the shared setup into a `prepareCranePull()` private helper that both saveImage and `saveImageArchive` now use

src/core/constants.ts
jan-milenkov: created a pr and linked it to the issue, the comment is also updated

src/commands/cache.ts
jan-milenkov: Removed

src/commands/cache.ts
```diff
@@ -484,10 +486,10 @@ export class CacheCommand extends BaseCommand {
     return {
       title: 'Pull and cache container images',
       task: async ({config: {imageCacheHandler}}, task): Promise<SoloListr<AnyListrContext>> => {
-        return task.newListr(
-          await imageCacheHandler.pull(),
-          constants.LISTR_DEFAULT_OPTIONS.WITH_CONCURRENCY_COLLAPSABLE,
-        );
+        return task.newListr(await imageCacheHandler.pull(), {
+          ...constants.LISTR_DEFAULT_OPTIONS.WITH_CONCURRENCY,
```
Ivo-Yankov: Please use renderer options with `collapseSubtasks: true`, because downloading and then loading ~30 images floods the terminal output. With `collapseSubtasks: true`, image tasks become the majority of rows shown to the user.

src/commands/cache.ts
```diff
@@ -496,14 +498,15 @@ export class CacheCommand extends BaseCommand {
     return {
       title: 'Load images into cluster',
       task: async ({config: {imageCacheHandler, context}}, task): Promise<SoloListr<CacheLoadContext>> => {
```
Ivo-Yankov, line 500: Is there a way to check if the images are already loaded into the cluster and possibly skip this listr task? If I run `deploy`, `destroy`, `deploy` on the same cluster, I would expect that the images do not need to be loaded on the second `deploy`.

Ivo-Yankov (changes requested):

src/commands/cache.ts
jan-milenkov: its useful as this is where failed image pulls are captured, also useful for referencing

src/commands/cache.ts
jan-milenkov: automatically done

Ivo-Yankov (approved):

jeromy-cannon (approved):

PR #4928, opened 2026-07-01
feat: solo home directory cleanup

Ivo-Yankov:
## Description

This PR fixes the following:
- Ensures that no files in SOLO_HOME have higher file permissions than 750.
- Deprecates the `--dev` flag and introduces a new, more intuitively named `--debug` flag.
- Keys are no longer stored on disk in SOLO_HOME.
- Removes the `~/.solo/cache/last-one-shot-deployment.txt` file, as it is no longer needed. It used to store the name of the deployment, and that was useful when it was randomly generated. But since it now defaults to the constant `one-shot`, the file is not needed.

### Related Issues

* Closes #1560 
* Related to #

### Pull request (PR) checklist

* \[ ] This PR added tests (unit, integration, and/or end-to-end)
* \[ ] This PR updated documentation
* \[ ] This PR added no TODOs or commented out code
* \[ ] This PR has no breaking changes
* \[ ] Any technical debt has been documented as a separate issue and linked to this PR
* \[ ] Any `package.json` changes have been explained to and approved by a repository manager
* \[ ] All related issues have been linked to this PR
* \[ ] All changes in this PR are included in the description
* \[ ] When this PR merges the commits will be squashed and the title will be used as the commit message, the 'commit message guidelines' below have been followed

### Testing

* \[ ] This PR added unit tests
* \[ ] This PR added integration/end-to-end tests
* \[ ] These changes required manual testing that is documented below
* \[ ] Anything not tested is documented

The following manual testing was done:

* TBD

The following was not tested:

* TBD

<details>
<summary>
Commit message guidelines
</summary>
We use 'Conventional Commits' to ensure that our commit messages are easy to read, follow a consistent format, and for automated release note generation. Please follow the guidelines below when writing your commit messages:

1. BREAKING CHANGE: a commit that has a footer BREAKING CHANGE:, or appends a ! after the type/scope, introduces a breaking API change (correlating with MAJOR in Semantic Versioning). A BREAKING CHANGE can be part of commits of any type.  NOTE: currently breaking changes will only bump the MAJOR version.
2. The title is prefixed with one of the following:

| Prefix    | Description                                         | Semantic Version Update | Captured in Release Notes |
|-----------|-----------------------------------------------------|-------------------------|---------------------------|
| feat:     | a new feature                                       | MINOR                   | Yes                       |
| fix:      | a bug fix                                           | PATCH                   | Yes                       |
| perf:     | performance                                         | PATCH                   | Yes                       |
[...]

PR #4993, opened 2026-07-13
feat: clear cached secrets and keys from the cache after loading them into the cluster

### LINKED ISSUE #1055: Production Readiness: clear out the secrets/credentials/keys from the cache after it has been loaded into the cluster (clean up after ourselves)

Fantomasa:
## Summary

Clear cached secrets/keys from `SOLO_CACHE_DIR` once they have been loaded into the cluster, so private key material is not left behind on disk.

Builds on #4928 (already in `main`), which introduced the on-disk key cleanup for `network deploy` — it wipes `<cacheDir>/keys` after uploading the gossip/TLS keys to cluster Secrets (gated by `--debug`) and reworked the key reads to fall back to those Secrets. This PR closes the remaining cache-cleanup gaps that #4928 did not cover.

### 1. Explorer / mirror-node ingress TLS — `src/core/key-manager.ts`
`createTlsSecret` generated `<domain>.crt` / `<domain>.key` (self-signed, private key `0600`) into the cache **root** and left them there after creating the k8s Secret. It now removes both files once the Secret is created (best-effort, so a cleanup hiccup cannot fail an otherwise-successful deploy). Safe because the cert is generate-then-read within the same call and regenerated on the next deploy — nothing reads it from disk afterward.

### 2. Standalone `node add` / `node update` — `src/commands/node/tasks.ts`, `handlers.ts`
These flows write gossip/TLS keys to `<cacheDir>/keys` but did not run the `network deploy` cleanup, so keys persisted there afterward. Added a `--debug`-gated `removeCachedKeys()` task, wired as the **final** task of the add-execute and update-execute phases — after `copyNodeKeysToSecrets` and every other `keysDir` consumer (e.g. `setupNetworkNodes`). It intentionally does **not** run in `node add-prepare` (keys not yet uploaded to the cluster) or `node keys` (that command exists to produce keys on disk).

## Intentionally out of scope

- **WRAPS / TSS proving-key material** (`<cacheDir>/wraps-<version>/` + tarball) is **not** cleaned: those `.bin` files are public proving-key parameters downloaded from a public URL, and the directory is deliberately cached/reused across deploys (`network.ts` skips the re-download when it already exists). Deleting it would force a re-download on every deploy with no security benefit.
- Matches #4928's `keysDir`-scope; staging directories are handled the same as before.

## Verification

- Unit test added for the ingress-TLS cleanup (`test/unit/core/key-manager.test.ts`).
- `tsc` and ESLint clean; verified **no lint-warning-count increase** vs `main` on every changed file.
- Node-lifecycle cleanup is validated by CI E2E — **Node Add Local** and the separate-node-add suite exercise `node add`; **Standard** / **Dual Cluster Full** exercise the explorer/mirror ingress-TLS path.

Closes #1055


JeffreyDallas (dismissed):

Fantomasa:
@JeffreyDallas sorry I solved merge conflicts

JeffreyDallas (approved):

PR #5038, opened 2026-07-14
fix: prevent `deployment refresh port-forwards` from binding the mirror port-forward to the wrong pod

### LINKED ISSUE #4966: bug: solo refresh port-forwards only restores Explorer and HAProxy; Mirror REST and Relay require manual kubectl
**Reporters:** Luis Vega

**Solo version:** 0.77.0
**OS:** macOS 16GB

**Observed behavior**
After running `solo refresh port-forwards`, only the Explorer and HAProxy port-forwards are restored. The Mirror REST API and JSON-RPC Relay port-forwards are not restored and require the user to manually run `kubectl port-forward` for each.

The docs reference `solo refresh port-forwards` as the mechanism to restore port-forwards after a terminal restart or reconnect, but this is incomplete.

**Expected behavior**
`solo refresh port-forwards` should restore all port-forwards that were active at the time of `solo one-shot single deploy`, including:
- Mirror ingress controller (default 38081)
- JSON-RPC relay (default 37546)
- Explorer (default 38080)
- Any other endpoints listed in the deploy summary

Alternatively, the docs should enumerate which endpoints are NOT restored and provide the manual `kubectl port-forward` commands for each.

**User feedback**
Luis Vega: "Documented port-forwards not all restored. refresh port-forwards only covers Explorer + HAProxy; Mirror REST and Relay need manual kubectl."

jan-milenkov:
## Description

`refresh port-forwards` resolved the MirrorNode pod via the shared `app.kubernetes.io/name=haproxy-ingress`
label and, when the mirror ingress pod wasn't matched, fell back to `pods[0]` — which could be another
controller (e.g. the Explorer ingress). It then bound mirror's port-forward to the wrong pod and reported
"Restored".

Now mirror only binds its own ingress pod (never `pods[0]`), and when a pod can't be resolved the command
prints the exact manual `kubectl port-forward` command instead of a misleading success.

### Related Issues

* Closes #4966


Fantomasa (approved):
LGTM

jeromy-cannon (approved):

PR #5048, opened 2026-07-15
fix: reconcile mirror Postgres passwords on re-deploy with a reused PVC

### LINKED ISSUE #4961: bug: stale PVCs on re-deploy cause silent cascading failure (postgres auth, mirror, accounts)
**Reporters:** Luis Vega

**Solo version:** 0.77.0
**OS:** macOS 16GB

**Observed behavior**
After a failed or partial first deploy followed by `solo one-shot single destroy` and a fresh `solo one-shot single deploy`, Solo silently reuses stale PVCs. The Postgres pod retains the old password from the previous deployment. This causes:
- mirror-rest, mirror-restjava, mirror-pinger, mirror-web3 pods to fail
- 8 of 10 account creation calls to fail
- The root cause (Postgres auth mismatch) is buried in diagnostics logs and is not surfaced in the deploy output

**Expected behavior**
On re-deploy, Solo should either:
1. Detect and delete stale PVCs automatically, or
2. Provide an explicit warning that stale PVCs exist and offer a `--reset` / `--reuse` flag, or
3. Surface the Postgres auth failure with an actionable error: "Postgres auth failed — run `solo deployment reset --shared-resources`"

**Steps to reproduce**
1. Run `solo one-shot single deploy` — allow it to fail mid-way (or fail intentionally)
2. Run `solo one-shot single destroy`
3. Run `solo one-shot single deploy` again
4. Observe mirror pods failing silently; inspect logs to find Postgres auth mismatch

**User feedback**
Luis Vega: "Stale PVCs on re-deploy → silent cascading failure. Postgres kept old password; mirror-rest, restjava, pinger, web3 and 8/10 accounts failed. Root cause hidden in diagnostics logs."

jan-milenkov:
## Description

When a Postgres PVC survives a destroy and is reused on the next deploy, the DB keeps the old
passwords while the `mirror-passwords` Secret is regenerated. The init script saw the `solo-initialized`
sentinel and exited early, so it never re-applied the new passwords - causing an auth mismatch that
broke `mirror-rest/restjava/pinger/web3` and account creation.

Now, when the sentinel is found, the init script re-applies the current Secret's passwords to every
mirror role instead of skipping, keeping the DB in sync with the Secret.

### Related Issues

* Closes # https://github.com/hiero-ledger/solo/issues/4961


src/core/shared-resources/postgres.ts
```diff
@@ -214,7 +214,16 @@ export class PostgresSharedResource {
...
-          `  echo "Initialization sentinel found on database '${databaseName}' — already complete, skipping."`,
+          `  echo "Initialization sentinel found on database '${databaseName}' — reconciling role passwords and skipping re-init."`,
+          '  # The database and roles already exist, but the mirror-passwords Secret may have been',
+          '  # regenerated while this PVC survived a prior destroy, leaving the stored passwords stale.',
+          '  # Re-apply the current passwords to every role to avoid an auth mismatch.',
+          '  # Best-effort per role, since optional roles (graphql/rosetta) may not exist.',
+          '  reconcile_roles=("${OWNER_USERNAME}" mirror_importer "${REST_USERNAME}" mirror_rest_java mirror_web3 mirror_grpc mirror_graphql mirror_rosetta)',
+          '  reconcile_passwords=("${OWNER_PASSWORD}" "${IMPORTER_PASSWORD}" "${REST_PASSWORD}" "${REST_JAVA_PASSWORD}" "${WEB3_PASSWORD}" "${GRPC_PASSWORD}" "${GRAPHQL_PASSWORD}" "${ROSETTA_PASSWORD}")',
+          '  for i in "${!reconcile_roles[@]}"; do',
+          '    psql -c "ALTER USER ${reconcile_roles[$i]} WITH PASSWORD \'${reconcile_passwords[$i]}\';" || true',
```
Fantomasa: If one of the `reconcile_roles` doesn't exist, the failing `ALTER USER` leaks the password
  into the postgres server log: with the default `log_min_error_statement=error` postgres logs
  the failed statement verbatim (`ERROR:` + `STATEMENT: ALTER USER ... WITH PASSWORD '...'`),
  and that pod log gets collected into diagnostics bundles. Guard on role existence instead
  of `|| true` — then a failing statement is never sent, and real `ALTER` failures are no
  longer swallowed:

  ```suggestion
            '    role_exists=$(psql -tc "SELECT 1 FROM pg_roles WHERE rolname =
  \'${reconcile_roles[$i]}\'" | tr -d \'[:space:]\')',
            '    if [[ "${role_exists}" == "1" ]]; then',
            '      psql -c "ALTER USER ${reconcile_roles[$i]} WITH PASSWORD
  \'${reconcile_passwords[$i]}\';"',
            '    fi',
  ```

jan-milenkov: thank you, updated

Fantomasa (approved):
LGTM

jeromy-cannon (approved):

PR #5316, opened 2026-07-24
fix: handle parseable-but-partial local config instead of silently loading it as empty

### LINKED ISSUE #5123: fix: handle parseable-but-partial local config (`missing deployments/clusterRefs`)
## Description

A partial `local-config.yaml` (parses, but missing required top-level keys) silently null-coalesces to a
valid-but-empty config and only surfaces later as a confusing `DeploymentNotFound`
(`local-config.ts:24-40,58`) instead of "your config is incomplete".

- **Do:** ensure local config is created/written **atomically** (all-or-nothing) so an interrupted write
  can't leave a partial file; and/or leverage the import command (SC-LC-5) to reconstruct an incomplete
  config. **Related to SC-LC-5.**
- **Acceptance:** an interrupted/partial write never yields a silently-empty config — either the write is
  atomic, or the incomplete state is detected and the user is guided to recover with `solo deployment config import`.

## Related to

- Blocked by: https://github.com/hiero-ledger/solo/issues/5122

Fantomasa:
## Description

A `local-config.yaml` that parses as valid YAML but is missing required top-level keys (e.g. left behind by an interrupted write) silently null-coalesced into a valid-but-empty config and only surfaced later as a confusing `DeploymentNotFound`. This PR closes that hole from both directions required by the issue's acceptance criteria:

- **Atomic writes:** `FileStorageBackend.writeBytes` now writes to a temp file in the same directory and renames it over the target, so an interrupted write can no longer leave a partial file. Failed writes clean up the temp file.
- **Detection:** at load, the raw parsed config must contain the `clusterRefs` and `deployments` top-level keys (present in every schema version, including legacy pre-`schemaVersion` files — the v0 migration path is covered by a test). A partial file now fails fast with the new `IncompleteLocalConfigError` (SOLO-1005) naming the file path and the missing keys, and pointing at `solo deployment config import` for recovery.
- **Recovery path unblocked:** `solo deployment config import` itself loaded the local config and would have died on the very error that recommends it. Its Initialize task now catches `RefreshLocalConfigSourceError`/`IncompleteLocalConfigError`, warns the user, and regenerates the config, which the import then fills from the cluster's remote config.

New tests cover: partial config missing both/one required key, legacy v0 config still migrating, atomic overwrite with no leftover temp files (success and failure paths), and end-to-end recovery of `importConfig` from a partial local config.

> [!NOTE]
> Stacked on #5314 (`fix: point the malformed local-config error at the config-import recovery command`) — its commit appears in this diff until it merges.

## Related Issues

Fixes #5123


jan-milenkov (approved):

src/core/errors/solo-errors.ts
jan-milenkov, line 1: leaving this comment to avoid adding to merge queue while merges are paused (because of planned release)

jan-milenkov (dismissed):

ddeskov-limechain:
Just a hint: Before regenerating, rename the broken file aside (e.g. local-config.yaml.invalid) instead of overwriting it. It is one `fs.renameSync` line, and the warning message can name the backup. That keeps the import retry-able and the manual path alive after a failed import.

jan-milenkov (approved):

PR #5356, opened 2026-07-27
feat: add load-balancer switch to relay and explorer

### LINKED ISSUE #3311: Add load-balancer switch to Relay and Explorer
Need to add a load-balancer switch to Relay and Explorer (currently Nathan had to do a values file override)

Fantomasa:
## Description

Adds the `--load-balancer` switch to the relay and explorer `deploy`/`upgrade` commands, so exposing them externally no longer requires a values-file override, and makes the commands wait until the load balancer address is actually provisioned.

Design notes:

* Reuses the existing `Flags.loadBalancerEnabled` flag (previously used only by `network deploy`) rather than introducing a new one; its description is generalized since it now applies to three commands.
* When set, relay flips `relay.service.type` and `ws.service.type` to `LoadBalancer` (both default to `ClusterIP` in the `hedera-json-rpc` chart), and explorer flips the chart's root `service.type` (also `ClusterIP` by default).
* The explorer ingress-controller path needs no change: the haproxy-ingress chart already defaults `controller.service.type` to `LoadBalancer`, and `--explorer-static-ip` already targets it.
* A new `Services.waitForLoadBalancerAddress` method on the K8 client wrapper polls until every LoadBalancer-typed service matching the label selector has an ingress IP or hostname. Relay and explorer run a `Check load balancer is assigned` task (skipped unless `--load-balancer`), so the command only succeeds once the address is up. The `network deploy` load-balancer check is refactored onto the same method, resolving the inline TODO there and fixing a tight retry loop that skipped both the attempt counter and the delay when an ingress entry had no address yet.
* `LoadBalancerNotFoundSoloError` now accepts an optional cause, and its troubleshooting hint no longer assumes the network-node label since the error is raised for relay and explorer too.

Unit tests cover the enabled and disabled chart-value paths for relay and explorer, the new wait method (address assigned late, non-LB services ignored, timeout), and the updated explorer task flows. Verified the flag surfaces on all four subcommands via `--help`.

### Related Issues

* Fixes #3311

### DIFF
--- a/src/commands/explorer.ts
+++ b/src/commands/explorer.ts
@@ -59,6 +59,7 @@ interface ExplorerDeployConfigClass {
   explorerStaticIp: string | '';
   explorerVersion: string;
   componentImage: Optional<string>;
+  loadBalancerEnabled: boolean;
   namespace: NamespaceName;
   tlsClusterIssuerType: string;
   valuesFile: string;
@@ -99,6 +100,7 @@ interface ExplorerUpgradeConfigClass {
   explorerStaticIp: string | '';
   explorerVersion: string;
   componentImage: Optional<string>;
+  loadBalancerEnabled: boolean;
   namespace: NamespaceName;
   tlsClusterIssuerType: string;
   valuesFile: string;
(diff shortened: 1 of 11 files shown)


jan-milenkov (approved):

src/commands/explorer.ts
jan-milenkov, line 1: Leaving comment to avoid accidental merge, resolve once the merge queue is unblocked

jan-milenkov (approved):

PR #5357, opened 2026-07-27
feat: allow passing ports to consensus node service endpoints

### LINKED ISSUE #3306: Allow passing ports to service endpoints
Need to be able to pass in ports to use when building the service endpoints that we pass to CN via genesis network and config.txt

jan-milenkov:
## Description

* Adds two optional flags that override the ports used when building the consensus node
  endpoints it publishes to the network:
  * `--gossip-endpoint-port` — port for the gossip endpoints (default: `50111`, from
    `SOLO_NODE_EXTERNAL_GOSSIP_PORT`)
  * `--service-endpoint-port` — port for the gRPC service endpoints (default: `50211`, from
    `GRPC_PORT`)
* Each flag accepts either a single port applied to every node, per-node `alias=port` pairs, or a
  combination where the alias-less port is the default for nodes not listed explicitly:

  ```bash
  --gossip-endpoint-port 50111
  --service-endpoint-port node1=50211,node2=50212
  --service-endpoint-port 50211,node2=50212
  ```

### Related Issues

* Closes # https://github.com/hiero-ledger/solo/issues/3306

### DIFF
--- a/src/commands/flags.ts
+++ b/src/commands/flags.ts
@@ -2995,6 +2995,32 @@ export class Flags {
     prompt: undefined,
   };
 
+  public static readonly gossipEndpointPort: CommandFlag = {
+    constName: 'gossipEndpointPort',
+    name: 'gossip-endpoint-port',
+    definition: {
+      describe:
+        'Port used when building the consensus node gossip endpoints published to the network' +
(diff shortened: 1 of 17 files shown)


ddeskov-limechain (approved):
LGTM

Ivo-Yankov (approved):

PR #5358, opened 2026-07-27
ci: bump cn to 0.75

JeffreyDallas:
## Description

* Bumps the default Consensus Node version to `v0.75.1` and Block Node version to `0.38.1`.
* Updates version-upgrade test coverage to exercise `v0.74.0 -> v0.75.1` with prior Block Node `v0.37.0`, Mirror Node `v0.158.0`, and Relay `0.76.0`.
* Adds CN `v0.74+` block-stream compatibility handling so Solo can default to pure `BLOCKS` mode when TSS/block-node integration is enabled, while preserving explicit `BOTH` or `RECORDS` modes for compatibility tests.
* Keeps MinIO and record-stream import enabled when workflows intentionally select `BOTH` or `RECORDS`, so mirror/relay smoke tests are not forced through the block-node live-subscriber path.
* Adds Block Node RSA/TSS bootstrap support, including REST Java bootstrap URL selection, generated `rsa-bootstrap-roster.json` values, refreshed bootstrap state after node add/update, block-node source mapping, and larger TSS message-size/memory defaults.
* Stabilizes node/network upgrades with `--freeze-block-drain-seconds`, `--skip-node-start`, upgrade prepare/execute config-file flags, post-upgrade block-node readiness checks, and safer gossip endpoint/address handling for node lifecycle tests.
* Adds endpoint/load-balancer handling for node lifecycle and multi-cluster tests: generated gossip `LoadBalancer` services for new nodes when needed, waits for assigned LB IP/hostname, avoids unsafe FQDN gossip endpoints, validates saved IP endpoints against current service IPs, and publishes routable gRPC node-service addresses so pinger/SDK clients do not hang on bootstrap FQDNs.
* Updates migration and example workflows with block progress gates, mirror/relay/BN smoke checks, pinger preservation during mirror upgrade, and temporary `BOTH` mode workarounds while upstream block-stream importer issues are tracked.
* Updates the bundled block-node smoke-test proto archive so `grpcurl` can parse hook-aware contract result protos.
* Adds/updates unit and e2e coverage for block node values, mirror block-node integration, network stream mode selection, profile/application properties handling, node upgrade behavior, local-build PVC validation, address handling, and block-node smoke checks.
* Documents the CN `v0.75.1` migration experiments and failure analysis in `test/migration-upgrade-experiments.md`.

### Migration notes

CN `v0.75.x` changes the block-stream/TSS path enough that migration tests exposed several boundary problems:

* CN can permanently reject a block node after connection or range errors near the upgrade boundary.
* Block Node live-subscriber behavior can make mirror importer fail on batches that begin with `ROUND_HEADER`.
* Block Node WRB/RSA verification requires RSA bootstrap data from Mirror REST Java, not the legacy Node.js REST service.
* Freeze/upgrade boundaries can stop consensus nodes before the final block is available to downstream services unless Solo waits for drain/progress.

The migration workflow intentionally keeps the source path in `BOTH` mode and disables mirror importer block-node profiles as a temporary compatibility workaround while still deploying and validating Block Node.

### Related Issues

* Related to hiero-ledger/hiero-consensus-node#26456
* Related to hiero-ledger/hiero-consensus-node#26498
* Related to hiero-ledger/hiero-block-node#3150
* Related to hiero-ledger/hiero-block-node#3249

* [ ] This PR added no TODOs or commented out code

### Testing

* [x] This PR added integration/end-to-end tests

* GitHub Actions PR checks passed, including Linux/Windows unit tests, code style, docs automation, CodeQL, install validation, GCS storage tests, standard deploy/destroy, Windows deploy/destroy, migration process, performance test, and e2e suites.
* Migration and version-upgrade workflow failures were iterated through in GitHub Actions; findings are documented in `test/migration-upgrade-experiments.md`.
* Rebuilt and verified `test/data/proto.zip` so the block-node smoke test can parse `HookId`/`HookEntityId` protos before calling `BlockAccessService/getBlock`.

The following was not tested:

* N/A


jan-milenkov (dismissed):

src/commands/backup-restore.ts
```diff
@@ -1184,8 +1184,10 @@ export class BackupRestoreCommand extends BaseCommand {
   private async restoreDatabaseDump(inputDirectory: string): Promise<void> {
     const databaseDumpPath: string = PathEx.join(inputDirectory, 'database-dump.sql');
     if (!fs.existsSync(databaseDumpPath)) {
-      this.logger.info(`No database dump found at ${databaseDumpPath}; skipping database restore`);
-      return;
+      throw new SoloError(
```
jeromy-cannon: we should not throw `new SoloError(message)` directly — every error must be a dedicated subclass registered in `SoloErrors`. The rest of this file follows the pattern correctly.

The fix:
1. Create `src/core/errors/classes/validation/backup-database-dump-not-found-solo-error.ts` extending `SoloError` with `message`, `code` (new entry in `ErrorCodeRegistry`), and `troubleshootingSteps`.
2. Register it in `SoloErrors.validation` in `solo-errors.ts`.
3. Replace this throw with `throw new SoloErrors.validation.backupDatabaseDumpNotFound(databaseDumpPath)`.

See `src/core/errors/classes/validation/backup-no-log-files-solo-error.ts` for the shape to follow.

.github/workflows/script/launch_network.sh
```diff
@@ -551,149 +649,306 @@ fi
...
+    #   reconnects mid-block (hiero-block-node#3150). Mirror v0.159.0 rejects that
+    #   stream shape with "Incorrect first block item case ROUND_HEADER", reconnects
+    #   rapidly, and contract-result ingestion can stall long enough for smoke tests
+    #   to time out. Keep this migration-only buffer/HTTP2 mitigation until BN fixes
+    #   the subscriber stream boundary behavior.
+    SERVER_HTTP2_MAX_RAPID_RESETS: "500"
+    MESSAGING_BLOCK_ITEM_QUEUE_SIZE: "65536"
+  initContainers:
+    - name: init-storage-dirs
+      image: busybox
```
jeromy-cannon: `image: busybox` must be pinned to an exact version — floating tags can pull a different image on every deploy (supply-chain risk). Same fix applies at line 831.
```suggestion
      image: busybox:1.36.1
```

.github/workflows/script/launch_network.sh
```diff
@@ -551,149 +649,306 @@ fi
...
+        - name: application-state-storage
+          mountPath: /application-state-pvc
+        - name: archive-storage
+          mountPath: /archive-pvc
+        - name: live-storage
+          mountPath: /live-pvc
+        - name: logging-storage
+          mountPath: /logging-pvc
+    - name: cleanup-block-ranges
+      image: busybox
```
jeromy-cannon: same unpinned `busybox` — pin to `busybox:1.36.1`.
```suggestion
      image: busybox:1.36.1
```

src/commands/block-node.ts
```diff
@@ -407,6 +424,98 @@ export class BlockNodeCommand extends BaseCommand {
...
+
+    const bootstrapJson: string = JSON.stringify({
+      addressBooks: [{addressBook: {nodeAddress: nodeAddresses}, startBlock: '0', endBlock: '-1'}],
+    });
+    const content: string = yaml.stringify({
+      blockNode: {
+        initContainers: [
+          {
+            name: 'init-storage-dirs',
+            image: 'busybox',
```
jeromy-cannon: pin to an exact version — `'busybox:1.36.1'`. Floating `busybox` can pull a different image on every deploy.

src/core/profile-manager.ts
```diff
@@ -675,4 +718,216 @@ export class ProfileManager {
...
+      fs.writeFileSync(configFilePath, configLines.join('\n'));
+      return configFilePath;
+    } catch (error: Error | unknown) {
+      throw new SoloErrors.component.genesisDataGenerationFailed(
+        error instanceof Error ? error : new Error(String(error)),
+      );
+    }
+  }
+
+  private static hasMultipleKubernetesContexts(consensusNodes: ConsensusNode[]): boolean {
```
jeromy-cannon: this seems like a duplicate of `NodeCommandTasks.hasMultipleKubernetesContexts` (tasks.ts:241) — identical three-line private static method. Consider extracting to `Helpers` and calling from both sites. Checklist §1.

src/commands/block-node.ts
```diff
@@ -407,6 +424,98 @@ export class BlockNodeCommand extends BaseCommand {
     return chartValues;
   }
 
+  private shouldConfigureRsaMirrorBootstrapSource(): boolean {
```
jeromy-cannon, line 427: the version-check + `BLOCK_STREAM_STREAM_MODE` guard here is nearly identical to `DeployArgvBuilders.shouldInjectRsaBootstrapValuesFile` (deploy-argv-builders.ts:81) and `shouldSkipMinioSetup` (deploy-argv-builders.ts:65). Consider extracting a shared predicate to `Helpers` — something like `Helpers.requiresRsaBootstrap(consensusNodeVersion, streamMode)` — and calling it from all three sites. Checklist §1.

src/commands/node/handlers.ts
```diff
@@ -76,6 +76,10 @@ export class NodeCommandHandlers extends CommandHandler {
   private static readonly DESTROY_CONTEXT_FILE: string = 'node-destroy.json';
   private static readonly UPDATE_CONTEXT_FILE: string = 'node-update.json';
   private static readonly UPGRADE_CONTEXT_FILE: string = 'node-upgrade.json';
+  // CN does not signal when the final block has been flushed to the block node after a freeze.
+  // Without this delay, stopNodes races the in-flight blocks and the block node may miss the
+  // last block before the freeze boundary, causing a gap in mirror node ingestion.
+  private static readonly FREEZE_BLOCK_STREAM_DRAIN_MILLISECONDS: number = 20_000;
```
jeromy-cannon: the `upgradeNode` path reads `--freeze-block-drain-seconds` via `drainBlockStreamAfterFreeze()` (configurable), but `updateNode` (line 1258 in this file) uses this hardcoded constant and silently ignores any `--freeze-block-drain-seconds` the user supplies. Either call `drainBlockStreamAfterFreeze()` here too, or have the constant read `context_.config.freezeBlockDrainSeconds`. Checklist §1 (DRY) + §7 (user expectation).

src/commands/flags.ts
```diff
@@ -497,6 +497,29 @@ export class Flags {
...
+      type: 'number',
+    },
+    prompt: undefined,
+  };
+
+  public static readonly skipNodeStart: CommandFlag = {
+    constName: 'skipNodeStart',
+    name: 'skip-node-start',
+    definition: {
+      describe: 'Skip starting consensus nodes after staging a network upgrade',
```
jeromy-cannon: description references "a network upgrade" which is command-specific. Flags should be generic across the CLI. Checklist §4.
```suggestion
        describe: 'Skip starting consensus nodes after staging a freeze upgrade',
```

.github/workflows/script/launch_network.sh
```diff
@@ -551,149 +649,306 @@ fi
...
+
+# Step 1: Upgrade BN while CN source version is running, unless the BN bypass is re-enabled.
+ACTIVE_BLOCK_NODE_VERSION="${PREV_BLOCK_VERSION_NO_V}"
+if [[ "${SKIP_BLOCK_NODE_UPGRADE_UNTIL_BN_3150_FIXED}" == "true" ]]; then
+  echo "$(date '+%Y-%m-%d %H:%M:%S') - TEMPORARILY skipping BN upgrade to ${CURRENT_BLOCK_VERSION}"
+  echo "Reason: hiero-block-node#3150 can make mirror importer reject live-stream batches starting with ROUND_HEADER."
+  echo "Block node remains on ${PREV_BLOCK_VERSION_NO_V}; continuing component migration coverage."
+  dump_bn_log "BN upgrade skipped due to hiero-block-node#3150"
+elif [[ "${PREV_BLOCK_VERSION_NO_V}" != "${CURRENT_BLOCK_VERSION}" ]]; then
+  TEMP_BN_UPGRADE_VALUES_FILE="$(mktemp -t bn-upgrade-values-XXXX.yaml)"
```
jeromy-cannon, line 738: `TEMP_BN_UPGRADE_VALUES_FILE` is created here but is not in the `on_exit` trap's cleanup list. If the `block node upgrade` command below exits non-zero, `set -e` fires before the inline `rm -f` and the file leaks. For consistency with the other temp vars, initialise it to `""` at the top of the script and add a guard in `on_exit`.

src/commands/backup-restore.ts
JeffreyDallas: Done — created `BackupDatabaseDumpNotFoundSoloError` (SOLO-4080) in `src/core/errors/classes/validation/backup-database-dump-not-found-solo-error.ts`, registered it in `ErrorCodeRegistry` and `SoloErrors.validation`, and replaced the throw site.

.github/workflows/script/launch_network.sh
JeffreyDallas: Done — pinned to `busybox:1.36.1` at both sites (lines 807 and 831).

.github/workflows/script/launch_network.sh
JeffreyDallas: Fixed — pinned to `busybox:1.36.1`.

src/commands/block-node.ts
JeffreyDallas: Done — pinned to `'busybox:1.36.1'`.

src/core/profile-manager.ts
JeffreyDallas: Done — extracted as `Helpers.hasMultipleKubernetesContexts()` in `src/core/helpers.ts`. Both `NodeCommandTasks` and `ProfileManager` now call the shared method; the private copies have been removed.

src/commands/block-node.ts
JeffreyDallas: Done — extracted `Helpers.requiresRsaBootstrap(consensusNodeVersion, streamMode)` to `src/core/helpers.ts`. `shouldConfigureRsaMirrorBootstrapSource()`, `shouldInjectRsaBootstrapValuesFile()`, and `shouldSkipMinioSetup()` all delegate to it for the version+mode check.

src/commands/node/handlers.ts
JeffreyDallas: Done — added `freezeBlockDrainSeconds` to `FREEZE_FLAGS`, `NodeFreezeConfigClass`, and `freezeConfigBuilder`. The hardcoded `FREEZE_BLOCK_STREAM_DRAIN_MILLISECONDS` constant is removed; the `freeze()` handler now calls `drainBlockStreamAfterFreeze()` which reads `context_.config.freezeBlockDrainSeconds` (default 20 s from the flag).

src/commands/flags.ts
JeffreyDallas: Done — updated to `'Skip starting consensus nodes after staging a freeze upgrade'`.

.github/workflows/script/launch_network.sh
JeffreyDallas: Done — initialized `TEMP_BN_UPGRADE_VALUES_FILE=""` at the top of the script and added a guard block to the `on_exit` trap.

.github/workflows/script/launch_network.sh
```diff
@@ -147,6 +158,65 @@ show_service_ips() {
   kubectl get svc -n "${namespace}" network-node1-svc network-node2-svc -o custom-columns=NAME:.metadata.name,CLUSTER-IP:.spec.clusterIP,CREATED:.metadata.creationTimestamp
 }
 
+get_latest_mirror_block_number() {
```
jeromy-cannon: duplicate code fragment — `get_latest_mirror_block_number` and `wait_for_mirror_block_progress` (and the `wait_for_mirror_block_count_progress` wrapper below) are defined again in `solo_smoke_test.sh`. Both scripts already `source "${SCRIPT_DIR}/helper.sh"`. Could make more DRY by moving these into `helper.sh` and calling from both places. (checklist §1)

.github/workflows/script/solo_smoke_test.sh
```diff
@@ -142,16 +145,72 @@ EOF
   done
 }
 
+function get_latest_mirror_block_number ()
```
jeromy-cannon: duplicate of the functions just added to `launch_network.sh` — `get_latest_mirror_block_number` and `wait_for_mirror_block_progress` appear verbatim in both files. These should live in `helper.sh`, which both scripts already source.

src/commands/node/tasks.ts
```diff
@@ -196,6 +202,127 @@ const {gray, cyan, red, green, yellow} = chalk;
...
+      port: address.port,
+      ipAddressV4: parseIpAddressToUint8Array(address.ipAddressV4),
+    });
+  }
+
+  private static shouldAvoidGossipFqdn(consensusNodes: ConsensusNode[], gossipFqdnRestricted: boolean): boolean {
+    return gossipFqdnRestricted || Helpers.hasMultipleKubernetesContexts(consensusNodes);
+  }
+
+  private static buildRsaAddressBookHistory(consensusNodes: ConsensusNode[], keysDirectory: string): string {
```
jeromy-cannon: this RSA address-book JSON build (`{addressBooks:[{addressBook:{nodeAddress:[...]},startBlock:'0',endBlock:'-1'}]}`) is constructed again inline in `BlockNodeCommand.writeRsaBootstrapInitContainerValuesFile`. Consider having one call the other so the JSON shape stays in sync. (checklist §1)

src/commands/block-node.ts
```diff
@@ -407,6 +425,93 @@ export class BlockNodeCommand extends BaseCommand {
...
+    return Helpers.requiresRsaBootstrap(consensusNodeVersion, blockStreamMode);
+  }
+
+  private resolveMirrorNodeReleaseName(): string {
+    const mirrorNodeId: number =
+      this.remoteConfig.configuration.state.mirrorNodes?.[0]?.metadata.id ?? BlockNodeCommand.DEFAULT_MIRROR_NODE_ID;
+    return Templates.renderMirrorNodeName(mirrorNodeId);
+  }
+
+  private writeRsaBootstrapInitContainerValuesFile(
```
jeromy-cannon, line 440: this RSA address-book building duplicates `NodeCommandTasks.buildRsaAddressBookHistory` in `tasks.ts`. Both iterate consensus nodes, read the gossip PEM, export SPKI DER, and produce the same JSON shape. One should call the other. (checklist §1)

src/core/helpers.ts
```diff
@@ -111,12 +118,54 @@ export class Helpers {
...
+    streamMode: string,
+    writerMode: string = constants.BLOCK_STREAM_WRITER_MODE,
+  ): void {
+    Helpers.upsertApplicationProperty(lines, 'blockStream.streamMode', streamMode);
+    Helpers.upsertApplicationProperty(lines, 'blockStream.writerMode', writerMode);
+
+    if (streamMode === 'BLOCKS') {
+      Helpers.upsertApplicationProperty(lines, 'blockStream.streamWrappedRecordBlocks', 'false');
+    } else if (streamMode === 'BOTH') {
+      Helpers.upsertApplicationProperty(lines, 'blockStream.streamWrappedRecordBlocks', 'true');
```
jeromy-cannon: for BOTH mode, this sets `streamWrappedRecordBlocks=true`, but every call site then calls `ensureWrappedRecordBlocksDisabled(lines, 'BOTH')` which immediately overwrites it to `false`. The `true` here appears to be a dead write. Should this branch set `false` directly to match the documented intent (BN/mirror rejects ROUND_HEADER batches starting with `ROUND_HEADER`)? Or is there a code path where `ensureWrappedRecordBlocksDisabled` is not called?

src/commands/node/tasks.ts
```diff
@@ -4307,6 +4710,16 @@ export class NodeCommandTasks {
     };
   }
 
+  public drainBlockStreamAfterFreeze(): SoloListrTask<NodeUpgradeContext> {
```
jeromy-cannon, line 4690: it seems like this is a workaround — CN does not signal when the final block has been flushed to the block node after a freeze, so Solo sleeps instead. This pause will be invisible to SREs debugging a slow upgrade. Is there an upstream CN issue for a deterministic "final block flushed" event so this sleep can be replaced by a real check? (checklist §13)

src/commands/node/tasks.ts
```diff
@@ -3524,22 +3794,155 @@ export class NodeCommandTasks {
...
+
+      await blockNodeContainer.execContainer(['mkdir', '-p', NodeCommandTasks.BLOCK_NODE_APPLICATION_STATE_DIRECTORY]);
+      await blockNodeContainer.copyTo(bootstrapFilePath, NodeCommandTasks.BLOCK_NODE_APPLICATION_STATE_DIRECTORY);
+      await blockNodeContainer.execContainer([
+        'test',
+        '-r',
+        `${NodeCommandTasks.BLOCK_NODE_APPLICATION_STATE_DIRECTORY}/${NodeCommandTasks.BLOCK_NODE_RSA_BOOTSTRAP_FILE}`,
+      ]);
+
+      await k8.pods().delete(podReference);
```
jeromy-cannon, line 3900: deleting block-node pods as a side effect of a key-copy step is a significant operational action. The parent Listr task is titled "Copy node keys to secrets"; the nested title says "Refresh block node RSA bootstrap state" — but the actual effect (pod deletion + wait for readiness) won't be obvious to an operator watching progress. Consider pulling this into a standalone top-level task so the restart is clearly visible, or at least adding the pod name to the nested task title.

src/core/errors/error-code-registry.ts
```diff
@@ -213,6 +213,7 @@ export const ErrorCodeRegistry: Record<string, string> = {
   BACKUP_INPUT_PATH_NOT_FOUND: 'SOLO-4073',
   BACKUP_INPUT_MUST_BE_ZIP: 'SOLO-4074',
   BACKUP_NO_LOG_FILES: 'SOLO-4075',
+  BACKUP_DATABASE_DUMP_NOT_FOUND: 'SOLO-4080',
```
jeromy-cannon: surrounding codes are SOLO-4075 → SOLO-4076 with no SOLO-4079 visible. Was 4079 intentionally reserved, or should this be `SOLO-4079`?

src/core/constants.ts
```diff
@@ -226,7 +226,7 @@ export const SOLO_MIRROR_WEB3_NAME_LABEL: string = 'app.kubernetes.io/name=web3'
 export const SOLO_MIRROR_POSTGRES_NAME_LABEL: string = 'app.kubernetes.io/name=postgres';
 export const SOLO_MIRROR_REDIS_NAME_LABEL: string = 'app.kubernetes.io/name=redis';
 export const SOLO_MIRROR_RESTJAVA_NAME_LABEL: string = 'app.kubernetes.io/name=restjava';
-export const SOLO_BLOCK_NODE_NAME_LABEL: string = 'app.kubernetes.io/name=block-node-1';
+export const SOLO_BLOCK_NODE_NAME_LABEL: string = 'block-node.hiero.com/type=block-node';
```
jeromy-cannon, line 229: changing the selector from `app.kubernetes.io/name=block-node-1` to `block-node.hiero.com/type=block-node` means any live cluster whose block-node pods carry only the old label will be invisible to Solo after upgrade. Is there a fallback selector or migration note for existing deployments that were created with the previous chart?

src/commands/node/tasks.ts
JeffreyDallas: The 20-second sleep is a workaround for CN not signaling when the final block has been flushed after a freeze. There is a CN upstream issue tracking this: https://github.com/hiero-ledger/hiero-consensus-node/issues/26498. Once CN provides a reliable signal, this sleep can be replaced with an explicit poll.

src/core/constants.ts
JeffreyDallas: The label key changed from `app.kubernetes.io/name=block-node-1` to `block-node.hiero.com/type=block-node` to match the new label that the BN Helm chart (v0.38+) sets on all block-node pods regardless of instance ID. The old selector would only match the first block-node pod and would break with multiple block nodes. Existing deployments that upgrade to this version of Solo must also upgrade the BN chart to v0.38+ (which the CN v0.75 upgrade does via the BN chart bump in this PR).

.github/workflows/script/launch_network.sh
JeffreyDallas: Fixed. All three functions (`get_latest_mirror_block_number`, `wait_for_mirror_block_progress`, `wait_for_mirror_block_count_progress`) have been moved to `helper.sh`, which both scripts already source. The 4-param variant in `launch_network.sh` has been replaced with the unified 5-param version (adding `required_new_blocks`). The `launch_network.sh` call site at `wait_for_mirror_block_progress "source deployment after one-shot" -1 90 2` has been updated to the new signature with an explicit `required_new_blocks=1`.

.github/workflows/script/solo_smoke_test.sh
JeffreyDallas: Fixed. Both functions removed from `solo_smoke_test.sh` — they now come from `helper.sh` via the existing `source` at the top of the script. The solo_smoke_test.sh call at line 467 uses the 5-param signature unchanged (`required_new_blocks=${SMOKE_MIRROR_BLOCK_SETTLE_BLOCKS}`).

src/commands/node/tasks.ts
JeffreyDallas: Fixed. Extracted to `Helpers.buildRsaAddressBookJson(consensusNodes, keysDirectory): string` in `helpers.ts`. `NodeCommandTasks.buildRsaAddressBookHistory` has been removed and the call site updated to use `Helpers.buildRsaAddressBookJson`. `block-node.ts` now does its existence checks, then calls `Helpers.buildRsaAddressBookJson` to build the JSON rather than duplicating the build logic inline.

src/commands/block-node.ts
JeffreyDallas: Fixed — see the reply to the related comment on tasks.ts. `writeRsaBootstrapInitContainerValuesFile` now checks that all key files exist (returning `undefined` if any are missing), then delegates to `Helpers.buildRsaAddressBookJson` for the actual JSON construction.

src/core/helpers.ts
JeffreyDallas: Fixed. The `else if (streamMode === 'BOTH')` branch that sets `streamWrappedRecordBlocks=true` has been removed. Every call site already calls `ensureWrappedRecordBlocksDisabled` afterwards which unconditionally sets it back to `false`, so the `true` write was always dead.

src/commands/node/tasks.ts
JeffreyDallas: Fixed. Both task titles now read `'Refresh block node RSA bootstrap state (restarts block-node pods)'` so the pod restart is visible to operators watching the task list.

src/core/errors/error-code-registry.ts
JeffreyDallas: Fixed. `BACKUP_DATABASE_DUMP_NOT_FOUND: 'SOLO-4080'` has been moved to after `VALUES_FILE_PARSE_FAILED: 'SOLO-4079'` so the codes now appear sequentially in file order (4075 → 4076 → 4077 → 4078 → 4079 → 4080).

jeromy-cannon (approved):

Issue #860, opened 2024-11-21
Update Solo for Network transplant process (node-overrides.yaml)


derektriley:
### Requested Feature

**Background**
In order for Solo to start/spin up networks with a saved state generated from a different network it will need to support placing the correct files on disk in the correct location. 

- An `override-network.json` file will be looked for by the consensus node in `NetworkAdminConfig::upgradeSysFilesLoc` (default of `data/config`).
- If the above file is not found, then the consensus node will look for a scoped override network file from a subdirectory named after the round number (for example `<NetworkAdminConfig::upgradeSysFilesLoc>/123/override-network.json`).
- The [V057RosterSchema.java](https://github.com/hashgraph/hedera-services/pull/16594/files#diff-952eb1264ac5e82c11422564f00af9d7b6e7752d30b86c954f97ee510d56482fR81) in hedera-services provides a restart-only schema which sets the appropriate active roster for genesis, override, or upgrade). It may look for the override file for the round number specified in the migration context

See this [PR](https://github.com/hashgraph/hedera-services/pull/16594) in hedera-services for more implementation details if necessary.

**Acceptance Criteria**
- Solo has been updated to support putting the `override-network.json` file(s) in the correct location for networks it spins up.
- Test to verify this process is sucessful.

**Dependencies**
- None

### Business Need

For Solo to support the network transplant process, which has been recently updated on the consensus node.

### Linked Issues
- required by: https://github.com/hiero-ledger/solo/issues/2379

jeromy-cannon:
some additional info to help:
Per Dan: hey Jeromy — circling back to above we did get some answers, there’s some hard coded samples and then some code to generate them (albeit in python)
https://github.com/hashgraph/infrastructure/issues/7487
https://github.com/hashgraph/infrastructure/pull/7610
also a few comments in #7487 on the data format and how to generate it, in general the file is named override-network.json - i kept calling it network-override.json and nothing would work :upside_down_face: belongs in data/config

JeffreyDallas:
@jeromy-cannon and @derektriley 

Need clarification how to get / generate `override-network.json`

I did the following test steps:
1. start a solo network and run some account create transactions
2. stop the network and save state file, and the original `genesis-network.json`

3. start a new solo network from scratch, copy state file to pod, 
4. copy `genesis-network.json` to data/config as `override-network.json`
5. start network, but failed with exeception

```
[...]

PR #5375, opened 2026-07-28
fix: support state save and restore redeploy workflow

### LINKED ISSUE #2379: Enhance Solo to support starting the network from an existing state with transplant
This is needed for CITR as well as other uses case.  This would also help users for using Solo for various test cases.  This might be just validation once #860 is completed.  As this command might do what we need:
```
solo node start --deployment ${SOLO_DEPLOYMENT} --state-file ${STATE_FILE_ZIP}
```
 
- depends on: https://github.com/hiero-ledger/solo/issues/860

NOTE: may be able to leverage some of Tom's solutions that allows him to start from an existing state

JeffreyDallas:
## Description

This pull request makes the `state-save-and-restore` workflow safe for restoring a frozen consensus-node state into a fresh cluster with different network endpoints.

It:

* adds a stable-state wait before downloading state (`resources/wait-for-stable-saved-state.sh` plus an in-pod poll), requiring the saved-state files to remain unchanged across consecutive checks, and fails when a node is neither frozen nor stopped
* updates the freeze workflow to wait for the signed frozen state and to drain the block stream to the block node before stopping nodes and capturing state
* normalizes every node's downloaded archive to one common, fully signed round so all nodes restore from the same state/PCES boundary
* adds `src/core/pces-trimmer.ts`, which trims preconsensus events past the selected round so a restored snapshot does not replay events created after it (for example a freeze transaction ordered moments later, which would immediately re-freeze the restored node); `backup-restore` uses it because that flow resumes the network as a live deployment
* makes `consensus node start` wait for whichever terminal status a restore actually reaches, ACTIVE or FREEZE_COMPLETE, and run the ACTIVE-only follow-up work (port forwarding, TSS, gRPC web endpoint, stakes, started event) only when the nodes came up active
* adds `generate-override-network.mjs`, which builds `override-network.json` from the saved network definition and the fresh Kubernetes service endpoints, then copies it into the consensus-node pods
* adds the fresh-cluster restore and verification workflow to the example, including checks that both nodes reach `FREEZE_COMPLETE` and no `Invalid State Signature` is reported
* narrows the example to consensus nodes plus a block node, removing its external PostgreSQL/mirror-node path (`scripts/init.sh`). External-database coverage stays in `examples/external-database-test` and the `test-e2e-external-database` E2E job
* adds three tunables to `src/core/constants.ts` for the saved-state stability poll (`STATE_DOWNLOAD_STABLE_MAX_ATTEMPTS`, `STATE_DOWNLOAD_STABLE_DELAY`, `STATE_DOWNLOAD_STABLE_POLLS_REQUIRED`)
* documents the complete workflow in the example README
* restores `cluster-ref config setup` to the example's `init-solo`, which this branch had dropped. It installs the cluster-level MinIO operator and its CRDs; the restored network does not need them (its block node supplies stream storage, so solo renders it with `cloud.minio.enabled=false`) but the transplant target has no block node, so solo enables MinIO and the chart renders a `minio.min.io/v2` Tenant that cannot be created without them
* repairs the example after the merge from `main`: the merge grafted main's `destroy-database` name onto this branch's `destroy-network` body, leaving `restore` and `destroy` pointing at a task that no longer existed, and pulled the README back to describing the removed PostgreSQL/mirror-node workflow (including a step-by-step block that skipped the now-mandatory freeze)

Behavior changes worth reviewer attention:

* `consensus state download` now fails unless the node is `FROZEN` or `STOPPED`. Previously it would archive a running node's state, which can be captured mid-write and is not safely restorable.
* `consensus node start --state-file` no longer assumes the restore lands in ACTIVE. It accepts ACTIVE or FREEZE_COMPLETE and adapts, so a restore that replays back into a freeze no longer waits out the full ACTIVE timeout, and one that replays into ACTIVE still gets the full active-start treatment.

### Related Issues

* Closes #2379
* Related to hiero-ledger/hiero-consensus-node#26950 — upstream request for an explicit "round N is fully signed and safe to export" signal, which the in-pod filesystem poll added here works around

* [ ] This PR has no breaking changes

The "no breaking changes" box is intentionally unchecked: `consensus state download` now errors for a running node where it previously produced an archive. That guard is deliberate, since such an archive can be mid-write, but it is a user-visible behavior change.

### Testing

* [x] This PR added integration/end-to-end tests

The following automated tests were added:

* `test/unit/core/pces-trimmer.test.ts` — 9 cases covering a file below the truncation cutoff, a file straddling it, a record ending exactly at the buffer boundary, an unknown version header, and the deletion path
* `test/unit/core/network-nodes.test.ts` — state archive round normalization to one common signed round, plus platform status name mapping
* End-to-end coverage is the rewritten `examples/state-save-and-restore` workflow, which runs in the nightly Test Examples matrix (it is not a PR-gating job)

* Full local run of `examples/state-save-and-restore` (`task`) on a fresh kind cluster, arm64, exercising this branch end to end: freeze ran both the restored block-stream drain (20s) and the stable-state wait (5s); `consensus state download` normalised both archives to common signed round 1442; the cluster was deleted and recreated; `override-network.json` was generated with 6 rewritten endpoints and staged at round 1443; and the restore reported `node1=FREEZE_COMPLETE, node2=FREEZE_COMPLETE` in 5s followed by `State verification complete - both restored nodes are FREEZE_COMPLETE with no ISS`. The non-restore start path was confirmed unaffected (`Check all nodes are ACTIVE or FREEZE_COMPLETE [SKIPPED: No state file supplied]`, ACTIVE group running, and each of the previously duplicated tasks executing exactly once).

* Ran the complete `examples/state-save-and-restore` workflow on a clean kind cluster, including freeze, stable state capture, cluster recreation, endpoint override generation/copy, and restore.
* Verified both restored consensus nodes reached `FREEZE_COMPLETE`, both proxies became active, and no `Invalid State Signature` was reported.
* Ran `task verify-state` against the restored cluster; both nodes reported platform status `6.0`.
* Ran TypeScript compilation, shell syntax checks, reference checks, and `git diff --check`.

The following was not tested:

* The State Save and Restore example failed on commit `0a0b65fb1` ([run 34519290147](https://github.com/hiero-ledger/solo/actions/runs/34519290147/job/103012638711)), where the restore verification was chosen by reading `FREEZE_STATE` off the archive's selected round. That premise was wrong: the selected round is an ordinary signed round, but the retained preconsensus events still carry the freeze transaction, so the nodes went `REPLAYING_EVENTS -> FREEZE_COMPLETE` while the archive said otherwise. Commit `b959816d6` replaces the prediction with an observed-status wait. The later merge from `main` then broke every entry point of the example outright (`task: Task "destroy-network" does not exist`), repaired in `4a54163c2`. Both have since been validated by a full local run of the example on a kind cluster (see Testing below); **they have still not run through Test Examples**, which is nightly/dispatch-only and never runs on a pull request; the repair is verified only by `task <name> --dry` resolving for `default`, `restore`, `destroy`, `transplant`, `verify-state`, `save-state` and `stop-network`, plus a check that every `task:` reference and every task named in the README resolves.
* The example's `transplant` tasks and the `--transplant` flag come from #5839 on `main`, not from this branch. The interaction with the restore changes has now been measured rather than assumed: a transplant settles in **ACTIVE** (`node1=ACTIVE, node2=ACTIVE` in 14s), so `checkAllNodesAreActiveOrFrozen` routes it down the same active path it took on `main` and the frozen-only tasks skip. The transplant phase then fails for a pre-existing reason unrelated to this branch — `setGrpcWebEndpoint` signs a node update with the target deployment's admin key while the restored state carries the source network's node records — tracked in #5996.
* No PR-gating job exercises `consensus network freeze` with a block node present: the only E2E that freezes (`separate-node-add`) has no block node, and `block-node.test.ts` never freezes. The combination is covered only by the nightly example run.
* The saved-state stability poll's timeout path and the shell script's round-selection branches are exercised through the example workflow rather than by unit tests.

<details>
<summary>
Commit message guidelines
</summary>

</details>


jan-milenkov (dismissed):

examples/state-save-and-restore/Taskfile.yml
jeromy-cannon, line 1: this https://github.com/hiero-ledger/solo/issues/2379 says it is dependent on https://github.com/hiero-ledger/solo/issues/860.  I'm not seeing the changes that enable the override-network.json logic to change the IP address or FQDN versus what was the network was backed up with.

JeffreyDallas: test updated with build new override-network.json then copy to pod

src/core/network-nodes.ts
```diff
@@ -167,6 +336,15 @@ export class NetworkNodes {
 
       const k8: K8 = this.k8Factory.getK8(context);
       const zipFileName: string = `${HEDERA_HAPI_PATH}/${podReference.name}-state.zip`;
+      // A frozen node should yield a freeze round; a merely stopped node may only
+      // have a non-freeze signed round available.
+      const requireFreezeRound: boolean = deploymentPhase === DeploymentPhase.FROZEN;
```
Fantomasa, line 361: requireFreezeRound is never actually enforced - its value only changes log and error text. Is this the behavior that we expect?

JeffreyDallas: It is used in line 400 by script `resources/wait-for-stable-saved-state.sh`

Fantomasa (dismissed):
LGTM

ddeskov-limechain (dismissed):
LGTM

examples/state-save-and-restore/scripts/generate-override-network.mjs
```diff
@@ -0,0 +1,173 @@
...
+      throw new Error(
+        `source network only has ${String(sourceNetwork.nodeMetadata.length)} nodeMetadata entries for ${String(nodeAliases.length)} aliases`,
+      );
+    }
+
+    const rewrittenNetwork = structuredClone(sourceNetwork);
+    let changedEndpointCount = 0;
+
+    for (const [nodeIndex, nodeAlias] of nodeAliases.entries()) {
+      const serviceName = `network-${nodeAlias}-svc`;
```
jan-milenkov, line 38: I think on `Templates` we have a method for it, use it instead of manually building the string

JeffreyDallas: This is a mjs file under example directory, examples supposed to be self contained, should not use any typescript source.
(unzip as individual project directory to run themselves)

src/core/pces-trimmer.ts
```diff
@@ -0,0 +1,222 @@
...
+ * A restored state snapshot at round N is already a complete, fully signed state; any
+ * preconsensus event with a birth round greater than N was created after that snapshot and
+ * should not be replayed when the snapshot is restored elsewhere (for example, replaying a
+ * freeze transaction ordered moments after the snapshot was taken would immediately re-freeze
+ * the restored node). This trims strictly at a record boundary, so no footer/checksum needs to
+ * be regenerated (see the PCES file format used by hiero-consensus-node's PcesFileIterator,
+ * which already tolerates a file ending mid-record the same way it tolerates an abrupt
+ * shutdown).
+ */
+export class PcesTrimmer {
```
jeromy-cannon, line 32: This whole class hand-decodes the PCES protobuf wire format (varints, field numbers, wire types) to find/truncate events by birth round. That's a lot of trust in an internal, unversioned-from-Solo's-perspective binary layout, and it has no unit tests despite being the riskiest new code in this PR (a subtle off-by-one in `findTruncationOffset`/`readVarint` wouldn't throw — it would just silently trim at the wrong boundary). Please add unit tests covering: a file below the truncation cutoff, a file straddling it, a record ending exactly at the buffer boundary, a wrong/unknown version header, and the `truncateAtOffset <= 4` deletion path.

Also worth a question: should trimming-to-round live in hiero-consensus-node itself (as a supported export/trim capability) rather than Solo re-implementing a slice of its wire format?

src/core/network-nodes.ts
```diff
@@ -187,6 +385,89 @@ export class NetworkNodes {
     this.logger.debug(`getNodeState(${pod.podReference.name.name}): ...end`);
   }
 
+  private async waitForStableSavedState(
```
jeromy-cannon, line 388: This copies a script into the pod, chmods it, and polls by hashing the saved-state tree to detect when writes have stopped. It seems like a workaround for the platform not exposing an explicit "state is stable/exportable" signal. Are there other options (a status the consensus node already surfaces, e.g. tied to `FREEZE_COMPLETE` or a per-round signed event, or a metrics value from the metrics port) that would avoid the filesystem-fingerprint heuristic and the up-to-6-minute polling loop entirely?

resources/wait-for-stable-saved-state.sh
```diff
@@ -0,0 +1,85 @@
...
+fi
+
+if [[ "${prefer_freeze_round}" == "true" && "${selected_kind}" != "freeze" ]]; then
+  selected_kind="frozen-fallback"
+fi
+
+# Fingerprint the entire saved-state tree, not just the chosen round directory,
+# so the caller can detect when background flushes have stopped changing disk
+# contents across consecutive polls.
+find "${saved_dir}" -type f -print0 \
```
jeromy-cannon:
```suggestion
find "${saved_dir}" -type f -printf '%s %T@ %p\n' \
```
Hashing the full contents of every file under `data/saved` on every 2s poll (up to 180 times) could be expensive for a real state size and adds repeated I/O load exactly while the node is trying to quiesce. Consider fingerprinting `size + mtime` per file (cheap `stat`, as sketched above) instead of re-hashing file contents each pass, unless mtimes aren't reliably updated on write in this environment.

examples/state-save-and-restore/Taskfile.yml
```diff
@@ -255,44 +240,105 @@ tasks:
             fi
 
             cp "${SRC_STATE_FILE_PATH}" "${DEST_STATE_FILE_PATH}"
-            echo "Prepared state for ${node}: ${DEST_STATE_FILE_PATH}"
+            SELECTED_ROUND="$(unzip -Z1 "${DEST_STATE_FILE_PATH}" | awk -F/ '$1 == "com.hedera.services.ServicesMain" && $4 ~ /^[0-9]+$/ {print $4}' | sort -n | tail -n 1)"
```
jeromy-cannon: This re-derives the saved-state round by parsing the zip listing with `awk`, duplicating the `com.hedera.services.ServicesMain/<node>/<realm-shard>/<round>` layout that `NetworkNodes.findSavedStateRoundRoot` already encodes in TypeScript (`src/core/network-nodes.ts:285`). `normalizeDownloadedStateArchives` already computes and returns/logs the selected round — could the example consume that instead of re-parsing the archive itself, so there's one source of truth for the layout?

src/commands/node/tasks.ts
```diff
@@ -3304,12 +3333,48 @@ export class NodeCommandTasks {
...
+          const nodeComponent: ConsensusNodeStateSchema = this.remoteConfig.configuration.components.getComponent(
+            ComponentTypes.ConsensusNode,
+            Templates.renderComponentIdFromNodeAlias(nodeAlias),
+          );
+          const deploymentPhase: DeploymentPhase = nodeComponent.metadata.phase;
+
+          if (![DeploymentPhase.FROZEN, DeploymentPhase.STOPPED].includes(deploymentPhase)) {
+            this.logger.showUser(
+              chalk.yellow(
+                `Warning: node ${nodeAlias} is in phase '${deploymentPhase}'. State download is only supported when consensus nodes are frozen or stopped.`,
```
jeromy-cannon: This warning is immediately followed by throwing `SoloErrors.validation.illegalArgument` with essentially the same message a couple lines down — the separate `showUser` warning looks redundant since the thrown error's message will be shown to the user anyway.
```suggestion
```

examples/state-save-and-restore/Taskfile.yml
```diff
@@ -255,44 +240,105 @@ tasks:
...
-      - cmd: sleep 30  # Wait for nodes to stabilize
       - cmd: |
-          echo "Generating test transactions to verify network functionality..."
-          for i in {1..3}; do
-            $SOLO_COMMAND ledger account create --deployment {{ .DEPLOYMENT }}
-            sleep 2
+          echo "Verifying restored frozen state..."
+          for node in $(echo {{ .NODE_ALIASES }} | tr ',' ' '); do
+            kubectl exec network-${node}-0 -n {{ .NAMESPACE }} -c root-container -- \
+              sh -c 'curl -sf http://localhost:9999/metrics | grep platform_PlatformStatus | grep -v "^#" | grep -q " 6\\.0$"'
```
jeromy-cannon, line 334: Magic number `6.0` for `FREEZE_COMPLETE` platform status — a short comment naming the enum value would save the next reader a lookup.

resources/wait-for-stable-saved-state.sh
JeffreyDallas: Updated

src/commands/node/tasks.ts
JeffreyDallas: Removed

examples/state-save-and-restore/Taskfile.yml
JeffreyDallas: Added in aaabc8fa1: `# platform_PlatformStatus value 6 == FREEZE_COMPLETE (see NodeStatusCodes in src/core/enumerations.ts)` right above the check.

src/core/pces-trimmer.ts
JeffreyDallas: Added a dedicated unit test file (test/unit/core/pces-trimmer.test.ts, 9 cases).

On the design question: I agree in principle this would be cleaner living in hiero-consensus-node as a supported "export/trim to round" capability — `PcesGraphSlicer`/`pcli slice` already exists there but always rewrites birth rounds and re-signs/re-keys, which is overkill and inappropriate for a production restore (new identities). A pure truncate-preserving-birth-rounds tool doesn't currently exist upstream. I don't have visibility into whether the consensus-node team would want to own this, so I kept the fix scoped to Solo for now rather than block this PR on an upstream ask — happy to file that as a follow-up issue if that seems like the right direction.

src/core/network-nodes.ts
JeffreyDallas: No code change here, just answering the question: yes, this is a workaround, and I don't think there's a single existing signal that covers all the cases this needs to handle. `FREEZE_COMPLETE` only covers the frozen path — this same wait is also used for a plain `STOPPED` deployment (`consensus node stop` without a freeze), where the round we want is "the newest fully signed round" with no freeze-specific status to key off. A metrics value would need to expose per-round signing-weight progress (i.e. roughly what `stateMetadata.txt`'s SIGNING_WEIGHT_SUM/TOTAL_WEIGHT already gives us on disk) rather than a single platform status, since the platform's own status transitions (ACTIVE → FREEZING → FREEZE_COMPLETE, or ACTIVE → OBSERVING/CHECKING when stopped and restarted) don't tell the caller whether *this specific round's* signatures have finished landing on disk. If the consensus-node metrics endpoint already exposes something equivalent to signing-weight-per-round that I'm missing, I'd rather use that — but the filesystem check was the only thing I found that reliably answers "is the round I'm about to zip up actually done changing."

JeffreyDallas: Filed hiero-ledger/hiero-consensus-node#26950 to track this upstream — no existing issue covered this exact gap (checked ~25 search variations against both repos). Linked back to this PR/thread for context.

src/commands/node/handlers.ts
```diff
@@ -1252,7 +1260,7 @@ export class NodeCommandHandlers extends CommandHandler {
         this.tasks.identifyExistingNodes(),
         this.tasks.sendFreezeTransaction(),
         this.tasks.checkAllNodesAreFrozen('existingNodeAliases'),
-        this.tasks.drainBlockStreamAfterFreeze<NodeFreezeContext>(),
+        this.tasks.waitForFrozenStateToBeStable('existingNodeAliases'),
```
Fantomasa, line 1315: Really like the stable-state wait here! One question though: it replaces `drainBlockStreamAfterFreeze` rather than running alongside it, and the wait only watches `data/saved`, so the block stream may not have fully reached the block node before the nodes stop (the upgrade flow still drains). Was dropping the drain intentional?

src/commands/node/handlers.ts
```diff
@@ -1168,35 +1167,44 @@ export class NodeCommandHandlers extends CommandHandler {
...
+      // Must precede checkNodesAndProxiesAreActive: when --debug-node-alias is set the JVM starts
+      // with suspend=y and will never reach ACTIVE until a debugger connects via this port-forward.
+      this.tasks.enableDebuggerPortForwarding(),
+    ];
+
+    if (restoringState) {
+      // A freeze-captured archive is expected to restore into FREEZE_COMPLETE.
+      // Do not require ACTIVE or run ACTIVE-only TSS/start-event tasks for it.
+      startTasks.push(
+        this.tasks.checkAllNodesAreFrozen('nodeAliases'),
```
Fantomasa: One scenario I'm curious about: this branch keys off `--state-file` presence, but state download also happily accepts STOPPED nodes, and those archives normalize to a non-freeze round that replays back to ACTIVE. Starting such a state would poll for FREEZE_COMPLETE until timeout. Is stopped-state restore intentionally out of scope for now, or would keying off the archive's selected round kind be safer?

resources/wait-for-stable-saved-state.sh
```diff
@@ -0,0 +1,87 @@
...
+# states (it reports SIGNING_WEIGHT_SUM: 0 for them).
+prefer_freeze_round="${1:-false}"
+saved_dir="${2:-/opt/hgcapp/services-hedera/HapiApp2.0/data/saved}"
+
+if command -v sha256sum >/dev/null 2>&1; then
+  hash_cmd=(sha256sum)
+elif command -v shasum >/dev/null 2>&1; then
+  hash_cmd=(shasum -a 256)
+elif command -v openssl >/dev/null 2>&1; then
+  hash_cmd=(openssl dgst -sha256)
```
Fantomasa:
```suggestion
  hash_cmd=(openssl dgst -sha256 -r)
```
Sneaky one: without `-r`, openssl prints `SHA2-256(stdin)= <hash>`, so the awk at the bottom grabs the constant label as the fingerprint and the stability check would pass even while files are still changing. Luckily the consensus containers all have sha256sum, so this is just future-proofing the fallback.

src/core/constants.ts
```diff
@@ -526,6 +526,13 @@ export const NETWORK_NODE_GRPC_READINESS_DELAY: number =
 export const NETWORK_NODE_GRPC_READINESS_REQUIRED_SUCCESSES: number =
   +getEnvironmentVariable('NETWORK_NODE_GRPC_READINESS_REQUIRED_SUCCESSES') || 3;
 
+// Saved State Stability Checks
+export const STATE_DOWNLOAD_STABLE_MAX_ATTEMPTS: number =
```
Fantomasa, line 530: Small docs follow-up: these three new env vars would be great to have on the solo-docs env page (env.md moved out of this repo). Mind linking a companion docs PR?

src/core/network-nodes.ts
```diff
@@ -149,12 +157,193 @@ export class NetworkNodes {
...
+
+  /**
+   * Wait for a fully signed freeze state before a freeze workflow stops the node.
+   * A FROZEN platform status alone is not enough: stopping immediately can leave
+   * the archive with only a non-freeze state and misaligned PCES replay data.
+   */
+  public async waitForFrozenStateToBeStable(podReference: PodReference, context?: string): Promise<void> {
+    const containerReference: ContainerReference = ContainerReference.of(podReference, ROOT_CONTAINER);
+    const container: Container = this.k8Factory.getK8(context).containers().readByRef(containerReference);
+    await this.waitForStableSavedState(container, podReference.name.name, false);
```
Fantomasa: **AI suggestion**: Tiny mismatch: the JSDoc above promises a fully signed freeze state, but `requireFreezeRound` is `false` here, and the script returns any stable signed round either way. Passing `true` would at least surface the frozen-fallback warning and the nicer timeout message.

src/core/pces-trimmer.ts
```diff
@@ -0,0 +1,222 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import fs from 'node:fs';
+import path from 'node:path';
```
Fantomasa: Super minor: the style guide (§3.3.5) prefers `PathEx` over `node:path`, and `PathEx.join`/`PathEx.basename` cover everything used here. Otherwise this class is a really nice piece of work, the fail-safe handling and tests are great!

examples/state-save-and-restore/Taskfile.yml
```diff
@@ -11,31 +11,26 @@ env:
...
   # Cluster Configuration
-  CLUSTER_NAME: "state-restore-cluster"
-  CONTEXT: "kind-state-restore-cluster"
-  CLUSTER_REF: "kind-state-restore-cluster"
+  CLUSTER_NAME: 'state-restore-cluster'
+  CONTEXT: 'kind-state-restore-cluster'
+  CLUSTER_REF: 'kind-state-restore-cluster'
 
   # State Save Configuration
-  STATE_SAVE_DIR: "{{ .USER_WORKING_DIR }}/saved-states"
```
jan-milenkov, line 25: are those leftover, do we still test the external database test elsewhere

src/core/network-nodes.ts
JeffreyDallas: updated to `true`

resources/wait-for-stable-saved-state.sh
JeffreyDallas: updated

examples/state-save-and-restore/Taskfile.yml
JeffreyDallas: this example is now consensus-state-only  ; external DB is covered by examples/external-database-test and the test-e2e-external-database E2E job.

src/core/constants.ts
JeffreyDallas: solo docs link https://github.com/hiero-ledger/solo-docs/pull/312

src/commands/node/handlers.ts
JeffreyDallas: Good catch — dropping it wasn't a deliberate trade, and the stable-state wait was never meant to absorb the drain's job. They answer different questions: the drain (added in #5358) gives the block stream time to reach the block node, while the new wait only fingerprints `data/saved`. And because a freeze state is normally already written by the time we observe `FREEZE_COMPLETE`, the stability check settles in a few seconds rather than the 20s #5358 was tuned to — so freeze was stopping nodes *sooner* than before, with nothing watching the block stream.

The tell was that `--freeze-block-drain-seconds` was still in `FREEZE_FLAGS` and still populated by `freezeConfigBuilder`, so it had become a silent no-op for `consensus network freeze` (upgrade still honoured it).

Fixed in 0a0b65fb1 — both now run before `stopNodes`:

```
checkAllNodesAreFrozen -> drainBlockStreamAfterFreeze -> waitForFrozenStateToBeStable -> stopNodes
```

Drain first, stability check last, so the final gate before the nodes go down is still "the saved state stopped changing", which is what state download depends on.

Worth noting for anyone reading later: no PR-gating job covers this combination. The only E2E that runs `consensus network freeze` is `separate-node-add`, which has no block node, and `block-node.test.ts` never freezes. The one place freeze and a block node coexist is the `state-save-and-restore` example, and Test Examples only runs on the nightly cron, not on PRs.

src/commands/node/handlers.ts
JeffreyDallas: You're right, and it's worse than just the wait — `--state-file` presence was the wrong signal in two directions.

State download explicitly accepts `STOPPED` nodes and only passes `FROZEN` into `normalizeDownloadedStateArchives` when *every* node is frozen, so a stopped capture normalizes against the highest common **signed** round. Feeding that archive to `node start --state-file` still took the frozen branch purely because the flag was set. On top of the ~5 minute `FREEZE_COMPLETE` timeout (300 x 1s, then `nodeNotReady`), that branch also skipped `enablePortForwarding`, `emitNodeStartedEvent`, `waitForTss`, `setGrpcWebEndpoint` and `addNodeStakes`, and left the phase at `FROZEN` — all of which ran for state-file restores before this PR, so it was a regression for that path rather than only a missing feature.

Fixed in 0a0b65fb1 by deriving the wait from the archive, as you suggested:

- `NetworkNodes.isFreezeStateArchive()` reads just the `stateMetadata.txt` entry via `AdmZip({readEntries: true})` (no full extraction, so it stays cheap on multi-GB archives), takes the highest round when several are present, and reuses the existing `selectHighestRound` / `readStateMetadataValue` helpers. Unreadable or metadata-less archives fall back to the ACTIVE path, i.e. pre-PR behaviour, rather than hanging.
- `uploadStateFiles` sets `config.restoredFromFreezeState` only when *every* uploaded archive is a freeze round; mixed or non-freeze sets replay into ACTIVE.
- `start` no longer branches on argv. Both verification groups are always in the task list, gated by `onlyWhenRestoringFreezeState` / `skipWhenRestoringFreezeState`.

One subtlety that would have been a silent bug: six of those tasks already declare their own `skip` (`enablePortForwarding` on `debugNodeAlias`/`forcePortForward`, `waitForTss` on `tssEnabled`, `setGrpcWebEndpoint` and `checkNodeProxiesAreActive` on `app`, `changeAllNodePhases` on `remoteConfig.isLoaded()`), so a `{...task, skip}` spread would have clobbered them. `withAdditionalSkip` composes instead, awaiting the declared skip because `checkNodeProxiesAreActive`'s is async. Side note: the existing `skipTaskWhenNodeStartSkipped` helper has the same clobbering flaw — I left it alone as out of scope, but it's worth a follow-up.

Also added 6 unit tests for the detection (freeze, stopped, multi-round, round-less `node add` layout, no metadata, unreadable archive).

JeffreyDallas: Correction to my previous reply: the archive-based approach I described was wrong, and it broke the State Save and Restore example ([run 34519290147](https://github.com/hiero-ledger/solo/actions/runs/34519290147/job/103012638711) on 0a0b65fb1 — `Node 'node1' is not ACTIVE [attempt = 300/300]`).

The reason is worth recording, because it also rules out keying off the selected round kind as you suggested. In that run the normalized round (1591) is an ordinary signed round — this CN version reports `SIGNING_WEIGHT_SUM: 0` for freeze states, so `normalizeDownloadedStateArchives` never selects a freeze round as fully signed — yet the nodes still logged `REPLAYING_EVENTS -> FREEZE_COMPLETE`. They re-froze because the *retained preconsensus events* carry the freeze transaction, not because the snapshot round was a freeze state. So `FREEZE_STATE` on the selected round says nothing about the status the restore will settle in; my detection correctly returned "not a freeze archive" and the nodes still came up frozen.

b959816d6 stops predicting. `checkAllNodesAreActiveOrFrozen` polls each node until it settles on ACTIVE or FREEZE_COMPLETE, records which, and the ACTIVE-only follow-up work (port forwarding, TSS, gRPC web endpoint, stakes, started event) is gated on the observed status. That keeps your original point fixed — a restore that replays into ACTIVE gets the full active-start treatment instead of timing out on FREEZE_COMPLETE — without depending on a property of the archive that does not determine the outcome. The archive inspection is reverted.

Worth flagging: Test Examples does not run on PRs, so b959816d6 has not been re-validated end to end yet.

src/core/network-nodes.ts
JeffreyDallas: Following up with what run [34519290147](https://github.com/hiero-ledger/solo/actions/runs/34519290147/job/103012638711) showed about this, since it changes how the flag reads in practice.

Passing `true` is right, but on this consensus-node version the `frozen-fallback` path is now the *normal* outcome rather than an edge case. Freeze states are reported with `SIGNING_WEIGHT_SUM: 0`, so the script never finds a fully signed freeze round, always falls back to the newest signed non-freeze round, and relabels the kind to `frozen-fallback` — which fires the warning on every stable poll. In that run's `solo.log` it appears **12 times** (6 per pod, across both the freeze flow and the state download, ~3 polls each):

```
[state-download] network-node1-0: deployment is FROZEN but no fully signed freeze round exists on disk yet; using the newest fully signed non-freeze round
```

So the behaviour is correct, but the wording oversells it: "yet" implies the freeze round is about to appear, when for this CN version it never will. Worth either downgrading it to `debug` or rewording to something like "no fully signed freeze round on disk (expected when the platform does not sign freeze states); using the newest fully signed non-freeze round". Happy to do either — say which you prefer.

This is also the same property that made the archive-based restore detection in 0a0b65fb1 wrong (see the `handlers.ts` thread): the selected round is a non-freeze round even when the network is frozen.

src/core/constants.ts
JeffreyDallas: Confirming the docs side is now actually in place: solo-docs#312 carries the three variables as of commit `69d58b0`, added to the existing **Pod and Network Readiness** table in `content/en/docs/advanced-solo-setup/using-environment-variables.md` alongside `NETWORK_NODE_ACTIVE_MAX_ATTEMPTS` and friends:

| Variable | Default |
| --- | --- |
| `STATE_DOWNLOAD_STABLE_MAX_ATTEMPTS` | `180` |
| `STATE_DOWNLOAD_STABLE_DELAY` | `2000` |
| `STATE_DOWNLOAD_STABLE_POLLS_REQUIRED` | `3` |

Note that PR's title still only mentions `EXPERIMENTAL_COPY_WRAPS_LIB_IN_PARALLEL`, so it now covers two unrelated additions — I'll broaden the title there.

resources/wait-for-stable-saved-state.sh
JeffreyDallas: Taken in f279168fe. One more gap in the same fallback chain, since your point was about future-proofing it: the fingerprint at the bottom now uses `find -printf`, which is GNU-specific.

```
find "${saved_dir}" -type f -printf '%s %T@ %p\n'
```

A container minimal enough to lack both `sha256sum` and `shasum` (busybox/alpine) would almost certainly also ship a `find` without `-printf`, so that branch would fail before the openssl hash ever ran. Not worth hardening for the consensus containers we actually target — they have GNU findutils and `sha256sum` — but it means the openssl fallback is effectively unreachable on the platforms that would need it. Fine to leave as best-effort; flagging so the chain isn't mistaken for real portability.

src/core/network-nodes.ts
```diff
@@ -149,12 +160,244 @@ export class NetworkNodes {
...
+  /**
+   * Determine whether a downloaded state archive was captured from a freeze state.
+   *
+   * A freeze-captured archive replays back into FREEZE_COMPLETE, while an archive
+   * captured from a node that was only stopped replays back into ACTIVE. Callers
+   * that restore an archive need to know which platform status to wait for, and
+   * the archive itself is the only source of that answer: the same
+   * `--state-file` restore can carry either kind.
+   */
+  public isFreezeStateArchive(archivePath: string): boolean {
```
JeffreyDallas: `isFreezeStateArchive` has no callers in `src/` — it's dead code left over from the archive-based prediction that `b959816d6`'s commit message says was reverted ("Reverts the archive inspection (isFreezeStateArchive and its uploadStateFiles bookkeeping)"). Please delete this method; the `AdmZip` import at line 27 becomes unused too once it's gone (double-check `STATE_METADATA_FILE_NAME` is still referenced elsewhere in the file before removing that constant).

src/commands/node/config-interfaces/node-start-config-class.ts
```diff
@@ -14,4 +14,6 @@ export interface NodeStartConfigClass extends NodeCommonConfigWithNodeAliases, C
   stagingDir: string;
   forcePortForward: Optional<boolean>;
   grpcWebEndpoints: string;
+  /** Set by uploadStateFiles: true when every uploaded archive was captured from a freeze state. */
```
JeffreyDallas: This comment describes the reverted archive-based mechanism. `uploadStateFiles` no longer sets `restoredFromFreezeState` — it's now set by `checkAllNodesAreActiveOrFrozen` (`tasks.ts:2494`) from the *observed* post-start platform status.
```suggestion
  /** Set by checkAllNodesAreActiveOrFrozen: true when every node settled in FREEZE_COMPLETE rather than ACTIVE. */
```

resources/wait-for-stable-saved-state.sh
```diff
@@ -0,0 +1,87 @@
...
+
+if command -v sha256sum >/dev/null 2>&1; then
+  hash_cmd=(sha256sum)
+elif command -v shasum >/dev/null 2>&1; then
+  hash_cmd=(shasum -a 256)
+elif command -v openssl >/dev/null 2>&1; then
+  hash_cmd=(openssl dgst -sha256 -r)
+else
+  echo "No SHA-256 implementation found in container" >&2
+  exit 14
```
JeffreyDallas: This exit code (no SHA-256 tool available) is a permanent condition, but the TS caller's catch treats every non-zero exit the same as "not stable yet" and retries for the full ~6-minute budget before giving up with a generic error. Not blocking since the consensus-node image always has one of these tools, but worth a fast-fail path if this script ever targets a leaner image.

src/commands/node/tasks.ts
```diff
@@ -2464,6 +2465,98 @@ export class NodeCommandTasks {
...
+
+        const statuses: string[] = await Promise.all(
+          nodeAliases.map((nodeAlias: NodeAlias): Promise<string> =>
+            this.waitForActiveOrFrozenStatus(context_, nodeAlias),
+          ),
+        );
+
+        // Only a network that came up entirely frozen skips the ACTIVE-only follow-up work; a
+        // mixed result is reported so the operator can see which node disagreed.
+        context_.config.restoredFromFreezeState = statuses.every(
```
JeffreyDallas: In a mixed result (e.g. one node ACTIVE, one FREEZE_COMPLETE), `restoredFromFreezeState` is `false`, so the ACTIVE-only path runs `checkNodesAndProxiesAreActive`, which will wait out the full timeout for the node stuck in FREEZE_COMPLETE (it can't reach ACTIVE without a fresh start). The task title reports the mismatch before that happens — was a dedicated mixed-result error considered, or is surfacing it via the eventual timeout acceptable?

src/core/network-nodes.ts
ddeskov-limechain: `isFreezeStateArchive` has no callers any more. It was the archive-based prediction that b959816d6 replaced with the observed-status wait. Please drop it together with the `AdmZip` import in this file (CLAUDE.md dead-code rule).

Same stale trace in `node-start-config-class.ts`: the doc comment on `restoredFromFreezeState` still says it is set by `uploadStateFiles`, but it is now set by `checkAllNodesAreActiveOrFrozen`.

src/commands/node/tasks.ts
ddeskov-limechain: When the nodes disagree (one ACTIVE, one FREEZE_COMPLETE), this sets `restoredFromFreezeState = false`, so the next task `checkNodesAndProxiesAreActive` runs and waits the full ACTIVE timeout on the frozen node before it fails. The only hint about what happened is the title of this task.

A split like that is not a valid restore result. I would throw here right away with the per-node summary in the message, so the operator sees the real cause in seconds instead of after the timeout.

examples/state-save-and-restore/Taskfile.yml
```diff
@@ -253,44 +238,106 @@ tasks:
...
-      - cmd: echo "Restoring database from dump..."
       - cmd: |
-          kubectl cp {{ .STATE_SAVE_DIR }}/database-dump.sql \
-          {{ .POSTGRES_CONTAINER_NAME }}:/tmp/database-dump.sql -n {{ .POSTGRES_DATABASE_NAMESPACE }}
+          if [ ! -d {{ .SAVED_KEYS_DIR }} ]; then
+            echo "⚠️  Saved key directory not found: {{ .SAVED_KEYS_DIR }}"
+            exit 1
+          fi
+      - cmd: mkdir -p "{{ .SOLO_CACHE_KEYS_DIR }}"
+      - cmd: find "{{ .SOLO_CACHE_KEYS_DIR }}" -mindepth 1 -maxdepth 1 -exec rm -rf {} +
```
ddeskov-limechain, line 283: This wipes everything in `~/.solo/cache/keys`, and that directory is shared by every deployment on the machine, not just this example. Someone running `task restore` locally loses the keys of their other networks.

The `cp` on the next line already overwrites the files this example needs, so the `rm -rf` can go.

ddeskov-limechain (commented):
Went through the whole diff again. The stable-state wait, round normalization and the observed-status start flow all read correctly, and the Test Examples run (including State Save and Restore and Multicluster Backup and Restore) passed on 1e312f12, so the "not yet re-run" note in the description is stale now. The current red job (One Shot Local Build) died in its `Install grpcurl` step, not in this code. Three mid-level points inline.

src/core/network-nodes.ts
JeffreyDallas: You're right, and that was my miss — when I reverted the archive-based approach I ran `git checkout -- src/core/network-nodes.ts`, which restored the file from the index where the method had already been committed in 0a0b65fb1, so it survived as dead code rather than being removed.

Both points are already fixed on the current head (a9b07b463): `isFreezeStateArchive` and the `AdmZip` import are gone from `network-nodes.ts` (0 occurrences), and the doc comment on `restoredFromFreezeState` now reads "Set by checkAllNodesAreActiveOrFrozen". `STATE_METADATA_FILE_NAME` is kept deliberately — it is still used by `normalizeDownloadedStateArchives`.

src/commands/node/tasks.ts
JeffreyDallas: Agreed — a split result is not a valid restore, and letting it fall through to the ACTIVE checks buried the real cause behind a 300-attempt timeout. Fixed on the current head (a9b07b463): `checkAllNodesAreActiveOrFrozen` now fails fast with the per-node summary rather than inferring `restoredFromFreezeState = false`.

```ts
const allFrozen: boolean = statuses.every((status: string): boolean => status === frozenStatusName);
const allActive: boolean = statuses.every((status: string): boolean => status !== frozenStatusName);
if (!allFrozen && !allActive) {
  throw new SoloErrors.component.nodeRestoreStatusMismatch(statusByNodeAlias);
}
```

`NodeRestoreStatusMismatchSoloError` is a registered `SoloErrors.component` subclass with its own code and troubleshooting steps, per the CLAUDE.md error-handling rule, and `test/unit/commands/node/check-all-nodes-active-or-frozen.test.ts` covers the split case alongside the all-frozen and all-active paths.

examples/state-save-and-restore/Taskfile.yml
JeffreyDallas: Agreed, and thank you — that would have destroyed real key material for anyone running the example on a machine with other deployments. Fixed in 19334bfcb.

I dropped the `rm -rf` as you suggested, and scoped the `chmod` the same way, since `chmod 600 "$SOLO_CACHE_KEYS_DIR"/*` reached every other deployment's keys for the same reason:

```yaml
- cmd: mkdir -p "{{ .SOLO_CACHE_KEYS_DIR }}"
# The Solo key cache is shared by every deployment on the machine, so only the files
# this example saved are written or chmod'ed; copying over them is enough to restore
# the network and any other deployment's key material is left untouched.
- cmd: |
    for saved_key_file in "{{ .SAVED_KEYS_DIR }}"/*; do
      cp "${saved_key_file}" "{{ .SOLO_CACHE_KEYS_DIR }}/"
      chmod 600 "{{ .SOLO_CACHE_KEYS_DIR }}/$(basename "${saved_key_file}")"
    done
```

Verified with a simulation of a cache holding both this example's keys and an unrelated deployment's: this example's files are overwritten and end up `600`, the other deployment's file keeps its contents and its mode. `task --list` parses the updated Taskfile.

jan-milenkov (approved):

PR #5382, opened 2026-07-28
feat: warn Homebrew users that Homebrew support is being removed

### LINKED ISSUE #5365: add a warning for the user when installing or upgrading Solo via Homebrew that we are removing Homebrew support and with plans to stop updating Homebrew on August 31st

ddeskov-limechain:
## Description

* Adds `src/core/homebrew-deprecation-notifier.ts`, a `HomebrewDeprecationNotifier` that prints a
  deprecation banner when the running Solo was installed by Homebrew: Homebrew support is being
  removed, the formula stops being updated after **August 31, 2026**, and the supported path is a
  global npm install (`brew uninstall solo && npm install -g @hiero-ledger/solo`), alongside the
  upgrade guide link the update notifier already publishes.
* Wires a single call into `src/index.ts`, next to the existing `VersionUpdateNotifier` at the end of
  `main()`, so the banner is the last thing a Homebrew user sees after each command they run.
* Identifies a Homebrew installation by the Cellar path segment Homebrew places **every** formula
  under — `<prefix>/Cellar/solo/<version>/libexec/lib/node_modules/@hiero-ledger/solo` — matched
  against `constants.ROOT_DIR`. That segment is the part that does not vary: the prefix differs per
  platform (`/opt/homebrew` on Apple silicon, `/usr/local` on Intel macOS,
  `/home/linuxbrew/.linuxbrew`, or a custom prefix), and the tap also ships versioned `solo@<version>`
  formulas, which the pattern covers. A global npm install that merely lives under a Homebrew prefix
  (`<prefix>/lib/node_modules/@hiero-ledger/solo`) is **not** a Homebrew install and does not match,
  so npm users never see the banner.
* Follows the conventions of the sibling `VersionUpdateNotifier`: skipped for non-TTY sessions so
  scripted and CI output is untouched, and every failure is swallowed into `logger.debug()` so the
  notice can never interfere with the command the user actually ran.
* Adds `test/unit/core/homebrew-deprecation-notifier.test.ts` — 6 cases covering detection across
  prefixes and versioned formulas, the npm-under-a-Homebrew-prefix false positive, banner content,
  the non-TTY skip, and the swallow-everything guarantee.

### Why the warning is raised at run time rather than at install time

Homebrew captures the output of a formula's install steps into its build log and shows the user a
spinner instead, so anything printed from Solo's `scripts/postinstall.mjs` — which is what the
formula's `npm install --prefix=#{libexec}` triggers — would never be seen by the person running
`brew install` or `brew upgrade`. The two surfaces Homebrew does put in front of the user are the
formula's own `caveats` block and the terminal they return to afterwards. This PR takes the second
one, which is guaranteed to be seen on the very next Solo command and keeps working for users who
installed before this deprecation was announced.

The `caveats` half lives in the formula, in `hiero-ledger/homebrew-tools`, outside this repository. It
is the natural complement to this change and I would suggest it as a follow-up there, so the warning
also lands at the exact moment of install and upgrade.

### Related Issues

* Closes #5365
* Related to #5364 (the README half of the same deprecation)

- [x] This PR added tests (unit, integration, and/or end-to-end)
- [x] This PR has no breaking changes
- [ ] Any technical debt has been documented as a separate issue and linked to this PR

The technical-debt box is unchecked deliberately: the complementary `caveats` change belongs to
`hiero-ledger/homebrew-tools`, and no issue has been opened there yet — say the word and I will file
one and link it here.

### Testing

- [x] This PR added unit tests
- [x] These changes required manual testing that is documented below

* Built the branch (`npx tsc` + `node resources/post-build-script.js`) and reproduced the layout the
  formula creates — `dist/` and `package.json` copied into
  `…/Cellar/solo/0.83.0/libexec/lib/node_modules/@hiero-ledger/solo` — then ran the **built** notifier
  from there. `ROOT_DIR` resolved inside the Cellar, `isInstalledViaHomebrew()` returned `true`, and
  the banner rendered:

  ```
  ================================================================================
    WARNING: Homebrew support for Solo is being removed.
    This Solo came from Homebrew, which stops being updated after August 31, 2026.
    Switch to npm to keep receiving updates:
      brew uninstall solo && npm install -g @hiero-ledger/solo
    Upgrade guide: https://solo.hiero.org/docs/simple-solo-setup/upgrading-solo/
  ================================================================================
  ```

* The same built code run from the repository checkout (`ROOT_DIR` = `…/GitHub/solo/dist`) reports
  `isInstalledViaHomebrew(): false` and prints nothing, which is the npm/development case.
* `npx mocha 'test/unit/core/homebrew-deprecation-notifier.test.ts'` → 6 passing. Full unit suite,
  `npx mocha 'test/unit/**/*.ts' --exclude 'test/unit/core/helm/**/*.ts'` → **1372 passing, 0
  failing**.
* `npx tsx solo.ts --version` still prints the version banner unchanged, confirming the added import
  and call do not disturb start-up.
* `npx tsc --noEmit` passes; `npx eslint` and `npx prettier --check` are clean on all three changed
  files (the one remaining `no-explicit-any` warning in `src/index.ts:29` is pre-existing on `main`).

The following was not tested:

* No real `brew install hiero-ledger/tools/solo` was performed. The Cellar layout used above was
  reproduced from the formula itself (`npm install … --prefix=#{libexec}` followed by
  `bin.install_symlink Dir["#{libexec}/bin/*"]`), not from an actual Homebrew run.
* `solo --version` (and its `--output json|yaml|wide` forms) exits through `SilentBreak` before the
  notifier is reached, so the banner does not appear there. That is deliberate — it keeps
  machine-readable output clean and matches where the existing update notifier fires.
* Non-TTY sessions are skipped by design, so a Homebrew user who only invokes Solo from scripts or CI
  is not warned by this change.
* The cut-off date is a constant (`August 31, 2026`); if the removal date moves, the constant has to
  move with it.


src/core/homebrew-deprecation-notifier.ts
jeromy-cannon, line 1: I think this actually belongs in the ruby formulas in the https://github.com/hiero-ledger/homebrew-tools repo

ddeskov-limechain: Done in [hiero-ledger/homebrew-tools#94](https://github.com/hiero-ledger/homebrew-tools/issues/94), which adds the warning to the formula's caveats. But they only show on brew install/brew upgrade. Since updates stop after August 31, existing users may never upgrade again and would never see it. This runtime banner is the only way to reach them, so I'd keep both.

jan-milenkov (commented):
Lgtm, please review too @JeffreyDallas, @jeromy-cannon

src/core/homebrew-deprecation-notifier.ts
jeromy-cannon: the pr check failed in that pr

src/core/homebrew-deprecation-notifier.ts
```diff
@@ -0,0 +1,68 @@
...
+ * installing or upgrading instead. Like the update notifier, it is skipped for non-interactive
+ * sessions and never surfaces errors to the user.
+ */
+export class HomebrewDeprecationNotifier {
+  /** The date after which the Homebrew formula stops being updated. */
+  private static readonly END_OF_UPDATES_DATE: string = 'August 31, 2026';
+
+  /** Published documentation for installing and upgrading Solo. */
+  private static readonly UPGRADE_GUIDE_URL: string = 'https://solo.hiero.org/docs/simple-solo-setup/upgrading-solo/';
+
```
jeromy-cannon, line 22: this is a duplicate of `VersionUpdateNotifier.UPGRADE_GUIDE_URL` (same string, same file pattern). Consider moving it to `constants.ts` and referencing it from both notifiers — checklist §1.

```suggestion
  /** Published documentation for installing and upgrading Solo. */
  private static readonly UPGRADE_GUIDE_URL: string = constants.SOLO_UPGRADE_GUIDE_URL;
```

and in `src/core/constants.ts`:
```
export const SOLO_UPGRADE_GUIDE_URL: string = 'https://solo.hiero.org/docs/simple-solo-setup/upgrading-solo/';
```

src/core/homebrew-deprecation-notifier.ts
ddeskov-limechain: yes, but I can't re-run it myself. I have no access to that functionality

jan-milenkov (approved):

PR #5390, opened 2026-07-28
fix: configure the container stack behind the Homebrew-installed podman on Linux

### LINKED ISSUE #5354: Solo does not configure the container stack it installs via Homebrew, so a stale system stack breaks podman
## Summary

Solo installs podman through Homebrew on Linux (`ClusterTaskManager` → `BrewPackageManager.installPackages(['podman'])`, because the distribution-native podman is too old for kind), but it never touches the system container configuration under `/etc/containers`. Rootful podman — the mode Solo uses on Linux (`PodmanMode.ROOTFUL`) — reads that system configuration regardless of which podman binary is invoked. On any machine that already carries an older container stack, the podman Solo just installed runs against configuration and helper binaries it cannot work with, and fails at cluster creation with errors that point nowhere near the cause.

This was found by the per-distro install validation work for #4888: the CI job added in #5353 hits the failures in sequence on a stock GitHub-hosted `ubuntu-latest` image, which preinstalls podman 4.x.

## Observed failures, in the order they surface

Each one only appears after the previous is worked around, so a user hits them one at a time.

1. **`registries.conf` format.** Podman 5+ refuses to load a v1-format file at all:
   `Error: loading registries configuration "/etc/containers/registries.conf": registries.conf must be in v2 format but is in v1 format`
2. **OCI runtime.** `brew install podman` installs its own `crun` into the brew prefix, but podman resolves the runtime from the absolute paths in `containers.conf` (`/usr/bin/crun`, …), never from `PATH`. So it uses the old system crun, which cannot parse the OCI spec version the new podman writes:
   `Error: OCI runtime error: crun: unknown version specified`
   (Observed with system crun 1.14.1 against Homebrew podman 6.x.)
3. **Network backend.** Homebrew packages no `netavark`/`aardvark-dns`, so podman falls through to the system ones from the old podman, which cannot deserialize what the new podman sends:
   `Error: netavark: failed to load network options: IO error: invalid type: sequence, expected a map`

## Why this matters for Solo

`solo cluster-ref ... / one-shot` on a Linux host that has ever had podman, docker with the containers stack, or a distro `containers-common` package installed can fail in any of these three ways. None of the messages mention Homebrew, the version skew, or `/etc/containers`, so the failure is very hard to attribute. Solo currently has no detection, no diagnostic, and no remediation for it.

## Suggested direction

Some combination of:

* After the brew podman install, detect the mismatch — compare the podman version against the resolved `crun`/`netavark`, or probe with a trivial `podman run` — and fail with an actionable `SoloError` naming the offending file or binary instead of surfacing the raw podman error.
* Have Solo write or overlay the container configuration it needs (`containers.conf` `[engine.runtimes]` pointing at the brew-prefix `crun`, `conmon_path`, `helper_binaries_dir`, and a v2 `registries.conf`), rather than inheriting whatever the host has. A drop-in under `containers.conf.d` is less invasive than replacing the user's files.
* Consider whether the helpers Homebrew does not ship (`netavark`, `aardvark-dns`) should be obtained the same way Solo already obtains podman itself — `PodmanDependencyManager` downloads from `api.github.com/repos/containers/podman/releases`, and both helpers publish checksummed release binaries at the same organisation.

## Reference

The workarounds `.github/workflows/flow-install-validation-runtime.yaml` applies in #5353 (a v2 `registries.conf`, a `containers.conf` pointing at the brew stack, and a pinned netavark/aardvark-dns install) are exactly what Solo does not do for its users. That workflow is CI standing in for this gap; it is not a fix.

Related to #4888, #4910.

Fantomasa:
## Description

Solo installs podman through Homebrew on Linux (the distribution podman is too old for kind) but left the system container configuration under `/etc/containers` untouched. On any host carrying an older container stack, the brew podman then ran against configuration and helper binaries it cannot work with, failing cluster creation with errors that point nowhere near the cause (v1 `registries.conf` load failure, `crun: unknown version specified`, `netavark: failed to load network options`).

This PR makes Solo configure the container stack it installs — without mutating the user's system files. When (and only when) the podman on `PATH` is Homebrew-managed, the rootful bootstrap now:

- **Downloads the network helpers Homebrew does not package.** New `NetavarkDependencyManager` and `AardvarkDnsDependencyManager` (shared `PodmanNetworkHelperDependencyManager` base) fetch checksum-verified `netavark`/`aardvark-dns` from the upstream `containers/*` GitHub releases into `~/.solo/bin/podman-helpers` — the same provenance and pattern as podman/gvproxy/vfkit. Version pins live in `version.ts` and track the brew podman major.
- **Generates a solo-owned `containers.conf` and v2 `registries.conf`** under `~/.solo/config`, pointing rootful podman at the brew `crun`/`conmon` and the downloaded helpers (`helper_binaries_dir` ordered solo-helpers → brew prefix → system directories). The user's `/etc/containers` is never written.
- **Forwards `CONTAINERS_CONF`/`CONTAINERS_REGISTRIES_CONF` everywhere rootful podman runs**: through the `sudo env … kind create cluster` invocation, `PodmanClient`'s sudo kind/podman commands (whose sudo prefix now also forwards `PATH`, since `secure_path` hides the brew bin directory), and the `KIND`/`CONTAINER_ENGINE` subprocess allowlists. Later solo invocations (cluster teardown, image loads) restore the variables from the persisted files — a stale v1 `registries.conf` breaks *every* podman command, not just cluster creation.
- **Probes the assembled stack** after configuration: verifies `crun`/`conmon` exist at the configured paths and runs rootful `podman info` under the generated configuration, failing with the new actionable `PodmanRuntimeConfigurationFailedSoloError` (SOLO-5077) instead of a raw podman error.

Hosts whose podman is not Homebrew-managed (distribution podman, docker) are left completely untouched — their container stack is presumed self-consistent.

### Related work

The workarounds `.github/workflows/flow-install-validation-runtime.yaml` applies in #5353 are exactly what this PR moves into the product. Once both land, that workflow's "Normalize System Container Configuration" and helper-install steps can be removed so CI validates Solo's own bootstrap instead of standing in for it.

### Testing

- 13 new unit tests: rootful config generation and placeholder substitution, env-argument assembly, persisted-configuration restore, helper `shouldInstall`/gunzip processing, `CONTAINERS_REGISTRIES_CONF` allowlist forwarding, and the sudo podman command shape (including the config arguments).
- `setupConfig` now sources its directories from the DI `HomeDirectory`/`CacheDir` tokens instead of `constants.SOLO_HOME_DIR`, so unit tests exercise it against the test container home rather than the developer's real `~/.solo`.
- End-to-end validation of the rootful flow on a stock `ubuntu-latest` image is what #5353's CI job provides; this PR is the product-side fix it stands in for.

Fixes #5354


JeffreyDallas:
If we decided to sunset Homebrew support, maybe no need to spend efforts to enhance/fix homebrew related features and bugs?

ddeskov-limechain (approved):
LGTM

Issue #5389, opened 2026-07-28
Remove deprecated command init


Ivo-Yankov:
Remove all code related to the deprecated command `solo init`. This should be done for solo `v0.90.0`

Issue #5388, opened 2026-07-28
Remove deprecated flag --chart-version


Ivo-Yankov:
Remove all code related to the deprecated flag `--chart-version`. This should be done for solo `v0.90.0`

Issue #5387, opened 2026-07-28
Remove deprecated flag --release-tag


Ivo-Yankov:
Remove all code related to the deprecated flag `--release-tag`. This should be done for solo `v0.90.0`

Issue #5386, opened 2026-07-28
Remove deprecated flag --relay-release


Ivo-Yankov:
Remove all code related to the deprecated flag `--relay-release`. This should be done for solo `v0.90.0`

PR #5391, opened 2026-07-28
feat: remove deprecated --image-tag flag

### LINKED ISSUE #5385: Remove deprecated flag --image-tag
Remove all code related to the deprecated flag `--image-tag`. This should be done for solo `v0.90.0`

ddeskov-limechain:
## Description

* Removes the deprecated `--image-tag` flag, superseded by `--component-image`. `--image-tag` took a
  bare tag and was rewritten into `block-node-server:<tag>`; `--component-image` takes the full image
  reference (registry or local) and is the only path the chart-values and kind-load logic use.
* Deleted: the `Flags.imageTag` definition and its entry in the `Flags.allFlags` registry, its
  registration on `block node add`, the `BlockNodeDeployConfigClass.imageTag` field, and the
  normalization block in `BlockNodeCommand` that copied the tag into `componentImage` when the latter
  was unset.
* Dropped the `--image-tag` key from the two checked-in generated `falcon-values.yaml` examples,
  which enumerate the `block node add` flag list.
* 4 files, 22 deletions, no additions — `--image-tag` had no default value and nothing else read it,
  so nothing had to be moved onto `--component-image`.

### Migration

`--image-tag 0.36.0-SNAPSHOT` becomes `--component-image block-node-server:0.36.0-SNAPSHOT`.

A falcon values file carrying an empty `--image-tag: ""` is unaffected: `appendConfigToArgv` skips
empty values, so the key never reaches yargs. A **non-empty** `--image-tag` in a values file or a
script will now fail against yargs' `.strict()` with an unknown-argument error, which is the intended
signal for a removed flag.

### Scheduling and conflicts

* The issue targets **v0.90.0**; `main` is currently `0.83.0`, so merging this now ships the removal in
  the next minor. That is a maintainer call — the branch is ready whenever the 0.90.0 window opens.
* Conflicts with the in-flight `05181-structured-deprecating-system` branch, which registers this same
  flag in a new `DeprecationRegistry` ("deprecated since v0.84.0, removed in v0.90.0"). If that lands
  first, this PR also has to drop the corresponding registry entry — a one-line rebase.
* The five removal PRs (#5385–#5389) touch overlapping files (`src/commands/flags.ts`,
  `src/core/config-manager.ts`, `resources/one-shot-falcon-prepare.yaml`,
  `test/unit/core/config-manager.test.ts`). Each is self-contained against `main`; whichever merges
  first, the others need a trivial rebase.

### Related Issues

* Closes #5385
* Related to #5386, #5387, #5388, #5389 (the other deprecated-surface removals scheduled for v0.90.0)

- [ ] This PR added tests (unit, integration, and/or end-to-end) — removal only; the existing suites cover the surviving `--component-image` path
- [ ] This PR has no breaking changes
- [x] Any technical debt has been documented as a separate issue and linked to this PR

**Breaking change, but the title cannot say so.** The repository's Title Check
(`step-security/conventional-pr-title-action`) rejects the `!` marker — a `feat!:` title fails with
"No release type found in pull request title" — so the title is plain `feat:`. The breaking change is
documented above and carried by the `BREAKING CHANGE:` footer in the commit body; please keep that
footer in the squash message so `semantic-release` picks it up (`.releaserc` maps
`{"breaking": true, "release": "minor"}`, so it bumps the minor version, not the major).

### Testing

- [ ] This PR added unit tests
- [x] These changes required manual testing that is documented below

* `npx tsc --noEmit` clean; `npx eslint` clean on both changed source files.
* Full unit suite `npx mocha 'test/unit/**/*.ts' --exclude 'test/unit/core/helm/**/*.ts'` →
  **1366 passing, 0 failing** (unchanged from `main`; no test referenced `--image-tag`).
* `npx tsx solo.ts block node add --help` no longer lists `--image-tag`, and still lists
  `--component-image`.
* Swept the repository for `image-tag` / `imageTag`: the only remaining hits are unrelated locals
  (`Helpers.checkDockerImageExists`, the mirror-node chart `*.image.tag` values, and the
  `BlockNodeLocalImageNotFound` error's constructor parameter).

The following was not tested:

* No cluster-backed run of `block node add` with `--component-image` — the removed code was a pure
  string rewrite ahead of the existing `componentImage` path, which is unchanged.
* The two `falcon-values.yaml` examples are checked-in generator output and were edited by hand; they
  were not regenerated with `solo one-shot falcon prepare`. Note they also still contain a stale
  `--block-node-chart-version` key from an older generator — pre-existing and out of scope here.


src/commands/block-node.ts
```diff
@@ -598,12 +596,6 @@ export class BlockNodeCommand extends BaseCommand {
               'Block node chart version',
             );
 
-            // --image-tag is kept for backwards compatibility; normalize it to componentImage
```
jan-milenkov, line 877: this is removed from the `ADD_FLAGS_LIST.optional` which will mean the flag won't be accepted when calling the command

jan-milenkov: i think this also invalidates the logic of the other PRs for flag deprecation, if this is logic you are relying on

Ivo-Yankov (changes requested):
Blocking for the duration of the deprecation period

src/commands/block-node.ts
ddeskov-limechain: Yes, this is intended as #5385 asks for full removal, so the flag is now rejected. The other PRs don't rely on this logic as each has its own aliasing.

PR #5396, opened 2026-07-28
fix(one-shot): reconcile printed versions with values-file overrides

### LINKED ISSUE #5384: Incorrect versions printed in falcon deploy
### Describe the bug

When using `solo one-shot falcon deploy` to create a deployment with a specific CN version, the printed versions at the end are incorrect.

The command should deploy the correct version list

### To Reproduce

falcon-values.yaml:
```
network:
  --consensus-node-version: "v0.77.0-rc.2"
  --node-aliases: "node1"
setup:
  --consensus-node-version: "v0.77.0-rc.2"
  --node-aliases: "node1"
consensusNode:
  --node-aliases: "node1"
```

```
solo one-shot falcon deploy --values-file falcon-values.yaml
```

Observe the printed versions when the command is completed:

```
 *** Versions Used: ***
-------------------------------------------------------------------------------
 - Solo Chart Version: 0.64.0
 - Consensus Node Version: v0.74.0
 - Block Node Version: 0.38.0
 - Mirror Node Version: v0.158.0
 - Explorer Version: 26.1.0
 - JSON RPC Relay Version: 0.77.0
```

### Additional Context

_No response_

sarbojitrana:
## Summary

Fixes #5384 — `solo one-shot falcon deploy` prints the wrong component versions in the "Versions Used" summary when a `--values-file` overrides a version for an individual component.

## Root Cause

In `DefaultOneShotDeployOrchestrator.buildDeployPipeline` (Initialize phase), `config.versions` is resolved from `argv` via `DeployArgvBuilders.resolveOneShotComponentVersions(argv, edgeEnabled)` **before** the `--values-file` is parsed:

```
src/commands/one-shot/orchestrator/deploy/default-one-shot-deploy-orchestrator.ts:171-204
```

The values file is parsed afterward and its per-component sections (`network`, `setup`, `blockNode`, `mirrorNode`, `explorerNode`, `relayNode`) are copied verbatim into `config.networkConfiguration`, `config.blockNodeConfiguration`, etc. Those per-component objects are what actually get merged into each component's deploy argv in `deploy-argv-builders.ts`, e.g.:

```ts
// buildMirrorNodeArgv, deploy-argv-builders.ts:240-249
const mirrorLocalConfig: AnyObject = {
  [optionFromFlag(Flags.mirrorNodeVersion)]: config.versions.mirror,
  ...config.mirrorNodeConfiguration,   // <- values-file override wins here
  ...
};
```

Because the object spread puts `config.xConfiguration` after the default version flag, a version set in the values file's per-component section always wins for the *actual deploy*. But `config.versions` itself is never updated to reflect that override, so `showVersions()` (`default-one-shot-deploy-orchestrator.ts:994-1005`), which prints straight from `config.versions`, shows the stale pre-values-file default instead of what was actually deployed.

This reproduces the issue exactly: the reporter's values file sets `--consensus-node-version: v0.77.0-rc.2` only under `network:`/`setup:` (not as a top-level CLI flag), so `resolveOneShotComponentVersions` — which only reads `argv` — resolves the default `v0.74.0`, and that stale value is what gets printed, even though `v0.77.0-rc.2` is what's actually deployed.

## Fix

Added `DefaultOneShotDeployOrchestrator.reconcileEffectiveVersions(config)`, called once the values file has been parsed and the network/setup consensus-version defaulting has run. For each of the 6 versions, it prefers the value found in that component's resolved configuration object (the same key the argv builders would use), falling back to the originally resolved default — mirroring the exact override precedence used when building each component's deploy argv, so the printed summary always matches what's actually deployed.

## Evidence

Added `test/unit/commands/one-shot/orchestrator/deploy/default-one-shot-deploy-orchestrator.test.ts` — `DefaultOneShotDeployOrchestrator reconcileEffectiveVersions`:
- unchanged when no values-file overrides are present
- reproduces the reported scenario: `--consensus-node-version` set only under `network`/`setup` is reflected in `config.versions.consensus` after reconciliation
- covers block/mirror/explorer/relay per-component version overrides

```
DefaultOneShotDeployOrchestrator reconcileEffectiveVersions
  ✔ leaves the resolved versions untouched when no values file overrides are present
  ✔ reflects a consensus node version overridden via the values file network/setup sections
  ✔ reflects per-component version overrides for block, mirror, explorer, and relay nodes

31 passing (57ms)
```

Also verified:
- `npx tsc --noEmit -p .` — clean (pre-existing unrelated errors in `test/unit/commands/wraps-key-path.test.ts` confirmed present on `main` before this change)
- `npx eslint` on both changed files — clean

## Test plan

- [x] Added unit tests covering the reported scenario and the general per-component override case
- [x] Existing orchestrator unit tests still pass (31/31)
- [x] Type-check and lint clean on changed files

### DIFF
--- a/src/commands/one-shot/orchestrator/deploy/default-one-shot-deploy-orchestrator.ts
+++ b/src/commands/one-shot/orchestrator/deploy/default-one-shot-deploy-orchestrator.ts
@@ -255,6 +255,9 @@ export class DefaultOneShotDeployOrchestrator implements OneShotDeployOrchestrat
             if (!config.setupConfiguration[releaseTagKey]) {
               config.setupConfiguration[releaseTagKey] = versions.consensus;
             }
+
+            this.reconcileEffectiveVersions(config);
+
             this.logger.addLogBindings({
               clusterReference: config.clusterRef,
               context: config.context,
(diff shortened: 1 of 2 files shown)


jan-milenkov:
Hi @sarbojitrana, thanks for the PR! Just a quick heads-up for future contributions: when you'd like to work on an issue, it's best to leave a comment asking to be assigned first. That helps everyone know the issue is already being worked on and avoids the chance of someone else picking it up at the same time, which could make your work overlap or become unnecessary. We really appreciate your contribution!

P.S. - Just so you know if you see a integration test failure is not caused by your PR, you don't need to fix it

jan-milenkov (approved):

sarbojitrana:
@jan-milenkov thank you for approving and merging my pull request. i'll try to contribute more and make sure to ask to be assigned before i start working.

Issue #5195, opened 2026-07-22
fix flaky "E2E Tests (Dual Cluster Full) / Dual Cluster Full" > failed: No pod found for: pods:app=haproxy-node3,solo.hedera.com/type=haproxy


jeromy-cannon:
https://github.com/hiero-ledger/solo/actions/runs/29771385446/job/88451387004?pr=5068

workflow logs:

[job-logs.txt](https://github.com/user-attachments/files/30281479/job-logs.txt)

solo logs:

[solo-test-e2e-dual-cluster-full.log.zip](https://github.com/user-attachments/files/30281480/solo-test-e2e-dual-cluster-full.log.zip)

JeffreyDallas:
Still happening sometimes

https://github.com/hiero-ledger/solo/actions/runs/30363263315/job/90300878817?pr=5363

JeffreyDallas:
Sees running ok now

PR #5426, opened 2026-07-29
fix: forward the JVM debug port before the node activeness check

### LINKED ISSUE #5399: Nightly extended jvm-debugger test fails consistently
### Describe the bug

The `Extended Test (jvm-debugger)` test fails consistently. It appears that reverting [this change](https://github.com/hiero-ledger/solo/pull/5257/changes#diff-4ad819485fe844983ae2f5931ac6744b172fe4f3606f2ac9d700ef2eb509eb0bR1092) fixes it, but it should be investigated further why that happens, and if determine wether this is a test/script issue, or a bug:

```
- this.tasks.enablePortForwarding(true),
this.tasks.checkNodesAndProxiesAreActive('nodeAliases'),
+ this.tasks.enablePortForwarding(true),
```

The test should pass.

### To Reproduce

https://github.com/hiero-ledger/solo/actions/runs/30336675581/job/90261611771

### Additional Context

_No response_

Fantomasa:
## Description

The nightly `Extended Test (jvm-debugger)` has failed consistently since #5257. That PR deferred all port forwarding in `consensus node start` until after `checkNodesAndProxiesAreActive`, so persistent forwards would never target an HAProxy pod its startup recovery had just replaced — a sound goal, but the activeness check is also where the *"attach JVM debugger now"* prompt lives. A node started with `--debug-node-alias` suspends its JVM (`suspend=y`) until a debugger attaches through the JDWP forward, which after the reorder no longer existed when the check ran. The check waited on a node that could never become ACTIVE: a deadlock that the test's timeout converted into a consistent failure (exit 124, hung at `Check network pod:` with no forward ever created).

Answering the question in #5399 directly: it is a product bug (a broken ordering contract for debug mode), not a test/script issue — but a plain revert would reintroduce the HAProxy startup flake #5257 fixed (#5195).

### The fix

`enablePortForwarding(enablePortForwardHaProxy)` bundled two forwards with opposite ordering needs, so it is split into single-purpose tasks:

* **`enableJvmDebugPortForwarding()`** — the JDWP forward to the consensus node pod, restored to *before* the activeness check. HAProxy recovery never recreates consensus node pods, so deferring this forward bought nothing.
* **`enableGrpcPortForwarding()`** — the HAProxy gRPC forward, kept *after* the proxy readiness checks, preserving #5257's protection.

Every other node flow (`restart`, `destroy`, add/update/delete) only ever performed the debug half — the HAProxy half was gated off by the boolean parameter — so they now call the debug task directly and the parameter is gone.

### Verification

Ran `.github/workflows/script/test-jvm-debugger.sh` locally end to end (kind cluster, 2-node network, `--debug-node-alias node2`): the run shows the fixed sequence — `Enable port forwarding for JVM debug port` → activeness check with the attach prompt, during which the JDWP probe connects through the forward, handshakes, and resumes the JVM (`gRPC readiness 3/3` on both nodes) → `Enable port forwarding for GRPC port` after the proxy checks — ending in `JDWP debugger connected and JVM resumed successfully` and `PASS: All verification steps passed` (exit 0).

Also: full unit suite passing, typecheck and lint clean.

Fixes #5399


JeffreyDallas:
Already have a fix PR pending https://github.com/hiero-ledger/solo/pull/5808

PR #5498, opened 2026-07-30
fix: handle missing RemoteConfig in one-shot deploy

### LINKED ISSUE #5444: feat: clean up and start over when the remote config is missing on a kind cluster
### Description

A missing `solo-remote-config` ConfigMap fails with `ResourceNotFound` (SOLO-5001)
(`remote-config-runtime-state.ts:331`). For a local kind cluster there is nothing to preserve, so the
user is left to clean up by hand.

- **Do:** when the remote config is not found **and** the target is a kind cluster, clean up the
  leftover state and start over. Keep the plain fail-fast for non-kind clusters.
- **Acceptance:** running against a kind cluster with a missing remote config recovers automatically
  (or with a single confirmation) instead of dead-ending on SOLO-5001.

Ivo-Yankov:
## Description

This PR fixes the case where a missing `RemoteConfig` configmap results in an unhandled error. The introduced changes put the deployment in a self-healing state, where it will destroy the existing deployment before continuing.

### Related Issues

* Closes #5444
* Related to #

### Testing

* TBD

The following was not tested:

* TBD

<details>
<summary>
Commit message guidelines
</summary>

</details>


src/core/helpers.ts
```diff
@@ -558,6 +560,10 @@ export class Helpers {
     return consensusNode ? consensusNode.context : undefined;
   }
 
+  public static isKindContext(context: string): boolean {
+    return typeof context === 'string' && context.startsWith(Helpers.KIND_CONTEXT_PREFIX);
```
ddeskov-limechain: The one-shot deploy orchestrator already has this exact method — `default-one-shot-deploy-orchestrator.ts:1254`, same `startsWith('kind-')` body — and it is still used at lines 981 and 1240. Could we delete the private one, import `Helpers` there, and call this instead? We also carry two sets of tests for it now: the new ones here in `helpers.test.ts`, and the old private-access ones in `default-one-shot-deploy-orchestrator.test.ts:98-115` which can go.

Small thing on the signature: the `typeof` check does the job the type should do — the test has to write `const unresolvedContext: string = undefined` to reach it.

```suggestion
  public static isKindContext(context: string | undefined): boolean {
    return !!context?.startsWith(Helpers.KIND_CONTEXT_PREFIX);
  }
```

ddeskov-limechain (commented):
Nice approach — detecting the orphan before the deploy and giving the other commands a typed error reads well, and the unit coverage is thorough. Two suggestions inline.

One question: the description still says "Manual testing: TBD" and the checklist is empty. This path can destroy a deployment on its own, so could you note what was run end to end — delete the kind cluster, recreate it, then `one-shot single deploy` — and confirm the auto-clean completes and the redeploy succeeds? There is no e2e covering the recovery itself.

src/business/runtime-state/config/remote/remote-config-runtime-state.ts
```diff
@@ -393,7 +397,21 @@ export class RemoteConfigRuntimeState implements RemoteConfigRuntimeStateApi {
...
-    await this.load(this.namespace, context);
+    try {
+      await this.load(this.namespace, context);
+    } catch (error) {
+      if (RemoteConfigRuntimeState.isMissingRemoteConfigError(error) && Helpers.isKindContext(context)) {
+        throw new SoloErrors.config.remoteConfigMissingOnKindCluster(
+          deploymentName,
+          // populateClusterReferences leaves this.namespace as the raw flag string when the deployment
+          // is not in the local config, so fall back to the flag rather than reading .name off a string.
+          this.namespace?.name ?? this.configManager.getFlag<NamespaceNameAsString>(flags.namespace),
```
ddeskov-limechain: The comment describes a bug instead of fixing it. `this.namespace` is typed `NamespaceName`, so `load()` on line 401 already passes that raw string into `configMaps().read()`, where `namespace?.name` is `undefined`. The fallback hides it here, but the load right above stays wrong.

```suggestion
          this.namespace?.name,
```

Then normalize at the source, lines 360-363. The flag holds either a string or a `NamespaceName` — `middlewares.ts:191` and the one-shot orchestrator set an object, the CLI sets a string. `Resolvers.namespace` (`resolvers.ts:26-29`) already handles both:

```ts
      const namespaceFromConfig: NamespaceName | string = this.configManager.getFlag(flags.namespace);
      if (namespaceFromConfig) {
        this.namespace =
          typeof namespaceFromConfig === 'string' ? NamespaceName.of(namespaceFromConfig) : namespaceFromConfig;
      }
```

jan-milenkov (dismissed):

jan-milenkov (approved):

PR #5527, opened 2026-07-30
fix(destroy): preserve shared cluster resources during one-shot destroy

### LINKED ISSUE #5506: fix: one-shot destroy runs cluster reset and deletes shared cluster-scoped resources
### Description

The destroy orchestrator renders a "Cluster reset" step that invokes `cluster-ref config reset`
(`default-one-shot-destroy-orchestrator.ts:190-196`), whose task list runs `uninstallMetricsServer`,
`uninstallPrometheusStack`, `uninstallMinioOperator`, and `uninstallPodMonitorRole`
(`cluster/tasks.ts:549-556`) - all cluster-scoped or cluster-shared. `uninstallPodMonitorRole`
additionally ignores the result of its own `clusterRoleExists` check and deletes unconditionally,
swallowing errors (`cluster/tasks.ts:400-412`), so destroying one deployment can break another live
deployment sharing the cluster.

- **Do:** decide whether one-shot destroy should run a cluster reset at all when the cluster is
  shared, and guard the ClusterRole deletion on the existence check (reference-count or scope it per
  deployment).
- **Acceptance:** destroying one deployment never removes cluster-scoped resources another live
  deployment depends on.

AdityaOP007:
# Fix shared cluster resource cleanup during one-shot destroy
#5506 
## Summary

This PR fixes **#5506** by preventing one-shot destroy from removing cluster-scoped resources that may still be required by other active deployments sharing the same Kubernetes cluster.

Previously, the destroy orchestrator always executed the **Cluster reset** step, which removed shared resources such as the Metrics Server, Prometheus Stack, MinIO Operator, and the PodMonitor ClusterRole. As a result, destroying one deployment could unintentionally impact other live deployments running in the same cluster.

## Changes

* Detect whether the target cluster is shared by checking the existing remote configuration records.
* Skip the **Cluster reset** step when other deployments are still present in the cluster.
* Preserve cluster-scoped resources for shared cluster environments while continuing to clean up deployment-specific resources.
* Refactor `uninstallPodMonitorRole()` to properly respect the ClusterRole existence check before attempting deletion.
* Remove silent error swallowing and allow unexpected Kubernetes API failures to propagate through the existing error handling flow.
* Add unit tests covering shared and dedicated cluster destroy behavior, as well as the updated ClusterRole uninstall logic.

### Shared Cluster

* Destroying a deployment removes only its deployment-specific resources.
* Shared cluster resources remain intact.
* Other deployments in the cluster continue to function without interruption.

### Dedicated Cluster

* Existing behavior is preserved.
* Cluster reset continues to execute as expected during destroy.

## Testing

The implementation was validated with unit tests covering:

* Shared cluster destroy flow
* Dedicated cluster destroy flow
* ClusterRole deletion when the resource exists
* ClusterRole deletion when the resource does not exist
* Kubernetes API failure handling during ClusterRole removal

The modified functionality passes all relevant tests. Any remaining test failures are unrelated pre-existing environment/Docker issues outside the scope of this change.

## Why

This change satisfies the acceptance criteria for **#5506** by ensuring that destroying one deployment never removes cluster-scoped resources that another active deployment depends on, while maintaining backward compatibility for dedicated cluster deployments.

**Fixes #5506**


AdityaOP007:
> Merging to `main` in this repository is managed by Trunk.
> 
> * [x]  To merge this pull request, check the box to the left or comment `/trunk merge` below.
> 
> After your PR is submitted to the merge queue, this comment will be automatically updated with its status. If the PR fails, failure details will also be posted here

/trunk merge

ddeskov-limechain (commented):
Thanks for picking this up @AdityaOP007 — the diagnosis is right, and a few of the decisions here are
exactly what I'd want: capturing the `clusterRoleExists` result and branching on it instead of
deleting inside the `try`, throwing `SoloErrors.system.clusterRoleCheckFailed` rather than swallowing,
and treating the shared-cluster case as "suppress the reset" rather than "warn and continue". The
ClusterRole rewrite in `cluster/tasks.ts` I'd take as-is.

A few things I think need addressing before this can close #5506.

### 1. The shared-cluster detection is fail-open

`ClusterChecks.hasOtherDeployments` catches its own error and returns `false`, and the orchestrator
call site wraps it in a second `try/catch` that only warns. Both paths leave `skipClusterReset`
unset, which means "not shared", which means proceed to delete. So a transient API error or a missing
RBAC permission reproduces the exact bug this PR is fixing — and reproduces it silently.

For a check whose answer gates a destructive operation, I think the lookup failure has to propagate.
An unreadable cluster must not be indistinguishable from an empty one.

### 2. The check is skipped when the destroyed deployment's own remote config is missing

The guard is `if (remoteConfigLoaded && !config.skipClusterCleanup)`, where `remoteConfigLoaded` comes
from `loadRemoteConfigOrWarn()` for the deployment *being destroyed*. If that deployment's remote
config is absent or corrupt, the shared-cluster check never runs at all and the reset proceeds.

That's the case I'd worry about most, because a half-destroyed or broken deployment is exactly when
someone reaches for `destroy`. Whether *other* deployments exist is independent of whether *this*
deployment's remote config parsed.

### 3. Checking at `Initialize` forces a namespace exclusion that checking later wouldn't need

The census runs during `Initialize`, before anything is destroyed, so the deployment being destroyed
is still counted and has to be filtered out via `cm.namespace.name !== excludeNamespace`. That leans
on `config.namespace` being resolved (it can be left undefined on the `skipAll` paths, where
`config.namespace.name` would throw — the surrounding `catch` would hide it), and it mis-handles a
deployment that shares a namespace with something else.

Worth noting: the `Destroy` composite phase is sequential, and consensus destroy
(`NetworkCommand.destroy`, `network.ts:1058`) deletes the deployment's own `solo-remote-config`
ConfigMap *before* `Cluster reset` runs. So if the census is taken at reset time instead, the
deployment being destroyed has already removed itself from the count and no exclusion logic is needed
— last one out sees zero and cleans up fully.

### 4. `solo cluster-ref config reset` invoked directly is still unguarded

Because the guard is a `skipClusterReset` field on `OneShotSingleDestroyConfigClass` read by the
orchestrator's skip callback, it only covers one-shot destroy. `solo cluster-ref config reset --force`
run by a user, a script, or the e2e suite still removes the shared resources.

`uninstallClusterChart` is the choke point all of those route through — putting the guard there covers
every caller, and has the side benefit of needing no new config field and no new DI injection.
Relatedly, the pre-existing `!argv.force &&` condition in `uninstallClusterChart` is untouched, so
`--force` remains a blanket bypass of the older safeguard underneath the new orchestrator-level skip.

### 5. Three of the four uninstalls are only covered indirectly

`uninstallMetricsServer`, `uninstallPrometheusStack` and `uninstallMinioOperator` are protected only
by the whole phase being skipped, and only in the one-shot flow. That does work — but it works because
the reset phase happens to contain no deployment-scoped cleanup, so skipping it wholesale leaks
nothing. That's the load-bearing fact behind the design and it'd be worth stating explicitly (in a
comment or the PR body), since a future addition to that task list would quietly invalidate it.

### Smaller things

- **`package-lock.json`**: the diff strips `"peer": true` from ~25 entries, which looks like an
  incidental `npm install` regeneration rather than an intended change. Probably worth reverting so
  the diff stays scoped to the fix.
- **`test/unit/commands/cluster.test.ts`**: `let rbacStub: any`, untyped `const tasks = ...`, and
  untyped arrow callbacks will trip the repo's `@typescript-eslint/no-explicit-any` and explicit-type
  rules (CLAUDE.md requires explicit annotations on every declaration and every callback, including
  `it()`/`describe()`); there's also trailing whitespace on a number of lines.
- **`default-one-shot-destroy-orchestrator.test.ts`**: `type MockListr = MockType` is unused, and
  driving the `Initialize` closure via `initializeTask?.task(...)` with a hand-rolled partial config
  couples the test fairly tightly to that closure's internals.

---

For what it's worth, I had an alternative implementation in progress on
`fix/5506-one-shot-destroy-cluster-reset-guard` before I saw this PR — it puts the guard in
`uninstallClusterChart`, takes the census at reset time, and fails closed. I'm not opening it as a
competing PR; happy for this one to be the vehicle. Feel free to pull anything useful from that branch
if it helps, and I'm glad to review a revision here.

AdityaOP007:
> Thanks for picking this up @AdityaOP007 — the diagnosis is right, and a few of the decisions here are exactly what I'd want: capturing the `clusterRoleExists` result and branching on it instead of deleting inside the `try`, throwing `SoloErrors.system.clusterRoleCheckFailed` rather than swallowing, and treating the shared-cluster case as "suppress the reset" rather than "warn and continue". The ClusterRole rewrite in `cluster/tasks.ts` I'd take as-is.
> 
> A few things I think need addressing before this can close #5506.
> 
> ### 1. The shared-cluster detection is fail-open
> `ClusterChecks.hasOtherDeployments` catches its own error and returns `false`, and the orchestrator call site wraps it in a second `try/catch` that only warns. Both paths leave `skipClusterReset` unset, which means "not shared", which means proceed to delete. So a transient API error or a missing RBAC permission reproduces the exact bug this PR is fixing — and reproduces it silently.
> 
> For a check whose answer gates a destructive operation, I think the lookup failure has to propagate. An unreadable cluster must not be indistinguishable from an empty one.
> 
> ### 2. The check is skipped when the destroyed deployment's own remote config is missing
> The guard is `if (remoteConfigLoaded && !config.skipClusterCleanup)`, where `remoteConfigLoaded` comes from `loadRemoteConfigOrWarn()` for the deployment _being destroyed_. If that deployment's remote config is absent or corrupt, the shared-cluster check never runs at all and the reset proceeds.
> 
> That's the case I'd worry about most, because a half-destroyed or broken deployment is exactly when someone reaches for `destroy`. Whether _other_ deployments exist is independent of whether _this_ deployment's remote config parsed.
> 
> ### 3. Checking at `Initialize` forces a namespace exclusion that checking later wouldn't need
> The census runs during `Initialize`, before anything is destroyed, so the deployment being destroyed is still counted and has to be filtered out via `cm.namespace.name !== excludeNamespace`. That leans on `config.namespace` being resolved (it can be left undefined on the `skipAll` paths, where `config.namespace.name` would throw — the surrounding `catch` would hide it), and it mis-handles a deployment that shares a namespace with something else.
> 
> Worth noting: the `Destroy` composite phase is sequential, and consensus destroy (`NetworkCommand.destroy`, `network.ts:1058`) deletes the deployment's own `solo-remote-config` ConfigMap _before_ `Cluster reset` runs. So if the census is taken at reset time instead, the deployment being destroyed has already removed itself from the count and no exclusion logic is needed — last one out sees zero and cleans up fully.
> 
> ### 4. `solo cluster-ref config reset` invoked directly is still unguarded
> Because the guard is a `skipClusterReset` field on `OneShotSingleDestroyConfigClass` read by the orchestrator's skip callback, it only covers one-shot destroy. `solo cluster-ref config reset --force` run by a user, a script, or the e2e suite still removes the shared resources.
> 
> `uninstallClusterChart` is the choke point all of those route through — putting the guard there covers every caller, and has the side benefit of needing no new config field and no new DI injection. Relatedly, the pre-existing `!argv.force &&` condition in `uninstallClusterChart` is untouched, so `--force` remains a blanket bypass of the older safeguard underneath the new orchestrator-level skip.
> 
> ### 5. Three of the four uninstalls are only covered indirectly
> `uninstallMetricsServer`, `uninstallPrometheusStack` and `uninstallMinioOperator` are protected only by the whole phase being skipped, and only in the one-shot flow. That does work — but it works because the reset phase happens to contain no deployment-scoped cleanup, so skipping it wholesale leaks nothing. That's the load-bearing fact behind the design and it'd be worth stating explicitly (in a comment or the PR body), since a future addition to that task list would quietly invalidate it.
> 
> ### Smaller things
> * **`package-lock.json`**: the diff strips `"peer": true` from ~25 entries, which looks like an
>   incidental `npm install` regeneration rather than an intended change. Probably worth reverting so
>   the diff stays scoped to the fix.
> * **`test/unit/commands/cluster.test.ts`**: `let rbacStub: any`, untyped `const tasks = ...`, and
>   untyped arrow callbacks will trip the repo's `@typescript-eslint/no-explicit-any` and explicit-type
>   rules (CLAUDE.md requires explicit annotations on every declaration and every callback, including
>   `it()`/`describe()`); there's also trailing whitespace on a number of lines.
> * **`default-one-shot-destroy-orchestrator.test.ts`**: `type MockListr = MockType` is unused, and
>   driving the `Initialize` closure via `initializeTask?.task(...)` with a hand-rolled partial config
>   couples the test fairly tightly to that closure's internals.
> 
> For what it's worth, I had an alternative implementation in progress on `fix/5506-one-shot-destroy-cluster-reset-guard` before I saw this PR — it puts the guard in `uninstallClusterChart`, takes the census at reset time, and fails closed. I'm not opening it as a competing PR; happy for this one to be the vehicle. Feel free to pull anything useful from that branch if it helps, and I'm glad to review a revision here.

Thanks @ddeskov-limechain for the detailed review and for taking the time to explain the reasoning behind each point — this is really helpful.

I agree with the concerns you've raised, especially around failing closed for destructive operations and moving the shared-cluster check closer to the actual reset point. Your suggestions about guarding the lower-level cleanup path instead of only the one-shot orchestrator also make sense and provide broader coverage.

I'll rework the implementation to address these points, clean up the unrelated changes and test issues, and update the PR with a revised approach. I'll also take a look at the ideas from your `fix/5506-one-shot-destroy-cluster-reset-guard` branch where they fit the existing design.

Thanks again for the thorough feedback!

AdityaOP007:
> Just testing if you are an AI agent with this comment

Just giving instant replies to all makes me behave like an AI agent, I'm on my vaccation and in my free time I love doing this!

AdityaOP007:
@ddeskov-limechain I've pushed a new revision addressing the points you raised. The shared-cluster guard now fails closed, the protection has been moved to the shared cleanup path, the orchestrator-specific logic has been removed, and I cleaned up the remaining review comments, including the DCO sign-off. I'd really appreciate it if you could take another look when you have time.

AdityaOP007:
@ddeskov-limechain Hey! just wanted to know , what's the update regarding this PR

AdityaOP007:
Hey @jeromy-cannon , @ddeskov-limechain What's the update for this PR , does it requires any changes from my side in this.

jeromy-cannon:
@AdityaOP007 we need to review again, but definitely this needs to be fixed: `Commits must have verified signatures.`

AdityaOP007:
> @AdityaOP007 we need to review again, but definitely this needs to be fixed: `Commits must have verified signatures.`


Sure

AdityaOP007:
hey @jeromy-cannon , @ddeskov-limechain I've initialize the new changes for fixing the checks, please take a look at this .

jeromy-cannon:
@AdityaOP007 , are you able to update your branch and resolve the merge conflicts?  Also, we had a team meeting today, so we will try to interact with you at least once a day to get this merged.

AdityaOP007:
> @AdityaOP007 , are you able to update your branch and resolve the merge conflicts?  Also, we had a team meeting today, so we will try to interact with you at least once a day to get this merged.

Thanks for reviewing this PR once again! sure I'll fix the merge conflicts to get it resolved ASAP and get you updated soon

AdityaOP007:
@jeromy-cannon please take a look at this almost all the checks have been passed but one is failing, if you have some time then it would be very helpful if you suggest some advices for resolving the remaining one.

jeromy-cannon:
@AdityaOP007 , I think we might have a flaky test on `main` that needs to be fixed, also.

jeromy-cannon:
@AdityaOP007 , also please correct the 'commits must have verified signatures'.
image

image

https://solo.hiero.org/docs/community-contributions/#2-cryptographically-signed-commits-required

AdityaOP007:
@jeromy-cannon Thanks for the feedback! I'll rebase onto the latest main, resolve the merge conflicts, investigate and address the flaky test behavior, and update the commits with verified signatures. I'll push the updated branch once everything is validated.

AdityaOP007:
@jeromy-cannon please check once all the checks have been passed now and the merge conflicts have also solved , please take a look.

AdityaOP007:
@jeromy-cannon, could you please take a look at this once I’ve fixed the failing checks? I’d really appreciate your review.

src/commands/network.ts
```diff
@@ -1056,6 +1056,25 @@ export class NetworkCommand extends BaseCommand {
       await Promise.allSettled(
         contexts.map(async (context): Promise<void> => {
           await this.k8Factory.getK8(context).configMaps().delete(namespace, constants.SOLO_REMOTE_CONFIGMAP_NAME);
+
+          let exists: boolean = await this.k8Factory
```
jeromy-cannon: duplicate code fragment — this exact 19-line polling block appears again at the second `configMaps().delete` call in the `else` branch of the same `destroyTask` method. Consider extracting:

```typescript
private async waitForConfigMapDeletion(context: Context, namespace: NamespaceName): Promise<void> {
  let exists: boolean = await this.k8Factory
    .getK8(context)
    .configMaps()
    .exists(namespace, constants.SOLO_REMOTE_CONFIGMAP_NAME);
  let attempts: number = 0;
  while (exists && attempts < constants.NETWORK_DESTROY_WAIT_TIMEOUT) {
    await sleep(Duration.ofSeconds(1));
    exists = await this.k8Factory
      .getK8(context)
      .configMaps()
      .exists(namespace, constants.SOLO_REMOTE_CONFIGMAP_NAME);
    attempts++;
  }
  if (exists) {
    throw new SoloErrors.system.timeout(
      `Timeout waiting for configMap ${constants.SOLO_REMOTE_CONFIGMAP_NAME} to be deleted.`,
    );
  }
}
```

Then both call sites become a single `await this.waitForConfigMapDeletion(context, namespace);`. §10.1.1 / checklist §1.

src/core/cluster-checks.ts
```diff
@@ -85,11 +86,18 @@ export class ClusterChecks {
         .configMaps()
         .listForAllNamespaces([constants.SOLO_REMOTE_CONFIGMAP_LABEL_SELECTOR]);
 
-      return configmaps.length > 0;
-    } catch (error) {
-      this.logger.error('Failed to find remote config:', error);
+      for (const cm of configmaps) {
```
jeromy-cannon: `listForAllNamespaces` already returns only configmaps that exist at the time of the call — calling `.exists()` on each one is redundant and adds an extra API round-trip per deployment. The original `return configmaps.length > 0` was correct.

```suggestion
      return configmaps.length > 0;
```

src/commands/cluster/tasks.ts
```diff
@@ -530,17 +536,16 @@ export class ClusterCommandTasks {
...
-            default: false,
-            message:
-              'There is remote config for one of the deployments' +
-              'Are you sure you would like to uninstall the cluster?',
-          });
-
-          if (!confirm) {
-            throw new UserBreak('Aborted application by user prompt');
-          }
+        const isShared: boolean = await this.clusterChecks.isRemoteConfigPresentInAnyNamespace(context);
```
jeromy-cannon: we should maintain backwards compatibility here. The old code honored `argv.force`: if the user passed `--force`, the shared-cluster check was bypassed and the cluster reset ran anyway. The new code always skips the reset when the cluster is shared, regardless of `argv.force`. A user who previously destroyed a shared cluster with `--force` will now find that flag silently ineffective.

Consider:
```suggestion
        const isShared: boolean =
          !argv.force && (await this.clusterChecks.isRemoteConfigPresentInAnyNamespace(context));
```

test/unit/commands/one-shot/orchestrator/destroy/default-one-shot-destroy-orchestrator.test.ts
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import {describe, afterEach} from 'mocha';
+import sinon from 'sinon';
+
+describe('DefaultOneShotDestroyOrchestrator', (): void => {
```
jeromy-cannon: this file has no test cases — it is just a shell with an `afterEach`. The PR description says this covers the shared and dedicated cluster destroy flows, but that coverage is absent. Either add the tests or remove the file.

AdityaOP007:
Thanks for the detailed review, @jeromy-cannon . I’m addressing all the comments now and will rerun the full checks without affecting the existing fix. I’ll update the PR once everything is ready for re-review.

AdityaOP007:
@jeromy-cannon Hi, I’ve addressed all four review comments and pushed the updates.

Extracted the duplicated ConfigMap polling into a helper.
Removed the redundant .exists() checks.
Preserved --force backward compatibility.
Removed the empty test file.

I also re-ran the available validations: task check, task build, TypeScript, ESLint, Prettier, and the relevant unit tests all pass. The changes are now pushed to the PR. Please have a look when you get a chance.

jeromy-cannon (approved):

AdityaOP007:
@jeromy-cannon Thanks for the review and approval! I’ve addressed all the requested changes and will leave the PR as-is for the merge.

AdityaOP007:
Hey @jeromy-cannon Any changes requires? Before merging this PR, just wanted to know.

jeromy-cannon:
@AdityaOP007 , our self hosted runners are down due to a GitHub outage, we are still waiting for them to come back online.  Assuming all of the PR checks pass we should be able to merge your PR.

AdityaOP007:
> @AdityaOP007 , our self hosted runners are down due to a GitHub outage, we are still waiting for them to come back online.  Assuming all of the PR checks pass we should be able to merge your PR.

Sure

jeromy-cannon:
@AdityaOP007 , thank you for your contribution!

AdityaOP007:
> @AdityaOP007 , thank you for your contribution!

Thank you, I really appreciate the review and guidance throughout the process. Glad to have contributed to Solo, and I’m looking forward to contributing more to the project.

PR #5531, opened 2026-07-31
feat: assert the highest migration version never exceeds the declared schema version

### LINKED ISSUE #5455: test: assert the highest migration version never exceeds the declared schema version
### Description

The created version, the schema version constant, and the latest migration have drifted apart
(create=v6, migrations→v8, const=1) and `transform()` never consults the declared version, so the
inconsistency goes undetected (`remote-config-runtime-state.ts:226`, `remote-config-schema.ts:14`).
Make sure we do this for local, remote config, cache target configuration (and any additional versioned configs)

- **Do:** add unit tests validating that the highest registered migration version is not newer than
  the declared static schema version, for both local and remote config schemas.
- **Acceptance:** adding a migration without bumping the declared version fails the test suite.

jan-milenkov:
## Description

- Bumped `RemoteConfigSchema.SCHEMA_VERSION` to 8 and `LocalConfigSchema.SCHEMA_VERSION` to 2.
- Replaced the hard-coded `6` in `RemoteConfigRuntimeState` with `RemoteConfigSchema.SCHEMA_VERSION.major`,
  so a newly created remote config is stamped with the current schema version.
- `SchemaDefinitionBase.validateMigrations()` now fails when a migration produces a version newer than
  the declared schema version.
- Fixed two pre-existing bugs in `validateMigrations()`: it threw a phantom "gap in the migration
  sequence" error for every non-empty migration list, and it sorted migration versions
  lexicographically instead of numerically.
- Added unit tests covering all versioned schema definitions (local, remote, solo), including a guard
  that fails if a new schema definition is added without being covered.

Adding a migration without bumping the declared schema version now fails the test suite.

### Related Issues

* Closes # https://github.com/hiero-ledger/solo/issues/5455

### DIFF
--- a/src/business/runtime-state/config/remote/remote-config-runtime-state.ts
+++ b/src/business/runtime-state/config/remote/remote-config-runtime-state.ts
@@ -224,7 +224,7 @@ export class RemoteConfigRuntimeState implements RemoteConfigRuntimeStateApi {
     );
 
     const remoteConfig: RemoteConfigSchema = new RemoteConfigSchema(
-      6,
+      RemoteConfigSchema.SCHEMA_VERSION.major,
       new RemoteConfigMetadataSchema(new Date(), userIdentity),
       new ApplicationVersionsSchema(cliVersion),
       [cluster],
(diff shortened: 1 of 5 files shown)


Fantomasa (approved):
LGTM

jeromy-cannon (approved):

PR #5535, opened 2026-07-31
fix: re-verify checksums of cached package downloads before reuse

### LINKED ISSUE #5500: fix: audit force=false download paths and re-verify checksums before reuse
### Description

When `force=false`, an already-present package file is trusted by existence and only fresh downloads
verify their checksum, so a truncated file from a crashed download is reused
(`package-downloader.ts:243`).

- **Do:** validate all call paths where `force` is not passed; re-verify the checksum of an existing
  file before reusing it. If `force` is genuinely being relied on as a feature somewhere, re-evaluate
  that usage.
- **Acceptance:** a truncated cached package is detected by checksum and re-downloaded rather than
  used.

ddeskov-limechain:
## Description

`PackageDownloader.fetchPackage()` in `src/core/package-downloader.ts` trusted an already present
package file by existence alone. The very first thing inside its `try` block was
`if (fs.existsSync(packageFile) && !force) { return packageFile; }`, which ran before the checksum
was even resolved, so the file was returned without ever being hashed. Only a fresh download reached
the `verifyChecksum()` call further down. A package file left truncated by a crashed or interrupted
download therefore stayed in the cache directory and was handed to every subsequent caller
indefinitely — `force` is `false` on every production call path, so nothing ever repaired it.

* Moved the cache-hit check in `fetchPackage()` to *after* the checksum resolution and made it run
  the class's existing `verifyChecksum()` helper against the already present file. A matching
  checksum still short-circuits the download, so the cache keeps working.
* On a mismatch the failure is logged as a warning, the file is removed with `fs.rmSync()`, and
  control falls through to the normal download path — which verifies the freshly downloaded file
  again, so a second corrupt fetch fails loudly through the existing
  `PackageDownloadFailedSoloError` rather than looping or being silently accepted.
* When the caller passes `verifyChecksum = false` there is nothing to compare against, so an already
  present file is still reused by existence exactly as before. This keeps the two WRAPs-library call
  sites unchanged.
* **Behaviour change worth calling out:** with `verifyChecksum = true`, the checksum is now resolved
  on the cache-hit path too (previously it was skipped entirely). If the checksum URL is unreachable
  or its contents are unreadable, `fetchPackage()` now fails instead of silently returning an
  unverified cached file. This was the least surprising option of the two available: it makes a warm
  cache behave identically to a cold one — where an unreachable checksum has always been a hard
  failure — and it keeps the rule "a file is never used unless it has been proven correct" free of
  exceptions. The consequence is that the existing `catch` cleanup will also delete the unverifiable
  cached package, which is consistent with the same cleanup deleting a package whose checksum
  mismatched.
* Updated the `@param [force]` doc comment, which previously described the old
  reuse-by-existence semantics.
* Removed the now-dead workaround in `PlatformInstaller.getPlatformRelease()`
  (`src/core/platform-installer.ts`). It existed only because of this bug — its own comment said
  *"fetchPlatform returns early on cache hit without re-downloading it"* — and re-fetched
  `build-<tag>.sha384` by hand. Since `fetchPackage()` now downloads the checksum file on the
  cache-hit path as well, that branch can never be taken and its comment had become false.

### Call-path audit

Every caller of `fetchPackage()` / `fetchPlatform()` / `fetchFile()` in `src/` and `test/`:

**`fetchPackage()` — 4 call sites**

| # | Call site | `force` | `verifyChecksum` | Conclusion |
|---|-----------|---------|------------------|------------|
| 1 | `src/core/package-downloader.ts:305` — `fetchPlatform()` delegating to `fetchPackage()` | forwarded from `fetchPlatform`, defaults to `false` | `true` (`sha384`) | **This is the vulnerable path.** Fixed at the choke point. |
| 2 | `src/core/dependency-managers/base-dependency-manager.ts:286` — `install()` downloading helm/kind/kubectl/crane/podman/gvproxy/vfkit | not passed, so `false` | `this.getVerifyChecksum()` — `true` for every manager except `VfkitDependencyManager`, which overrides it to `false` | Latently affected only. `install()` defaults its `temporaryDirectory` to `Helpers.getTemporaryDirectory()`, which is a fresh `fs.mkdtempSync()` per call, and `DependencyManager.install()` (`src/core/dependency-managers/dependency-manager.ts:97`) always calls it with no argument — so the destination is empty and the cache branch is not reached today. It is still covered by the fix if a caller ever passes a persistent directory. |
| 3 | `src/commands/network.ts:1831` — WRAPs library `.tar.gz` | explicitly `false` | explicitly `false` (checksum argument is the literal `'unusued'`) | Unaffected by design: with no checksum there is nothing to verify against, so reuse-by-existence is retained. Guarded by its own `fs.existsSync(extractedDirectory)` check. |
| 4 | `src/commands/node/tasks.ts:3577` — WRAPs library `.tar.gz` | explicitly `false` | explicitly `false` (same `'unusued'` literal) | Same as #3 — unaffected. |

**`fetchPlatform()` — 2 call sites**

| # | Call site | `force` | Conclusion |
|---|-----------|---------|------------|
| 5 | `src/core/platform-installer.ts:120` — `getPlatformRelease()`, downloading `build-<tag>.zip` into `<stagingDir>/build/` | not passed, so `false` | **The real-world exposure.** This directory is the persistent solo cache, shared across runs, so a truncated consensus-node build zip was reused forever. Now re-verified against `build-<tag>.sha384`. |
| 6 | `test/e2e/integration/core/package-downloader-end-to-end.test.ts:23` | not passed, so `false` | Test-only; benefits from the same fix. |

**`fetchFile()` — 3 non-test call sites**

| # | Call site | Conclusion |
|---|-----------|------------|
| 7 | `src/core/package-downloader.ts:252` — internal, fetches the checksum file | Unchanged. |
| 8 | `src/core/package-downloader.ts:264` — internal, fetches the package | Unchanged; still followed by `verifyChecksum()`. |
| 9 | `src/core/platform-installer.ts:127` — manual `.sha384` re-fetch workaround | **Removed** — dead after this fix (see Description). |

**Conclusion on `force`:** no call site anywhere in `src/` or `test/` passes `force = true`. The two
explicit `false` arguments (#3 and #4) are positional filler needed to reach the preceding
`verifyChecksum` and `algo` parameters, not a deliberate use of the flag. `force` is therefore not relied on as a feature
anywhere, and its semantics are left intact for API compatibility and for the existing e2e
`fetchPackage` stubs — it now means "download unconditionally, skipping the checksum check of an
already present file".

### Related Issues

* Closes #5500

- [x] This PR added tests (unit, integration, and/or end-to-end)
- [ ] This PR has no breaking changes
- [ ] Any technical debt has been documented as a separate issue and linked to this PR

### Testing

- [x] This PR added unit tests
- [ ] These changes required manual testing that is documented below

Five unit tests were added to the existing `test/unit/core/package-downloader.test.ts` under a new
`fetchPackage` describe block, covering: reuse of a matching cached file with no download, detection
and re-download of a truncated cached file (the issue's acceptance criterion), a corrupt replacement
download failing and being removed, `force = true` still downloading over a valid cached file, and
reuse-by-existence still applying when `verifyChecksum` is `false`.

```
npx tsc --noEmit
```

exit code 0, no output.

```
npx eslint src/core/package-downloader.ts src/core/platform-installer.ts test/unit/core/package-downloader.test.ts
```

exit code 0. Six warnings remain, all pre-existing and all in untouched regions of
`platform-installer.ts` (`unicorn/import-style` on the `node:path` import,
`@typescript-eslint/explicit-function-return-type` x3, `@typescript-eslint/no-explicit-any` x2). No
errors.

```
npx mocha 'test/unit/**/*.ts' --exclude 'test/unit/core/helm/**/*.ts'
```

1393 passing, 0 failing, exit code 0. The same command on `origin/main` before the change gave
1388 passing, 0 failing — so the delta is exactly the 5 added tests and there are no new failures
and no pre-existing failures.

The following was not tested:

* The e2e suites (`test/e2e/integration/core/package-downloader-end-to-end.test.ts`,
  `platform-installer-end-to-end.test.ts`, and the dependency-manager e2e tests) were not run —
  they require a Kubernetes cluster and real network downloads of platform artifacts.
* No manual end-to-end run against `builds.hedera.com` with a deliberately truncated cached
  `build-<tag>.zip`; the truncation scenario is covered by the added unit test instead, which stubs
  `fetchFile` and exercises the real hashing and comparison code.
* The changed `PlatformInstaller.getPlatformRelease()` path is only covered indirectly by the
  existing `test/unit/core/platform-installer.test.ts` `before` hook, which performs a real
  download; it passed as part of the unit run above.

---

**Update:** the `windows-2022` unit job failed all four new `fetchPackage` tests, and it exposed a
pre-existing bug rather than a bad test. `fetchPackage()` built both the package path and the checksum
path by interpolating a literal forward slash after the destination directory
(`` `${destinationDirectory}/${path.basename(packageURL)}` ``), so on Windows it returned a
mixed-separator path like `C:\Users\...\Temp\downloader-x/build-v0.0.1.zip`. Every `fs` call accepts
that, which is why it went unnoticed — the new tests are simply the first thing to compare the
returned value against a `PathEx.join`-built path.

A follow-up commit joins both paths with `PathEx.join`, matching how the rest of the codebase builds
paths, so the returned value is a native path on every platform. Re-verified on macOS:
`npx tsc --noEmit` clean, `npx eslint` and `npx prettier --check` clean on the changed file,
`npx dpdm` reports no circular dependency from the new import, and the unit suite is 1393 passing /
0 failing. The Windows job itself can only be confirmed by CI.

### DIFF
--- a/src/core/platform-installer.ts
+++ b/src/core/platform-installer.ts
@@ -119,7 +119,7 @@ export class PlatformInstaller {
     }
     const zipPath: string = await this.packageDownloader.fetchPlatform(tag, buildDirectory);
 
-    // fetchPlatform always downloads the checksum file, both on a fresh download and on a cache hit
+    // fetchPlatform ensures the checksum file is present, downloading it only when it is missing or unusable
     const checksumPath: string = PathEx.join(buildDirectory, `build-${tag}.sha384`);
 
     return [zipPath, checksumPath];
--- a/src/core/platform-installer.ts
+++ b/src/core/platform-installer.ts
@@ -119,13 +119,8 @@ export class PlatformInstaller {
     }
     const zipPath: string = await this.packageDownloader.fetchPlatform(tag, buildDirectory);
 
-    // Ensure the checksum file is also present (fetchPlatform returns early on cache hit without re-downloading it)
+    // fetchPlatform always downloads the checksum file, both on a fresh download and on a cache hit
     const checksumPath: string = PathEx.join(buildDirectory, `build-${tag}.sha384`);
-    if (!fs.existsSync(checksumPath)) {
-      const releaseDirectory: string = Templates.prepareReleasePrefix(tag);
-      const checksumURL: string = `${constants.HEDERA_BUILDS_URL}/node/software/${releaseDirectory}/build-${tag}.sha384`;
-      await this.packageDownloader.fetchFile(checksumURL, checksumPath);
-    }
 
     return [zipPath, checksumPath];
   }
--- a/src/core/package-downloader.ts
+++ b/src/core/package-downloader.ts
@@ -206,6 +206,38 @@ export class PackageDownloader {
   }
 
   /**
+   * Download a checksum file and return the checksum it contains
+   * @param checksumURL - URL of the checksum file
+   * @param checksumFile - path where the checksum file should be saved
+   * @returns the checksum read from the downloaded file
+   */
+  private async fetchChecksum(checksumURL: string, checksumFile: string): Promise<string> {
+    await this.fetchFile(checksumURL, checksumFile);
+    const checksumData: string = fs.readFileSync(checksumFile).toString();
+    if (!checksumData) {
+      throw new SoloErrors.system.checksumReadFailed(checksumFile);
+    }
+    return checksumData.split(' ', 1)[0];
+  }
+
+  /**
+   * Read a previously downloaded checksum file
+   * @param checksumFile - path of the cached checksum file
+   * @param algo - hash algorithm the checksum is expected to be for
+   * @returns the checksum, or undefined when the file is absent or not a well-formed digest for the algorithm
+   */
+  private readCachedChecksum(checksumFile: string, algo: string): string | undefined {
+    try {
+      const checksum: string = fs.readFileSync(checksumFile).toString().split(' ', 1)[0].trim();
+      const digestHexLength: number = crypto.createHash(algo).digest().length * 2;
+      return new RegExp(`^[0-9a-f]{${digestHexLength}}$`, 'i').test(checksum) ? checksum : undefined;
+    } catch {
+      // best-effort: an absent or unreadable cached checksum file simply means it must be downloaded
+      return undefined;
+    }
+  }
+
+  /**
    * Fetch a remote package
    * @param packageURL
    * @param checksumDataOrURL - package checksum URL or checksum data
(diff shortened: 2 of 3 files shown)


jan-milenkov (commented):
LGTM @JeffreyDallas @jeromy-cannon please review too

src/core/platform-installer.ts
```diff
@@ -119,13 +119,8 @@ export class PlatformInstaller {
     }
     const zipPath: string = await this.packageDownloader.fetchPlatform(tag, buildDirectory);
 
-    // Ensure the checksum file is also present (fetchPlatform returns early on cache hit without re-downloading it)
+    // fetchPlatform always downloads the checksum file, both on a fresh download and on a cache hit
```
jeromy-cannon: It seems that if we have the checksum file and the downloaded file, and the checksum matches there would be no necessity to redownload the checksum file which is a point of failure outside of our control if the user cannot establish the connection to the remote host or if the download is interrupted.

ddeskov-limechain: Fixed in c6676783 . A cache hit now verifies against the cached checksum file and only re-downloads it when it's missing, malformed, or mismatching.

jan-milenkov (approved):

PR #5573, opened 2026-08-03
test: add unit tests covering all flag prompts

### LINKED ISSUE #1503: test all of the prompts with unit tests
need to test all of the prompts with unit tests, they are breaking and we aren't realizing it with refactors, eg. values file got updated to be comma delimited but the prompt will throw an error if the string is not a file exists true for the entire parameter

Fantomasa:
## Description

* Adds `test/unit/commands/flag-prompts.test.ts` (229 tests), unit-testing all 67 flag prompt functions defined in `src/commands/flags.ts`:
  * **39 text prompts, 16 toggle prompts, 5 number prompts** — table-driven: each flag asserts that a provided value is returned without prompting, that the prompt fires with the expected prompt component, message, and default when the value is missing, and that a missing value without a TTY fails with `FlagInputFailedSoloError` instead of hanging.
  * **Empty-answer rejection** for the prompts that declare an empty-check message (`cluster-ref`, `cluster-setup-namespace`, `namespace`, `relay-release`, `file-id`, `file-path`).
  * **Dedicated suites for the bespoke prompts:** `num-consensus-nodes` (re-prompt loop), both `values-file` variants — including the comma-separated multi-cluster form whose breakage motivated the original issue — plus direct `Flags.parseValuesFilesInput` coverage, `chart-dir` (exists-check, replacement re-prompt, invalid-directory rejection), `tls-cluster-issuer-type` (select choices), `user` (alphanumeric re-prompt loop and `validate`), and `context` (select from the provided contexts).
  * **A completeness guard test** that fails whenever a flag in `Flags.allFlags` defines a `prompt` function without being covered by this suite, so future prompts cannot ship untested — which is the failure mode described in the issue.

Current-behavior quirks surfaced while writing the tests. They are pinned as-is (behavior changes are out of scope here) and can be split into follow-up issues:

* `generateGossipKeys` interpolates the raw input into its prompt message, producing `Would you like to generate Gossip keys? undefined undefined` — looks like leftover debug output.
* `tlsClusterIssuerType`'s prompt returns `undefined` when a value **is** provided, and `ConfigManager.executePrompt` then overwrites the provided value with that `undefined` via `setFlag`.
* `username`'s prompt always prompts, even when a username was provided.
* `numberOfConsensusNodes`'s prompt loops forever when the user enters `0`: the re-prompt loop treats `0` as falsy, but the inner prompt sees a number and returns it without prompting again. Not testable without hanging the suite, so it is only noted here.
* `outputDir` / `inputDir` are string flags whose prompts use the toggle (confirm) flow, so a provided string path triggers a prompt attempt instead of being accepted.

### Related Issues

* Closes #1503

### Testing


test/unit/commands/flag-prompts.test.ts
jan-milenkov, line 1: I think `textPromptCases` is bloat, all data should be able be on the `CommandFlag`, what i would suggest if you need to access to the prompt text, add it to the definition on the `CommandFlag`

```ts
export interface CommandFlag {
  constName: string;
  name: string;
  definition: Definition;
  prompt: PromptFunction;
  validate?: (input: any) => boolean;
}
```

```ts
export interface Definition {
  describe: string;
  defaultValue?: boolean | string | number;
  alias?: string | string[];
  type?: string;
  disablePrompt?: boolean;
  dataMask?: string;
  promptText?: string; // <--- example
}
```

jan-milenkov (approved):

PR #5590, opened 2026-08-03
feat: bump block & mirror & relay node version

JeffreyDallas:
## Description

* block & mirror & relay node version

### Related Issues

* Closes #
* Related to #

### Testing

* TBD

The following was not tested:

* TBD

<details>
<summary>
Commit message guidelines
</summary>

</details>


jeromy-cannon:
@JeffreyDallas , looks like the test migration process is having issues

jeromy-cannon:
@JeffreyDallas would you add a note about what will unblock it?

.github/workflows/script/launch_network.sh
```diff
@@ -158,6 +158,65 @@ show_service_ips() {
   kubectl get svc -n "${namespace}" network-node1-svc network-node2-svc -o custom-columns=NAME:.metadata.name,CLUSTER-IP:.spec.clusterIP,CREATED:.metadata.creationTimestamp
 }
 
+get_latest_mirror_block_number() {
```
jeromy-cannon: **Duplicates functions already shared via `helper.sh` (sourced at the top of this script), with a signature that diverges just enough to break an existing call site.**

`helper.sh` (merged in #5358) already defines `get_latest_mirror_block_number`, `wait_for_mirror_block_progress`, and `wait_for_mirror_block_count_progress`, and this script already `source`s it (line 6). Re-declaring the same function names here shadows the sourced versions for the remainder of the script (bash: last definition wins).

Concretely, the local `wait_for_mirror_block_progress` added here (line 184) drops the `required_new_blocks` parameter that the canonical `helper.sh` version has:

```
# helper.sh (canonical, still sourced above)
wait_for_mirror_block_progress(label, previous_block=-1, required_new_blocks=3, max_attempts=120, sleep_seconds=2)

# this file's new local shadow
wait_for_mirror_block_progress(label, previous_block=-1, max_attempts=90, sleep_seconds=2)
```

The pre-existing, unmodified call at line 739 —
`wait_for_mirror_block_progress "source deployment after one-shot" -1 1 90 2 > /dev/null`
— was written against the canonical 5-arg signature (`required_new_blocks=1, max_attempts=90, sleep_seconds=2`). After this shadowing, it silently binds `max_attempts=1, sleep_seconds=90` instead, turning a "retry up to 90 times over ~3 minutes" wait into a single check. With `set -eo pipefail` active, a mirror node that isn't ready yet will now abort the whole job instead of retrying.

Suggest deleting this whole block (`get_latest_mirror_block_number`, `wait_for_mirror_block_progress`, `wait_for_mirror_block_count_progress`) and relying on the already-sourced `helper.sh` versions, which the other call sites in this file (lines 860, 956) already assume.

.github/workflows/script/solo_smoke_test.sh
```diff
@@ -145,6 +145,56 @@ EOF
   done
 }
 
+function get_latest_mirror_block_number ()
```
jeromy-cannon: Same duplication pattern as `launch_network.sh` (see that file's comment): `helper.sh` is already `source`d here (line 13) and already defines `get_latest_mirror_block_number` and `wait_for_mirror_block_progress` with matching semantics — except this local copy swaps `return 1` for `log_and_exit 1` on timeout. The existing call at line 467 (`wait_for_mirror_block_progress "before smart contract test" ...`) doesn't check the return code, so this changes an unmodified call site from "log and continue" to "dump logs and hard-exit the script" on a mirror-block timeout. That may be the desired behavior for a smoke test, but it should be a deliberate change to the shared `helper.sh` (so both scripts pick it up consistently) rather than a local shadow copy.

src/commands/node/tasks.ts
```diff
@@ -239,6 +239,34 @@ export class NodeCommandTasks {
     return gossipFqdnRestricted || Helpers.hasMultipleKubernetesContexts(consensusNodes);
   }
 
+  private static hasMultipleKubernetesContexts(consensusNodes: ConsensusNode[]): boolean {
```
jeromy-cannon: `hasMultipleKubernetesContexts` and `buildRsaAddressBookHistory` (through line 272) duplicate `Helpers.hasMultipleKubernetesContexts` and `Helpers.buildRsaAddressBookJson` (`src/core/helpers.ts:609`, `:622`) verbatim, and neither new method is called anywhere in this file or the rest of the diff — this is unused dead code. Please remove both; the existing `Helpers` methods are already used elsewhere in this file (e.g. line 239, and `Helpers.hasMultipleKubernetesContexts` at the load-balancer check further down).

src/commands/block-node.ts
```diff
@@ -454,17 +455,25 @@ export class BlockNodeCommand extends BaseCommand {
       return undefined;
     }
 
+    const nodeAddresses: Array<{RSAPubKey: string; nodeId: number}> = [];
```
jeromy-cannon: This re-implements `Helpers.buildRsaAddressBookJson` (`src/core/helpers.ts:622`) inline rather than extending it, merging in the `fs.existsSync` check that used to run in a separate pass. That's a reasonable goal (avoid reading each key file twice), but now the cert->SPKI->hex JSON-building logic exists in two places that must be kept in sync. Consider generalizing `Helpers.buildRsaAddressBookJson` (e.g. return `undefined` when a key file is missing, or accept a pre-checked existence flag) instead of duplicating its body here.

src/commands/one-shot/orchestrator/deploy/deploy-argv-builders.ts
```diff
@@ -67,7 +67,12 @@ export class DeployArgvBuilders {
     if (!this.shouldDeployBlockNode(config)) {
       return false;
     }
+
     const consensusNodeVersion: string = config.versions.consensus || version.HEDERA_PLATFORM_VERSION;
+    if (new SemanticVersion<string>(consensusNodeVersion).lessThan(version.MINIMUM_HIERO_PLATFORM_VERSION_FOR_TSS)) {
```
jeromy-cannon: `Helpers.requiresRsaBootstrap` (`src/core/helpers.ts:614`) already returns `false` when `consensusNodeVersion` is below `MINIMUM_HIERO_PLATFORM_VERSION_FOR_TSS`, before it even looks at `streamMode`. Adding the same `SemanticVersion(...).lessThan(MINIMUM_HIERO_PLATFORM_VERSION_FOR_TSS)` check here can't change the return value of `shouldSkipMinioSetup` — it's redundant with logic that already lives in `Helpers.requiresRsaBootstrap`. Same duplication at line 86 in `shouldInjectRsaBootstrapValuesFile`. Suggest removing both added blocks.

src/commands/node/handlers.ts
```diff
@@ -1253,7 +1255,7 @@ export class NodeCommandHandlers extends CommandHandler {
         this.tasks.identifyExistingNodes(),
         this.tasks.sendFreezeTransaction(),
         this.tasks.checkAllNodesAreFrozen('existingNodeAliases'),
-        this.tasks.drainBlockStreamAfterFreeze(),
+        this.tasks.sleep('Drain block stream after freeze', NodeCommandHandlers.FREEZE_BLOCK_STREAM_DRAIN_MILLISECONDS),
```
jeromy-cannon: This replaces `this.tasks.drainBlockStreamAfterFreeze()` with a hardcoded `this.tasks.sleep(..., FREEZE_BLOCK_STREAM_DRAIN_MILLISECONDS)` (20s) here in `freeze()`, but `upgradeExecuteTasks()` (line 452, unchanged) still calls `this.tasks.drainBlockStreamAfterFreeze()`, which reads a configurable `context_.config.freezeBlockDrainSeconds ?? 20`. That means:
1. `freeze` silently loses the `freezeBlockDrainSeconds` config override that `upgrade` still honors.
2. There are now two different implementations of "drain block stream after freeze" in this codebase with the same title string, which will drift over time.

If `drainBlockStreamAfterFreeze()`'s `NodeUpgradeContext` typing doesn't fit `freeze()`'s context, consider generalizing that task (e.g. widen its context type, or take the drain-seconds value as a parameter) rather than adding a parallel hardcoded path.

.github/workflows/script/launch_network.sh
JeffreyDallas: Fixed in fa3f0e8d2: removed the three local definitions so this script uses the sourced helper.sh implementations and their five-argument signatures.

.github/workflows/script/solo_smoke_test.sh
JeffreyDallas: Fixed in fa3f0e8d2: removed the local copies and now use the shared helper.sh behavior consistently.

src/commands/node/tasks.ts
JeffreyDallas: Fixed in fa3f0e8d2: removed both unused duplicate methods; existing Helpers methods remain the single implementation.

src/commands/block-node.ts
JeffreyDallas: Fixed in fa3f0e8d2: generalized Helpers.buildRsaAddressBookJson() to return undefined when a public key is missing, and BlockNodeCommand now consumes that shared implementation.

src/commands/one-shot/orchestrator/deploy/deploy-argv-builders.ts
JeffreyDallas: Fixed in fa3f0e8d2: removed both redundant version guards and rely on Helpers.requiresRsaBootstrap() as the single source of that check.

src/commands/node/handlers.ts
JeffreyDallas: Fixed in fa3f0e8d2: generalized drainBlockStreamAfterFreeze() for both freeze and upgrade contexts, so freeze again honors freezeBlockDrainSeconds.

jeromy-cannon (approved):

PR #5612, opened 2026-08-04
feat: Enhance one-shot falcon to detect JSON values file

### LINKED ISSUE #3295: Enhance one-shot falcon to detect JSON values file
Enhance one-shot falcon to detect if falcon values file is json, maybe starts with '{'.  if so, then use json reader instead of yaml.  then update help docs and add test case to verify

jan-milenkov:
## Description

If the content starts with `{`, it was written as JSON, so parse it with `JSON.parse` and hold it to
JSON's rules. Everything else still goes through YAML.

### Related Issues

* Closes # https://github.com/hiero-ledger/solo/issues/3295


JeffreyDallas (dismissed):

src/core/util/values-file-parser.ts
```diff
@@ -12,16 +12,23 @@ import {ValuesFileParseFailedSoloError} from '../errors/classes/validation/value
...
-   * @throws ValuesFileParseFailedSoloError when the content is not valid YAML
+   * @throws ValuesFileParseFailedSoloError when the content is neither valid JSON nor valid YAML
    */
   public static parse(valuesFilePath: string, content: string): unknown {
+    // Trimmed because JSON.parse rejects the leading byte-order mark that Windows editors emit.
+    const trimmedContent: string = content.trimStart();
+
     try {
-      return yaml.parse(content);
+      return trimmedContent.startsWith('{') ? JSON.parse(trimmedContent) : yaml.parse(content);
```
jeromy-cannon, line 31: YAML flow-mapping files in the form `{hedera: {nodeId: 0}}` are valid YAML — they parsed correctly before this change and will now hit `JSON.parse` first, failing on unquoted keys with a generic "Unexpected token" message. Worth noting in the troubleshooting steps so users whose YAML file happens to start with `{` know to switch to block style or quote their keys.

src/core/errors/classes/validation/values-file-parse-failed-solo-error.ts
```diff
@@ -20,7 +20,7 @@ export class ValuesFileParseFailedSoloError extends SoloError {
         message: `Failed to parse values file ${valuesFilePath}: ${cause.message}`,
         code: ErrorCodeRegistry.VALUES_FILE_PARSE_FAILED,
         troubleshootingSteps:
-          `Open ${valuesFilePath} and correct the YAML syntax reported above\n` +
+          `Open ${valuesFilePath} and correct the syntax reported above\n` +
```
jeromy-cannon, line 23: consider adding a sentence here about the JSON detection so the error is self-explaining when a previously-working YAML file in flow-mapping style (starting with `{`) is now rejected:

```suggestion
          `Open ${valuesFilePath} and correct the syntax reported above\n` +
          `Note: values files starting with '{' are parsed as strict JSON — if your file is YAML flow-mapping style, rewrite it in block style or quote all keys\n` +
```

src/core/util/values-file-parser.ts
jan-milenkov: updated

src/core/errors/classes/validation/values-file-parse-failed-solo-error.ts
jan-milenkov: updated

jeromy-cannon (approved):

PR #5620, opened 2026-08-04
feat: add validate property in CommandFlag that will take a validation function

### LINKED ISSUE #1526: add validate property in CommandFlag that will take a validation function
add flags.ts validate properties to `CommandFlag`,`validate: (value: any) => boolean; // everything from the user should be validated`, we should be validating the podName for compliance: contain at most 63 characters, contain only lowercase //  - alphanumeric characters or '-', start with an alphanumeric character, end with an alphanumeric character. // - https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

jan-milenkov:
## Description

A flag declares any number of rules; a rule reports why a value is unacceptable, or `undefined`:

```ts
type FlagRule = (value: string) => string | undefined;
```

### Related Issues

* Closes # https://github.com/hiero-ledger/solo/issues/1526


src/core/middlewares.ts
```diff
@@ -202,6 +203,9 @@ export class Middlewares {
       // apply precedence for flags
       argv = configManager.applyPrecedence(argv, yargs.parsed.aliases);
 
+      // Before update(), which coerces some values and would report its own error instead of a flag-scoped one.
+      FlagValidation.assertAllValid(flags.allFlags, argv);
```
Fantomasa, line 316: `assertAllValid` only runs on the explicit `--deployment` / `--cluster-ref` path. when the flag is omitted the value is resolved from local config and `setFlag`'d after this middleware (see `resolvers.ts`), so it skips validation entirely. net effect: a non-DNS name still gets in through the picker, while `--deployment <that same name>` is rejected here — so the rule doesn't actually enforce the invariant, but it does break the explicit path for names we accepted before. was the asymmetry intended?

Fantomasa (dismissed):

src/core/middlewares.ts
jan-milenkov: Selection lists existing deployments intentionally. `executePrompt` was already validated; the free-text prompt now validates too.

src/commands/validation/flag-rules.ts
```diff
@@ -0,0 +1,64 @@
...
+    /^node[1-9]\d*$/.test(value) ? undefined : "must be a node alias of the form 'node<number>', such as 'node1'";
+
+  /** A solo-local label rather than a Kubernetes name, so underscores are allowed. */
+  public static readonly clusterReference: FlagRule = (value: string): string | undefined =>
+    /^[a-z0-9_-]+$/.test(value) ? undefined : "must contain only lowercase alphanumeric characters, '-' or '_'";
+
+  public static readonly alphanumeric: FlagRule = (value: string): string | undefined =>
+    validator.isAlphanumeric(value) ? undefined : 'must contain only letters and numbers';
+
+  public static readonly integer: FlagRule = (value: string): string | undefined =>
```
jeromy-cannon, line 36: `Number.isInteger(Number(value))` accepts scientific notation and a trailing period: `'1e2'` parses as `100` (integer), `'1.'` parses as `1` (integer). Users are unlikely to type these for a replica count, but the error message promises "must be a whole number" and these aren't.

```suggestion
  public static readonly integer: FlagRule = (value: string): string | undefined =>
    /^-?\d+$/.test(value.trim()) ? undefined : 'must be a whole number';
```

test/unit/commands/flag-prompts.test.ts
```diff
@@ -143,6 +149,11 @@ function expectedPromptDefault(flag: CommandFlag): unknown {
   return flag.definition.promptDefaultValue ?? flag.definition.defaultValue;
 }
 
+/** An answer that satisfies each flag's own rules; flags without rules accept any string. */
+function acceptableAnswer(flag: CommandFlag): string {
+  return FlagValidation.violationOf(flag, 'answered-value') ? 'node1' : 'answered-value';
```
jeromy-cannon: `'node1'` is a valid DNS label, node alias, cluster reference, and alphanumeric string — so it happens to satisfy every rule set used by flags that currently have both `prompt` and `promptText`. But it would not satisfy `[FlagRules.integer, FlagRules.atLeast(1)]`. If a future flag adds integer rules and `promptText`, this falls into infinite re-ask in `FakePromptTaskWrapper` (no second answer queued).

Consider a two-entry fallback list so one answer satisfies non-integer flags and another satisfies integer-type flags, or document explicitly that the fallback is intentionally `'node1'` and why it covers all current cases.

jan-milenkov: updated

src/commands/validation/flag-rules.ts
jan-milenkov: updated

jeromy-cannon (approved):

PR #5767, opened 2026-08-07
feat: add image cache manifest download and parsing module

ddeskov-limechain:
## Description

Stacked on #5766; merge in order B1 → B2 → B3. The diff against `main` will show B1's commit too until that merges — review the `feat/5748-cache-manifest-module` commit only.

`solo cache image pull` needs to know, for the exact Solo version that is running, which image archives to download, where they live on the CDN, and what each one should hash to. This adds the module that answers that: it downloads `cache-manifest.json` from the matching Solo release, validates it, and hands back a typed list with the CDN URLs already resolved. #5749 (download), #5750 (prune), and #5751 (load re-validation) all consume it.

#5742 (A1) has not defined the manifest schema yet, so this PR defines a concrete minimal one that satisfies the epic's constraints. It is posted on #5742 for review; if A1 lands on something different, this module is the only place that has to change.

**Manifest schema (v1)**

```json
{
  "schemaVersion": 1,
  "soloVersion": "0.86.0",
  "images": [
    {
      "image": "docker.io/library/busybox:1.36.1",
      "tarFile": "docker.io__library__busybox__1.36.1.tar",
      "hashFile": "docker.io__library__busybox__1.36.1.tar.sha256",
      "sha256": "3f57d9401f8d42f986df300f0c69192fc41da28ae8c15f9f5c0a4b1f1d59f7cf"
    }
  ]
}
```

* `schemaVersion` — integer, currently `1`. Lets a future manifest change be rejected cleanly by an older CLI instead of misparsed.
* `soloVersion` — the Solo version the manifest was generated for. Validated against the version that requested it, so a mis-published asset is caught rather than acted on.
* `images[].image` — the full image reference, registry, container and tag included.
* `images[].tarFile` — bare file name of the archive on the CDN, with registry, container and version embedded so it is unique across versions and safe in a flat bucket. The `__` separator matches the convention `FileSystemCacheCatalogStore` already uses for local cache file names.
* `images[].hashFile` — bare file name of the file holding the archive's hash. Carried explicitly rather than derived from `tarFile` so the extension convention can change without a CLI release, and because the epic asks for the hash filename to be part of the manifest.
* `images[].sha256` — the archive's SHA-256 as 64 lowercase hex characters. Both this and the contents of `hashFile` are checked in #5749, and the two must agree.

CDN layout is flat: `<base>/<tarFile>` and `<base>/<hashFile>`.

### What changed

[...]

PR #5766, opened 2026-08-07
refactor: remove registry-pull code paths from solo cache image commands

ddeskov-limechain:
## Description

First of a three-PR stack; merge in order B1 → B2 → B3 (this PR, then #5748's PR, then #5749's PR).

The image cache is moving to a CDN-download model (epic #5738): `solo cache image pull` will download prebuilt image archives and their SHA-256 files from `cdn.solo.hashgraph.io`, verify them, and load them into the cluster. Nothing other than `solo cache image pull` may write into the local image cache.

This PR is step B1 — the removal half. It deletes every code path that pulls container images from a registry into the local image cache, and everything that existed only to serve those paths.

**This PR on its own leaves `solo cache image pull` non-functional.** It becomes a single task that reports "skipped, CDN downloads are not available yet" and records that message in the end-of-run image cache summary. `solo cache image load` then finds no archives and skips each one (already a non-fatal path), so the cluster pulls whatever it is missing directly from the registries — the fallback the epic describes. `solo cache image pull` is restored by #5749 (B3), which lands on top of this branch. **Please merge B1 → B2 → B3 together or in quick succession.**

Kind config mirror registry entries in `resources/kind-config.yaml` are deliberately left in place, as the issue requires.

### What changed

* `src/integration/cache/impl/image-cache-handler.ts` — dropped `pull()`'s registry-save logic and everything under it: Docker Hub mirror-registry candidate resolution (`KIND_DOCKER_REGISTRY_MIRRORS`, the kind-config `containerdConfigPatches` scraper, the hardcoded Hashgraph mirror fallback), the rate-limit detection, and the crane-backed `saveImage`. Also removed `pullKindNodeImageIfMissing()` / `loadKindNodeImageIntoEngine()`. `pull()` is now an explicit, non-fatal "not available yet" task. `load()`, `clear()`, `healthcheck()`, `list()`, and `resolveRequiredArtifacts()` are unchanged.
* `src/core/cluster-task-manager.ts` — removed the `ENABLE_IMAGE_CACHE` block in `defaultCreateClusterTask` that saved the Kind node image into the Solo cache and loaded it back into the local engine. This was a registry pull writing into `~/.solo/cache`, which the epic forbids outside `cache image pull`. Kind now pulls its node image itself, and Docker's own image store already caches it locally across cluster recreates, so the only capability lost is surviving a `docker image prune`. The `ContainerEngineClient` injection became unused and was removed with it.
[...]

Issue #5751, opened 2026-08-06
[Image Cache] B5: Re-validate image tar hash before loading into cluster


jeromy-cannon:
## Overview
Before loading a cached image tar into a Kind cluster, re-run the SHA-256 check against the manifest's expected hash (via B2). Abort loading if the file does not match — this guards against tars that were corrupted or tampered with after download.

## Sub-Tasks
- [ ] Re-compute SHA-256 of local tar at cluster-load time
- [ ] Compare against manifest hash value (via B2)
- [ ] Abort cluster load and surface a clear error on hash mismatch
- [ ] Prompt the user to re-run `solo cache image pull` to redownload the affected file

### Dependencies
- [ ] requires #5749 — download-with-validation provides the initially verified tars

## References
- [Epic: Solo Image Cache Enhancements](https://github.com/hiero-ledger/solo/issues/5738)

Issue #5750, opened 2026-08-06
[Image Cache] B4: Auto-prune stale cache files not in current manifest


jeromy-cannon:
## Overview
After resolving the current Solo version's manifest (B2), automatically delete any local tar or `.sha256` files that are not listed in it. This prevents stale files from prior Solo versions accumulating on the user's filesystem over time.

## Sub-Tasks
- [ ] List all tars and hash files in the local image cache directory
- [ ] Compare against manifest entries for the current Solo version (B2)
- [ ] Delete any files not referenced in the current manifest
- [ ] Log each pruned file clearly so users understand what was removed

### Dependencies
- [ ] requires #5748 — manifest provides the authoritative list of current-version files
- [ ] required by #5752 — prune logic must be in place before migration runs

## References
- [Epic: Solo Image Cache Enhancements](https://github.com/hiero-ledger/solo/issues/5738)

PR #5768, opened 2026-08-07
feat: download image cache archives from the CDN with SHA-256 validation

### LINKED ISSUE #5749: [Image Cache] B3: CDN download with SHA-256 validation
## Overview
Pull each image tar and its `.sha256` file from `cdn.solo.hashgraph.io` during `solo cache image pull`. Verify the downloaded tar against the hash from the manifest (B2). On mismatch, delete the tar and mark for redownload on the next pull. On match, skip re-downloading. This is the only path that adds files to the local image cache.

## Sub-Tasks
- [ ] Download tar and `.sha256` from CDN URLs resolved via manifest module (B2)
- [ ] Verify tar SHA-256 against the manifest's expected hash value
- [ ] Validate that the `.sha256` file value and the manifest hash agree (dual-check)
- [ ] On hash mismatch: delete tar, log warning, trigger redownload on next pull
- [ ] On hash match: skip download entirely if file is already present and verified

### Dependencies
- [ ] requires #5748 — manifest module resolves CDN URLs and expected hashes
- [ ] required by #5751 — validated tars are prerequisites for cluster-load re-validation
- [ ] required by #5752 — download/validate logic must exist before migration can succeed

## References

ddeskov-limechain:
## Description

Stacked on #5767; merge in order B1 → B2 → B3. The diff against `main` will show the B1 and B2 commits too until they merge — review the `feat/5749-cdn-download-validation` commit only.

This restores `solo cache image pull`, which #5766 (B1) left as a stub. It resolves each required image through the manifest module added in #5767, downloads the image archive and the archive's hash file from `cdn.solo.hashgraph.io`, and writes into the local image cache only when everything checks out. Combined with B1, `solo cache image pull` is now the only code path in Solo that adds a file to the local image cache.

**Verification is three-way.** For each image the manifest carries a SHA-256, and the CDN publishes the same hash in a separate `.sha256` file. An archive is accepted only when the manifest hash, the contents of the published hash file, and the SHA-256 computed over the downloaded bytes all agree. The published hash file is fetched *before* the archive: if it disagrees with the manifest there is no value that could be trusted to verify the archive against, so the (large) download is not started at all.

**On disagreement** the archive and its hash file are deleted and a warning is logged, per the issue: the run is not failed and the archive is not retried in the same run, so the next `solo cache image pull` downloads it again.

**On a cache hit** an archive already on disk is rehashed and, if it still matches the manifest, the download is skipped. Rehashing rather than trusting a stored "verified" flag is what makes an archive corrupted since the last pull get caught and discarded, which is the behaviour the epic asks for. It costs one sequential read of the archive per pull.

Nothing in this path aborts the run. A missing manifest, an image the manifest does not list, or a failed download is reported in the end-of-run image cache summary and the image is left for the cluster to pull from its registry — the fallback the epic describes. That also means a Solo version whose release has no `cache-manifest.json` yet (every version today, until #5745/#5747 land) degrades to exactly the behaviour on `main` after B1, rather than failing `solo one-shot single deploy`.

### What changed

  * `pull()` fetches the manifest once, indexes it by full image reference, and emits one task per required target. A target with no manifest entry is skipped with a recorded reason. A manifest that cannot be fetched or parsed yields a single "skipped, no image cache manifest available" task rather than an exception.
  * `downloadArchive()` — creates the cache directory, downloads `<tarFile>.sha256` then the archive, and runs the checks described above. The hash file is kept next to the archive so a later re-validation (#5751) does not need another download.
  * `verifyCachedArchive()` — the cache-hit path: rehash, skip on match, discard on mismatch.
  * `discardArchive()` — removes the archive and its hash file together, so a discarded entry never leaves a stale hash behind. `clear()` now uses it too, which fixes hash files being orphaned by `solo cache image clear`.
  * `computeSha256()` — `node:crypto` `createHash` fed by a read stream through `stream/promises` `pipeline`, so archive size does not drive memory use.
  * The handler now takes an injected `PackageDownloader`, whose `fetchFile()` already streams to disk with `got`, follows redirects, and applies the configured connect/response timeouts. No new download code was written.
* `test/unit/integration/cache/image-cache-pull.test.ts` — new, 8 tests. `CacheManifestClient.fetchImages` is stubbed and `PackageDownloader.fetchFile` is replaced by a fake CDN that writes canned bytes to the destination path, so the tests exercise real hashing, real file writes and real deletions against a temp directory without any network.
* `test/unit/integration/cache/image-cache-handler.test.ts` — dropped B1's placeholder pull test, superseded by the file above.

Design note on the skip rule: the manifest is keyed by full image reference and matched against the targets rendered from `resources`/flags, so a component version supplied by a flag that the release manifest never saw simply is not cached. Making the manifest the authoritative image list instead is #5750's job (prune), not this PR's.

### Related Issues

* Fixes #5749
* Related to #5738, #5748

On documentation: no user-facing docs describe the image cache internals, and the one new environment variable (`SOLO_CACHE_CDN_BASE_URL`) is introduced in #5767.

### Testing

The following was run:

* `npx tsc --noEmit` — exit 0
* `npx mocha 'test/unit/**/*.ts' --exclude 'test/unit/core/helm/**/*.ts'` — **1717 passing, 0 failing** (1710 on the B2 branch: 8 added, 1 placeholder removed)

Unit coverage: fresh download of archive and hash file with both verified; skip when a valid archive is already cached; a cached archive whose hash no longer matches is deleted; a downloaded archive that fails its hash check is deleted along with its hash file; manifest and published hash disagreeing aborts before the archive is fetched; an image absent from the manifest is skipped; an unavailable manifest caches nothing without throwing; a failed download is recorded without throwing.

The following was not tested:

* No live download. Nothing has been fetched from `cdn.solo.hashgraph.io` — the CDN is still being provisioned (RESD-1110) and no `cache-manifest.json` has been published to a release yet, so the end-to-end path has never run against real artifacts.
* No E2E run: `solo cache image pull` followed by `solo cache image load` into a real Kind cluster was not exercised, nor was `solo one-shot single deploy`.
* Resume of a partially written archive is not handled — an interrupted download leaves a short file that the next pull rehashes, fails, deletes, and re-fetches on the pull after that. Correct, but it costs an extra pull; worth revisiting if it proves annoying in practice.
* Concurrency is whatever `CACHE_IMAGE_MAX_CONCURRENCY` already sets for the pull task list; no separate download throttling was added and disk throughput under parallel downloads was not measured.


jan-milenkov (commented):

jan-milenkov (dismissed):

jan-milenkov (dismissed):

Issue #5752, opened 2026-08-06
[Image Cache] B6: Migration from old image cache model to new CDN-based model


jeromy-cannon:
## Overview
Detect old-format cached files (from the previous registry-pull model) and properly prune them under the new CDN-based layout. Ensures existing users do not hit errors on their first upgrade to the new caching model. Runs after B3 (download) and B4 (prune) are in place.

## Sub-Tasks
- [ ] Detect old-format cache files by name pattern or metadata
- [ ] Remove old-format files not compatible with the manifest-driven model
- [ ] Log migration actions clearly so users understand what changed on upgrade
- [ ] Verify the new cache is correctly populated after migration completes (B3 + B4)

### Dependencies
- [ ] requires #5744 — registry-pull code must be removed before migration can be clean
- [ ] requires #5749 — new download/validate logic must exist for migration to populate cache correctly
- [ ] requires #5750 — prune logic must be in place before migration runs

## References
- [Epic: Solo Image Cache Enhancements](https://github.com/hiero-ledger/solo/issues/5738)

Issue #5748, opened 2026-08-06
[Image Cache] B2: Manifest download and parsing module


jeromy-cannon:
## Overview
Given the current running Solo version, fetch the manifest JSON from GitHub releases (`https://github.com/hiero-ledger/solo/releases/download/<version>/cache-manifest.json`) and expose the image list, CDN URLs, and expected hashes to the rest of the CLI via a typed interface. This module is shared by the download, prune, and validation commands.

## Sub-Tasks
- [ ] Implement manifest fetch from GitHub releases keyed by Solo version
- [ ] Parse manifest JSON per A1 schema (image list, CDN URLs, hash values)
- [ ] Expose typed interface consumed by B3 (download), B4 (prune), and B5 (load validation)
- [ ] Add environment variable to override CDN base URL for staging/testing

### Dependencies
- [ ] requires #5742 — manifest schema definition is needed for parsing
- [ ] required by #5749 — manifest resolves CDN URLs and expected hashes for download
- [ ] required by #5750 — manifest provides the authoritative list for pruning

## References
- [Epic: Solo Image Cache Enhancements](https://github.com/hiero-ledger/solo/issues/5738)

Issue #5745, opened 2026-08-06
[Image Cache] A3: Generate image tars and SHA-256 hashes in CI


jeromy-cannon:
## Overview
Add a CI script that dumps each container image present in the deployed cluster into a named tar file (registry/container/version embedded in filename per the A1 schema), computes SHA-256, and writes a `.sha256` file containing the hash value. Outputs feed the CDN upload step (A4) and manifest generation step (A5).

## Sub-Tasks
- [ ] Script to enumerate all images from the deployed cluster (output of A2)
- [ ] Export each image as a named tar using the filename convention from A1
- [ ] Compute SHA-256 for each tar
- [ ] Write `<tarname>.sha256` containing the hash value only

### Dependencies
- [ ] requires #5742 — manifest schema defines tar and hash filename conventions
- [ ] requires #5743 — running cluster provides the images to extract
- [ ] required by #5746 — tars and hashes are inputs to the CDN upload step
- [ ] required by #5747 — image list and hash values feed manifest generation

## References
- [Epic: Solo Image Cache Enhancements](https://github.com/hiero-ledger/solo/issues/5738)

PR #5772, opened 2026-08-07
feat: prune image cache files that the current manifest does not list

### LINKED ISSUE #5750: [Image Cache] B4: Auto-prune stale cache files not in current manifest

ddeskov-limechain:
## Description

Stacked on #5768; part of the B1→B6 chain (#5766 → #5767 → #5768 → this), merge in order. The diff against `main` will show the B1, B2 and B3 commits too until they merge — review the `feat/5750-cache-prune` commit only.

`solo cache image pull` now deletes every archive and hash file in the local image cache directory that the manifest for the running Solo version does not list. Without this, upgrading Solo leaves the archives of every version the user ever ran on disk — each one hundreds of megabytes — with nothing that ever removes them.

**Prune only runs when a manifest actually resolved.** The manifest is the only authoritative statement of what belongs in the cache; when it cannot be fetched or parsed, `pull()` returns its "no image cache manifest available" task and no prune task at all, so a network blip or an unpublished release can never delete archives that are still valid. This is the fail-safe direction: too little pruning wastes disk, too much pruning throws away good archives and forces a full re-download.

**Only image cache entries in the image cache directory are considered.** The directory is the one the catalog store resolves for image targets; the candidates are its own directory entries (not a recursive walk), filtered to regular files ending in `.tar` or `.sha256`. Because a candidate is a bare directory entry name, the join can only ever address a file inside that directory — no manifest value or user input reaches the path, and nothing outside the image cache can be named. A `cache-catalog.json`, a subdirectory, or anything else in there is left untouched.

**Each removal is reported.** Pruned files go into a new `Image cache maintenance` message group, printed by the existing end-of-run "Show user messages" step alongside the existing `Image cache failures` group, so a user sees exactly which files solo deleted on their behalf rather than a silent disk change. The task title carries the count.

### What changed

  * `pull()` now emits a `Prune stale image cache files` task ahead of the per-image tasks, on the manifest-resolved path only. Pruning before downloading also means a stale archive's disk space is released before the new one is fetched.
  * `pruneStaleFiles()` — reads the image cache directory, deletes every `.tar`/`.sha256` file not in the expected set, records each deletion, and returns what it removed. A missing directory (nothing cached yet) is a no-op, not an error.
  * `resolveExpectedFileNames()` — builds the keep set from **both** the manifest's `tarFile`/`hashFile` names and the local file names the catalog store resolves for the manifest-listed targets. The two follow the same convention (`<registry>__<container>__<version>.tar`), but the manifest names are produced by the release workflow (#5745/#5747) while the local names come from `FileSystemCacheCatalogStore.resolvePath()`. Keeping both spellings means a drift between the two repos' conventions can waste disk, but can never delete the archive this very pull just downloaded. That asymmetry is deliberate.
  * `fetchManifestImages()` — the manifest fetch and index that `pull()` already did, extracted so #5751's load-time re-validation reuses it instead of copying it.
  * `recordMaintenance()` / `recordMessage()` — the existing `recordFailure()` body generalised over the message group key, so maintenance output does not get filed under "failures".
* `src/core/constants.ts` — `CACHE_IMAGE_MAINTENANCE_MESSAGE_GROUP`, next to the existing failure group constant.
* `test/unit/integration/cache/image-cache-pull.test.ts` — 3 tests added to the existing pull suite, which already builds a real temp cache directory and a fake CDN.

Design note on scope: the existing `solo cache image prune` command wipes the whole cache directory and is unchanged — it is the user's "delete everything" button. This PR is the automatic, manifest-driven prune the epic asks for, and the two do not overlap.

### Related Issues

* Fixes #5750
* Related to #5738, #5748, #5749

On documentation: no user-facing docs describe the image cache internals, and no flag or environment variable was added.

### Testing

The following was run:

* `npx tsc --noEmit` — exit 0
* `npx mocha 'test/unit/**/*.ts' --exclude 'test/unit/core/helm/**/*.ts'` — **1720 passing, 0 failing** (1717 on the B3 branch, 3 added)

Unit coverage: a stale archive and its hash file are removed while the manifest-listed archive is kept and the deletion is reported; nothing is pruned when the manifest is unavailable; a non-cache file and a directory whose name ends in `.tar` are both left alone.

The following was not tested:

* No live manifest. `cdn.solo.hashgraph.io` is still being provisioned (RESD-1110) and no release publishes a `cache-manifest.json` yet, so prune has never run against real manifest file names — which is exactly why the keep set holds both name spellings.
* No E2E run against a real `~/.solo/cache/images` directory populated by a previous Solo version; removing archives an older Solo version left behind is #5752's job and is covered there.
* Disk space reclaimed was not measured.


jan-milenkov (dismissed):

src/integration/cache/impl/image-cache-handler.ts
```diff
@@ -128,6 +131,118 @@ export class ImageCacheHandler implements CacheOperationHandler {
...
+
+    for (const entry of entries) {
+      if (!entry.isFile() || keep.has(entry.name) || !ImageCacheHandler.isCacheFileName(entry.name)) {
+        continue;
+      }
+
+      // entry.name is a single directory entry, so the join can only ever address a file inside the
+      // image cache directory itself.
+      const filePath: string = PathEx.join(directory, entry.name);
+      await fs.rm(filePath, {force: true});
```
jan-milenkov, line 201: Found with AI, please validate behavior (might be a false flag)

`fs.rm(..., {force: true})` ignores a missing file. It does not ignore EPERM, EBUSY or EACCES.

A throw here fails the "Prune stale image cache files" task, and nothing turns `exitOnError` off on this path — it is explicitly `true` for the image cache group in one-shot deploy (`default-one-shot-deploy-orchestrator.ts:469`). So one unremovable file in `~/.solo/cache/images` stops the deploy. See #5302 for that failure on Windows.

The JSDoc on `pull()` says "Nothing here aborts the run", which prune breaks.

Wrap the removal, record it as a maintenance message, and keep going.

ddeskov-limechain: Relevant, confirmed: `fs.rm({force: true})` only swallows ENOENT, and `pull()` runs inside the one-shot 'Cache container images' group with exit-on-error, so an EPERM here would stop the deploy.

Fixed in this stack at #5774 (90e958c7), which moves prune into `ImageCacheHousekeeper`: `remove()` wraps every `fs.rm`, reports `Could not remove image cache file, left in place: … (EPERM …)` as a maintenance message and continues. Covered by a test that stubs `rm` to throw EPERM for a stale file and checks the manifest-listed image is still cached. Kept the fix in #5774 rather than backporting it here, since the stack merges in order and #5774 replaces this method.

jan-milenkov (dismissed):

Issue #5738, opened 2026-08-06
(needs CDN) Solo Image Cache Enhancements


jeromy-cannon:
## Goals
 
- add a job into the deploy and release artifact workflow that will run on a clean cluster `one-shot single deploy`
  - generate a manifest for the version of Solo being tagged containing
    - this version of Solo that is being tagged
    - the image tar file names with the registry/container/version embedded in the file name for images dumped from a cluster post `one-shot single deploy` for this current version
    - the image validation hash files using the same file name as the image but with different file extension
    - the validation hash (yes, we want both the hash value, as well as the filename that will also have the hash, then both can be checked to ensure they match for added security)
  - generate a tar and validation hash, and a file containing the validation hash for each image
  - upload the tar, and verification hashes to our CDN, see notes below for ongoing request to build this
    - File doesn't exist on CDN → upload it
    - File exists AND hash matches → skip (idempotent, safe)
    - File exists AND hash differs → fail the CI job
  - upload the manifest file as a release artifact (this gives us the ability to manually make changes if required and repost the release artifact)
- enhance existing `solo cache image *` commands to 
  - leverage this new manifest by pulling it based on the current version of Solo from (example: https://github.com/hiero-ledger/solo/releases/download/v0.86.0/cache-manifest.json)
  - construct links to images and validation files dynamically with the base `cdn.solo.hashgraph.io` (though subject to change)
  - validate the tar image files against the validation hash
  - validate that the manifest copy of the hash and the value in the validation hash file match
  - revalidate the hash match any time we are loading an image into the cluster to ensure they haven't been altered
  - automatically prune/delete any tar, validation hash files, that are not in the manifest for the currently running version of Solo
  - only pull and cache images on a `solo cache image pull` command
  - never add anything to the image cache on the users file system except when running the `solo cache image pull`
  - remove any solo cache image pull source code that pulls from container image registries. we either pull from the CDN to cache, then load the cache into the cluster, or we let k8s handle the normal pulling for anything missing.  the kind config mirror registry entries can be left in place.
  - migration from prior caching to new caching model must be put in place to properly prune/etc.
  - a hash mismatch should trigger a full redownload of that tar on the next `solo cache image pull` while those with a hash match can skip the download
- once we have the new Solo SEA in place we might be able to embed the manifest directly into the installer

## Dependency Map

```
A1 ─────────────────────┬──────────────────────────────────────────┐
                        │                                          │
[...]

PR #5773, opened 2026-08-07
feat: re-validate cached image archives against the manifest before loading them

### LINKED ISSUE #5751: [Image Cache] B5: Re-validate image tar hash before loading into cluster

ddeskov-limechain:
## Description

Stacked on #5772; part of the B1→B6 chain (#5766 → #5767 → #5768 → #5772 → this), merge in order. The diff against `main` will show the B1–B4 commits too until they merge — review the `feat/5751-load-revalidation` commit only.

A cached archive is verified when it is downloaded (#5768), but that was the last time anyone looked at it. Between the pull and the load it can be corrupted by a bad disk, truncated by a full one, or rewritten by anything with access to `~/.solo/cache/images`. `solo cache image load` now rehashes each archive and compares it against the manifest hash immediately before handing it to the container engine, so bytes solo cannot vouch for never reach the cluster.

**On mismatch the load aborts.** The archive and its hash file are deleted first, then a new coded error (`SOLO-5084`, `CacheArchiveHashMismatchSoloError`) is raised naming the image, the archive path, and both hashes, with troubleshooting steps that tell the user to run `solo cache image pull`. Deleting before throwing is what makes the advice true: the next pull sees no archive, downloads it fresh, and does not fail the same way twice. Unlike the load failures already handled here, this one is not swallowed into the end-of-run summary — a corrupted or tampered archive is exactly the case the epic wants to stop, and continuing would either load nothing for that image silently or leave the deployment half-built.

**With no manifest, the load proceeds.** This is the design decision worth arguing about, so, explicitly: when the manifest cannot be fetched, or when it does not list the image, there is no trusted hash to compare against, and the archive is loaded as it is (a debug line records why). The alternative — treat "nothing to compare against" as a failure — would make `solo cache image load` impossible today, because no Solo release publishes a `cache-manifest.json` yet (#5745/#5747) and the CDN is still being provisioned (RESD-1110), and it would still fail later for any image legitimately absent from the manifest, such as a component version supplied by a flag that the release manifest never saw. The archive also cannot have arrived by accident: after #5766 the only code path that writes into the image cache is `solo cache image pull`, which verifies everything it writes. Re-validation strictly adds a check on top of that; it does not become the only one. When manifests are published this branch stops being reachable in practice for the images the manifest covers.

The hashing helper, the discard helper and the manifest fetch are the ones #5768 and #5772 already added — nothing was duplicated for this path.

### What changed

  * `load()` resolves the manifest once per invocation and passes each image's entry to the per-image task, so the manifest is fetched once rather than once per image.
  * `resolveManifestImagesForLoad()` — wraps the shared `fetchManifestImages()` and degrades to an empty map on any failure, with the reason logged at debug level.
  * `assertArchiveMatchesManifest()` — rehashes the archive with the existing `computeSha256()`, returns on a match or when there is no manifest entry, and otherwise discards the archive with the existing `discardArchive()` and throws. Called after the "already loaded in the cluster" and "not cached" checks, so it only ever reads an archive that is actually about to be loaded.
* `src/core/errors/classes/system/cache-archive-hash-mismatch-solo-error.ts` — new. Non-retryable (retrying the load rehashes the same bytes), owned by `Infrastructure` (the local filesystem, not Solo and not user input).
* `src/core/errors/error-code-registry.ts` — `CACHE_ARCHIVE_HASH_MISMATCH: 'SOLO-5084'`.
* `src/core/errors/solo-errors.ts` — registered as `SoloErrors.system.cacheArchiveHashMismatch`.
* `test/unit/integration/cache/image-cache-handler.test.ts` — 3 tests added; the suite now builds a real temp cache directory so archives can be hashed for real, and stubs `CacheManifestClient.fetchImages` in `beforeEach` so no test reaches the network.

### Related Issues

* Fixes #5751
* Related to #5738, #5748, #5749, #5750

On documentation: the new error code carries its own troubleshooting steps and documentation URL through the standard `SoloError` machinery; no separate docs page describes image cache internals.

### Testing

The following was run:

* `npx tsc --noEmit` — exit 0
* `npx mocha 'test/unit/**/*.ts' --exclude 'test/unit/core/helm/**/*.ts'` — **1723 passing, 0 failing** (1720 on the B4 branch, 3 added)

Unit coverage: an archive whose hash matches the manifest is loaded; a corrupted archive aborts the load with `SOLO-5084`, a message naming both the archive path and the image, no call into the container engine, and the archive plus its hash file deleted; an image the manifest does not list is loaded anyway. The four pre-existing load cases now run against the no-manifest path, which is the one every user hits today.

The following was not tested:

* No live manifest and no live CDN — nothing has been fetched from `cdn.solo.hashgraph.io` (RESD-1110) and no release publishes a `cache-manifest.json`, so the mismatch path has never been exercised against a real published hash.
* No E2E run: `solo cache image load` against a real Kind cluster, and `solo one-shot single deploy`, were not run.
* The cost of rehashing every archive on every load was not measured. It is one sequential read per archive, the same read `pull` already does on a cache hit, and it happens while the container engine imports the previous archive.
* Behaviour under a concurrent `solo cache image pull` writing an archive while a load hashes it was not tested; that race exists on `main` too.


jan-milenkov (commented):
LGTML @jeromy-cannon please review

jan-milenkov (dismissed):

jan-milenkov (dismissed):

Issue #5744, opened 2026-08-06
[Image Cache] B1: Remove registry-pull code paths from solo cache image commands


jeromy-cannon:
NOTE: while this work can be done in advance, merging this prior to having the completed feature in place will break the existing `solo cache image pull/load`.

## Overview
Delete all source code that pulls container images directly from registries inside the `solo cache image` commands. The new model is CDN-pull-to-local-cache then load-from-cache-into-cluster, or let Kubernetes handle normal pulling for anything missing. Registry-pull paths are removed entirely. Kind config mirror registry entries may remain.

## Sub-Tasks
- [ ] Identify all code paths that pull from container registries in `solo cache image` commands
- [ ] Remove identified registry-pull logic
- [ ] Verify Kind config mirror registry entries are left in place (acceptable to keep)
- [ ] Update or remove affected unit/integration tests
- [ ] Ensure `solo cache image pull` is the only path that adds files to the local image cache

### Dependencies
- [ ] required by #5752 — registry-pull removal is a prerequisite for a clean migration

## References
- [Epic: Solo Image Cache Enhancements](https://github.com/hiero-ledger/solo/issues/5738)

Issue #5302, opened 2026-07-24
bug: one-shot deploy self-poisons ~/.solo on Windows — restrictToOwnerWindows empties a cache file DACL (SOLO-3035 EPERM), unrecoverable via destroy / rm -rf


Reccetech:
**Date:** 2026-07-23
**Source:** Code analysis of `C:\repos\solo` (`main`, commit `134aba372`) + live reproduction
**Tested:** PowerShell / Windows 11 (10.0.26200) / Solo `main` via `npm run solo-test` and released `@hashgraph/solo` 0.83.0; Docker Desktop 28.3.2 + kind 0.28.0
**Status:** Root cause confirmed and reproduced; one-line fix applied and verified to produce a working deploy.

## Related issues

- **#4948** (Harden install/uninstall/destroy lifecycle — Epic): this issue is a concrete, severe instance of its thesis that "the only reliable full reset is still `rm -rf ~/.solo`" — here **even `rm -rf ~/.solo` fails** until the bricked file's ACL is repaired.
- **#4937** (SOLO-3035, CLOSED): **same error code, different root cause.** #4937 was Mac + Homebrew formula merge-conflict + "remote configmap not found". This is Windows + an empty-DACL `EPERM` from Solo's own `icacls` hardening. **Not a duplicate** — a distinct 3035 trigger.
- **#4947** (cache orphaned archives, OPEN) and **#4580** (image-cache implementation issues, CLOSED): Defect 1 below (0-byte/corrupt archives loaded without size validation) is **adjacent** to these cache-hygiene issues and extends them with a size/validity check in `load()`/`healthcheck()`; the **primary** defect here (Defect 2, the ACL bug) is unrelated to and uncovered by either.
- **#4512 / #4966** (Windows port-forward lifecycle): tangential — recreating mirror/relay port-forwards was needed during recovery.

## Summary

A `one-shot falcon deploy` fails and **cannot self-recover** through the normal `destroy` → `deploy` cycle. The failures originate from corrupted / access-denied artifacts in the persistent local cache at `~/.solo/`, which `destroy` does not touch and `deploy` reuses without validation. The remedy suggested in the `SOLO-3035` error text (`rm -rf ~/.solo`) is heavy — it discards several GB of otherwise-valid cached images — and on Windows it **does not even succeed** until the poisoned file's ACL is repaired first (the file's empty DACL denies its own deletion; see below).

[...]

PR #5774, opened 2026-08-07
feat: migrate the image cache off the registry-pull layout on the first pull

### LINKED ISSUE #5752: [Image Cache] B6: Migration from old image cache model to new CDN-based model

ddeskov-limechain:
## Description

Stacked on #5773; part of the B1→B6 chain (#5766 → #5767 → #5768 → #5772 → #5773 → this), merge in order and last of the chain. The diff against `main` will show the B1–B5 commits too until they merge — review the `feat/5752-cache-migration` commit only.

Anyone upgrading to the CDN cache model already has a `~/.solo/cache/images` full of archives the old model produced. This removes them on the first `solo cache image pull` after the upgrade, says so in plain words, and downloads the published archives in their place in the same run.

**How an old-format entry is recognised.** An archive with no `.sha256` file next to it. Before #5766, archives were produced by exporting an image the local container engine had pulled from a registry; nothing published a hash for those bytes, so they were written on their own. Every archive the new model writes gets its hash file first and keeps it (#5768), so the presence of the sibling is exactly the line between the two models. Detecting the *layout* rather than a file name is what makes this hold up: image names, tags, and the naming convention itself can all change without breaking the rule, a current entry can never be mistaken for a legacy one, and a second run finds nothing to do — the idempotency the issue asks for falls out of the rule instead of needing a marker file.

**Why delete instead of keep.** Those archives cannot be verified now and cannot be verified later: there is no hash for them anywhere, and #5773 refuses to load an archive it cannot vouch for. Keeping them would mean either loading unverified bytes into the cluster or holding gigabytes that are never loaded and never cleaned up. Deleting is also cheap to recover from — the published archive is downloaded in the same run.

**Where it hooks in.** The top of `ImageCacheHandler.pull()`, before the manifest is fetched, so migration also runs when no manifest is available (the state of every Solo release until #5745/#5747 land). Migration must clean up regardless: a user upgrading today would otherwise keep unverifiable archives indefinitely, and `solo cache image load` would keep loading them. `pull` is the natural home — after #5766 it is the only code path allowed to write into the image cache, so it is the only place a repaired cache can be repopulated.

**Ordering, and why this PR also moves #5772's prune.** The issue requires migration to be followed by B3's download and B4's prune, but `solo cache image pull` runs the subtasks its handler returns *concurrently* (`CACHE_IMAGE_MAX_CONCURRENCY`). As tasks in that list, migration and prune would race the downloads: a download task that ran first would find the legacy archive on disk, rehash it, and report it as a corrupted cache entry — leaving the image uncached for the whole run instead of quietly replacing it. So migration and prune now run inline in `pull()` before the task list is built, which makes migrate → prune → download unconditional, and the two housekeeping tasks are replaced by a single `Tidy the image cache` line that reports what already happened. Same work, same output, no ordering to get wrong.

**What is deliberately not migrated:**

* `cache-catalog.json` — `FileSystemCacheCatalogStore.save()` has never had a production caller on `main` (only a unit test), so no user has this file. Nothing to migrate.
* `<cacheDir>/config/solo-cache-images-target.rendered.yaml` — still written and read by the current model; it is not an old-format artifact.
* Chart cache entries under `~/.solo/cache/charts` — a different handler and outside this epic's scope.
* Hash files with no archive next to them — a legitimate intermediate state of an interrupted download, not an old-model entry. Prune (#5772) removes them when the manifest does not list them.

### What changed

  * `migrateLegacyEntries()` — collects the image cache directory entries once, deletes every `.tar` whose `.tar.sha256` sibling is absent, records each removal, and returns what it removed. Collecting the names first means directory ordering cannot affect the outcome.
  * `pull()` — awaits migration and then prune before building the task list, for the reason above.
  * `buildHousekeepingTask()` — the single reporting line that replaces `buildPruneTask()` and the migration task.
  * `resolveImageCacheDirectory()` / `readImageCacheDirectory()` — the directory resolution and the missing-directory tolerance #5772 added, extracted so migration and prune share one definition of where the cache lives and one behaviour when nothing has been cached yet.
* `test/unit/integration/cache/image-cache-pull.test.ts` — 3 tests added. Four pre-existing tests now write the hash file alongside the archive they pre-seed, because that pair is the new model's invariant; without it those fixtures are legacy entries by definition and migration correctly removes them.

### Related Issues

* Fixes #5752
* Related to #5738, #5744, #5749, #5750

On breaking changes: no interface, flag or command changes. Archives cached by an older Solo version are deleted on the first pull after upgrading — that is the point of the issue, it is reported file by file in the end-of-run image cache maintenance summary, and the replacement is downloaded in the same run.

On documentation: no user-facing docs describe the image cache internals.

### Testing

The following was run:

* `npx tsc --noEmit` — exit 0
* `npx mocha 'test/unit/**/*.ts' --exclude 'test/unit/core/helm/**/*.ts'` — **1726 passing, 0 failing** (1723 on the B5 branch, 3 added)

Unit coverage: an archive with no hash file is removed, reported, and replaced by the published archive in the same run, with the cached-item result recorded; the same removal happens when the manifest is unavailable; a second pull over a migrated cache removes nothing, reports nothing, and downloads nothing. The pre-existing suite covers the other half — archives that do have their hash file are left alone by every one of these paths.

The following was not tested:

* No live manifest and no live CDN — `cdn.solo.hashgraph.io` is still being provisioned (RESD-1110) and no release publishes a `cache-manifest.json`, so migration has never been followed by a real download.
* No E2E run: upgrading a real installation with a populated `~/.solo/cache/images` from a pre-#5766 Solo and then running `solo cache image pull` / `solo one-shot single deploy` was not exercised.
* How much disk a real upgrade reclaims, and how long the replacement download takes, were not measured.


jan-milenkov (commented):
LGTM

src/integration/cache/impl/image-cache-handler.ts
```diff
@@ -239,6 +294,27 @@ export class ImageCacheHandler implements CacheOperationHandler {
...
+
+  /** Directory entries of the image cache, empty when nothing has been cached yet. */
+  private async readImageCacheDirectory(targets: readonly CacheTarget[]): Promise<readonly Dirent[]> {
+    if (targets.length === 0) {
+      return [];
+    }
+
+    try {
+      return await fs.readdir(this.resolveImageCacheDirectory(targets), {withFileTypes: true});
+    } catch {
```
jan-milenkov: This catch has no comment. CLAUDE.md requires one on any catch that swallows the error.

It also hides more than a missing directory. An EACCES here returns an empty list, which silently turns off both migration and prune.

#5772 had the explanation on this same catch (`// nothing has been cached yet, so there is nothing to prune`) and it was dropped when the method was extracted.

```suggestion
    } catch {
      // nothing has been cached yet, so there is no directory to read
```

src/integration/cache/impl/image-cache-handler.ts
```diff
@@ -133,7 +142,82 @@ export class ImageCacheHandler implements CacheOperationHandler {
...
+   * epic requires that only verified archives are loaded into a cluster. It is deleted, and
+   * {@link downloadArchive} fetches the published archive for that image in the same run.
+   *
+   * The rule is the layout rather than a file name, which is what makes this safe to run on every pull:
+   * every archive the new model writes gets its hash file first, so a current entry is never mistaken for a
+   * legacy one, and a second run finds nothing left to do.
+   *
+   * @returns the names of the files that were removed
+   */
+  private async migrateLegacyEntries(targets: readonly CacheTarget[]): Promise<readonly string[]> {
```
jan-milenkov: The handler is 612 lines and about 20 methods after this stack.

`migrateLegacyEntries`, `pruneStaleFiles`, `resolveImageCacheDirectory` and `readImageCacheDirectory` are one job — repairing the cache directory — and none of them touch the download or load paths.

Worth moving into their own class. Not blocking.

test/unit/integration/cache/image-cache-pull.test.ts
```diff
@@ -260,13 +263,61 @@ describe('ImageCacheHandler pull', (): void => {
...
+  });
+
+  it('finds nothing to migrate on a second run', async (): Promise<void> => {
+    await fs.writeFile(archivePath, ARCHIVE_CONTENTS);
+    await fs.writeFile(hashPath, ARCHIVE_HASH);
+    sinon.stub(CacheManifestClient, 'fetchImages').resolves([manifestImage()]);
+    const fetchFile: SinonStub = stubDownloader({});
+    const handler: ImageCacheHandler = createHandler(fetchFile);
+
+    await runPull(handler);
```
jan-milenkov, line 376: `runPull` awaits the subtasks one at a time. Production runs 12 at once (`CACHE_IMAGE_MAX_CONCURRENCY`, `constants.ts:468`).

Moving migration and prune inline was specifically to stop them racing the downloads, and no test covers that case. A test that starts the subtasks concurrently would.

jan-milenkov (dismissed):

src/integration/cache/impl/image-cache-handler.ts
ddeskov-limechain: Done in 89d66cae. The catch now swallows ENOENT only, with the comment; anything else (EACCES included) is raised so housekeeping is never silently skipped. Covered by a test that stubs readdir with EACCES and expects pull() to reject.

src/integration/cache/impl/image-cache-handler.ts
ddeskov-limechain: Done in 89d66cae: `migrateLegacyEntries`, `pruneStaleFiles` and the directory helpers now live in `ImageCacheHousekeeper` (store only, no logger). It returns the removed paths and the handler records the maintenance messages, so the handler keeps the download/load paths and the reporting. Handler is down to 481 lines.

test/unit/integration/cache/image-cache-pull.test.ts
ddeskov-limechain: Done in 89d66cae. `runPull` takes a `concurrent` flag that starts every subtask at once via Promise.all, and a new test uses it with a legacy archive on disk. Verified it discriminates: moving migration back into the task list makes that test fail, the inline ordering passes.

src/integration/cache/impl/image-cache-handler.ts
```diff
@@ -75,15 +73,29 @@ export class ImageCacheHandler implements CacheOperationHandler {
...
+   * cannot be read, which is raised rather than silently skipping the housekeeping.
    */
   public async pull(): Promise<SoloListrTask<AnyListrContext>[]> {
     const targets: readonly CacheTarget[] = await this.resolveRequiredArtifacts();
 
+    // Housekeeping runs here rather than as a task in the list returned below, because the command runs
+    // those tasks concurrently: a download that started before the cache was migrated would rehash the
+    // legacy archive, report it as a corrupted cache entry and leave the image uncached for the rest of the
+    // run, instead of quietly replacing it. Doing the work up front makes the order unconditional.
+    const migrated: readonly string[] = await this.housekeeper.migrateLegacyEntries(targets);
```
jan-milenkov: This runs before the manifest is fetched. If that fetch fails, lines 99-118 return early and nothing is downloaded. No release publishes a manifest yet, so that is the path every pull takes today.

Migration deletes any `.tar` with no `.sha256` next to it. Nothing on `main` ever wrote a `.sha256`, so that is every archive a user has. Those archives also still work right now — with no manifest entry, `assertArchiveMatchesManifest` skips the check and loads them.

So the behaviour today is: the whole cache is deleted, nothing replaces it, and every image goes back to being pulled from its registry. The test `removes an archive cached by an older solo version even without a manifest` locks that in.

`pruneStaleFiles` already guards against this exact thing — "deleting on a guess would throw away archives that are still valid". Migration needs the same guard: run it after the manifest resolves, not before.

src/integration/cache/impl/image-cache-housekeeper.ts
```diff
@@ -0,0 +1,168 @@
...
+        !fileName.endsWith(ImageCacheHousekeeper.ARCHIVE_FILE_EXTENSION) ||
+        fileNames.has(`${fileName}${ImageCacheHousekeeper.HASH_FILE_EXTENSION}`)
+      ) {
+        continue;
+      }
+
+      // fileName is a single directory entry, so the join can only ever address a file inside the image
+      // cache directory itself.
+      const filePath: string = PathEx.join(this.resolveImageCacheDirectory(targets), fileName);
+      await fs.rm(filePath, {force: true});
```
jan-milenkov: `fs.rm(..., {force: true})` ignores a missing file. It does not ignore EPERM, EBUSY or EACCES. Same at line 95 in `pruneStaleFiles`.

Both run before the task list is built, so a throw here escapes `pull()` and fails the parent "Pull and cache container images" task. That group runs with `exitOnError: true` in one-shot deploy (`default-one-shot-deploy-orchestrator.ts:469`), so the deploy stops.

One locked or bad-DACL file in `~/.solo/cache/images` and the deploy dies while tidying up. See #5302 for that failure on Windows.

89d66cae documented the readdir exception in `pull()`'s JSDoc ("a cache directory that exists but cannot be read"), but these two removals can throw too and are not covered by it.

Wrap each removal, report the failure as a maintenance message, and keep going — a file solo cannot delete should not cost the user their deploy.

jan-milenkov (commented):
Found with AI, please check if any of those are relevant

src/integration/cache/impl/image-cache-handler.ts
ddeskov-limechain: Relevant, confirmed: main never wrote a .sha256, no release ships a manifest yet, and the load path accepts an archive with no manifest entry. Fixed in 90e958c7: migration now runs after the manifest resolves, right before prune, and the no-manifest path leaves the cache untouched. The test that locked in the deletion is replaced by one asserting the legacy archive survives when the manifest is unavailable.

src/integration/cache/impl/image-cache-housekeeper.ts
ddeskov-limechain: Relevant, confirmed: the pull runs inside the one-shot 'Cache container images' group with exit-on-error. Fixed in 90e958c7: every removal goes through one wrapper that reports the outcome either way; a file that cannot be deleted is left in place with a 'Could not remove … (EPERM …)' maintenance message and the pull continues. Covered by a test that stubs rm to throw EPERM for a stale file and checks the manifest-listed image is still cached.

jan-milenkov (dismissed):

PR #5796, opened 2026-08-10
fix: solo `--version` throws a bunch of `SOLO-9007` warnings, fixed by trying to use the logger while having and unwritable logs directory

### LINKED ISSUE #5370: solo --version throws a bunch of SOLO-9007 warnings
I am trying to run `solo` on my mac following official solo docs here: https://solo.hiero.org/docs/simple-solo-setup/quickstart/

I installed `solo` using `npm`:

```shell
sudo npm install -g @hiero-ledger/solo
```

Then I was about to verify it had been installed correctly by checking the version:

```shell
$ solo --version
```

It prints the version, but then emits a number of errors like this:

```shell
Version                 : 0.83.0
╭─ ERROR ────────────────────────────────────────────────────────────────────────────────────────────╮
│ [SOLO-9007] uncaughtException uncaughtException failed — expected returned value to be true        │
│   → This is an internal Solo error. File a bug report: https://github.com/hiero-ledger/solo/issues │
│                                                                                                    │
│ Learn more: https://solo.hiero.org/docs/troubleshooting/errors/internal/SOLO-9007/                 │
╰────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

The errors are kept printing forever, the process doesn't exit cleanly, I have to kill it explicitly.

`solo -v` has the same output. `solo version` (which is not expected to work) prints usage info - and then the same endless list of SOLO-9007 errors.

Any ideas what could be wrong with my environment and how to debug and fix it? Thanks!

jan-milenkov:
## Description

When `$SOLO_HOME/logs` isn't writable — typically root-owned files left behind by an old
`sudo npm install -g` — `solo --version` printed SOLO-9007 boxes forever and never exited
(526,198 boxes / 271 MB in 25s on 0.83.0). The `uncaughtException` handler logged through the
broken logger, which raised another error straight back into the handler.

0.84.x stopped the loop only incidentally, when the logger moved to `rotating-file-stream`. The
handler itself is unchanged, so on latest a permissions problem is still reported as "internal Solo
error, file a bug report", exits 0, and never shows the EACCES or the offending path.

Solo now checks the log destination before building the logger and fails with **SOLO-5081**, naming
the exact path along with chown / `rm -rf ~/.solo` / `SOLO_HOME` remediation. Reporting of escaped
errors is latched, so a failing logger can no longer drive an unbounded report loop, the real error
is kept as `cause` rather than discarded, and the process exits non-zero. **SOLO-9014** replaces the
misused SOLO-9007 for errors that escape every handler.

### Related Issues

* Closes # https://github.com/hiero-ledger/solo/issues/5370


src/core/errors/classes/system/solo-logs-directory-not-writable-solo-error.ts
```diff
@@ -0,0 +1,35 @@
...
+ * an earlier `sudo npm install -g @hiero-ledger/solo`, which leaves the files owned by root.
+ */
+export class SoloLogsDirectoryNotWritableSoloError extends SoloError {
+  protected override readonly retryable: boolean = false;
+  protected override readonly ownership: ErrorOwnership = ErrorOwnership.Infrastructure;
+
+  public constructor(logPath: string, cause?: Error) {
+    super(
+      {
+        message: `Solo cannot write to its log destination: ${logPath}`,
```
Fantomasa: This error is rendered by `FatalErrorReporter.reportWithoutLogger`, which prints the code, `error.message`, the troubleshooting steps and the doc URL — but not `.cause`. So the underlying errno is preserved on the object yet never shown in the one path this error is actually used in, and the message carries no errno either. That means EACCES (chown), EROFS (`SOLO_HOME`) and ENOSPC (disk full) all render identically, even though two of the listed steps are wrong for the latter two.

Every sibling system error appends `: ${cause.message}`, and `UncaughtFatalErrorSoloError` in this same PR already handles the optional cause with a `detail` string — mirroring that here surfaces the errno:

```ts
const detail: string = cause?.message ? `: ${cause.message}` : '';
// …
message: `Solo cannot write to its log destination: ${logPath}${detail}`,
```

(The fs message repeats the path, so it'd show twice — matches the sibling shape; dedupe if you'd rather.)

src/core/errors/classes/system/solo-logs-directory-not-writable-solo-error.ts
```diff
@@ -0,0 +1,35 @@
...
+        code: ErrorCodeRegistry.SOLO_LOGS_DIRECTORY_NOT_WRITABLE,
+        troubleshootingSteps:
```
Fantomasa, line 25: These steps are Unix-only (`sudo chown`, `id -u`, `rm -rf`). The triggering scenario — root-owned files from `sudo npm install -g` — is itself Unix-specific, and on Windows `accessSync(W_OK)` won't meaningfully fail (you skip the denial cases there in the test for the same reason), so a Windows user is unlikely to ever hit this. Just confirming the Unix-only remediation is the intended scoping?

Fantomasa (changes requested):

Ivo-Yankov (dismissed):

Fantomasa (dismissed):

src/core/errors/classes/system/solo-logs-directory-not-writable-solo-error.ts
jan-milenkov: Yes, Unix-only is intended, and it matches what the rest of the error classes already do:

- `container-engine-not-running-error.ts:25` — `Start Docker Desktop, or the Docker daemon: sudo systemctl start docker`
- `one-shot-deploy-failed-solo-error.ts:27` — `rm -rf ~/.solo`
- 149 classes tell the user to `tail -n 100 ~/.solo/logs/solo.log`

So a Windows user already meets Unix-shaped remediation throughout solo, and diverging only here would be the inconsistent choice. If we want platform-aware steps, it's worth doing across the board in its own change rather than one-off here.

src/core/errors/classes/system/solo-logs-directory-not-writable-solo-error.ts
jan-milenkov: updated

Fantomasa (dismissed):
LGTM, Thank you!

test/unit/core/logging/solo-pino-logger.test.ts
```diff
@@ -352,3 +355,105 @@ describe('SoloPinoLogger stream configuration', (): void => {
...
+  const fileNames: string[] = ['solo.ndjson', 'solo.log'];
+  let logsDirectory: string;
+
+  beforeEach((): void => {
+    logsDirectory = PathEx.join(mkdtempSync(PathEx.join(tmpdir(), 'solo-logs-')), 'logs');
+  });
+
+  afterEach((): void => {
+    // Restore write permission so the temporary tree can be cleaned up.
+    chmodSync(logsDirectory, 0o700);
```
jeromy-cannon: `chmodSync` throws `ENOENT` when `logsDirectory` doesn't exist, which masks the real test failure if the directory was never created. `rmSync` two lines below already uses `force: true` — the `chmodSync` should be guarded the same way:

```suggestion
    if (existsSync(logsDirectory)) {
      chmodSync(logsDirectory, 0o700);
    }
```

src/core/errors/classes/system/solo-logs-directory-not-writable-solo-error.ts
jeromy-cannon: our goal is to provide the Windows users with a skimmed down msys2 system for Solo to leverage, we may want to consider how the users can leverage this msys2 system to run these commands

test/unit/core/logging/solo-pino-logger.test.ts
jan-milenkov: updated

src/core/logging/solo-pino-logger.ts
```diff
@@ -59,18 +59,13 @@ export class SoloPinoLogger implements SoloLogger {
...
-    } catch {
-      // no-op: if this fails, pino will attempt to create the files and error if impossible
-    }
-
     // Configure dual outputs: NDJSON (machine) + pretty (human)
     const ndjsonFileName: string = 'solo.ndjson';
     const prettyFileName: string = 'solo.log';
 
+    const logsDirectory: string = constants.SOLO_LOGS_DIR;
+    SoloPinoLogger.assertLogDestinationsWritable(logsDirectory, [ndjsonFileName, prettyFileName]);
```
JeffreyDallas: two things here, and they're the ones I'd most like resolved before this merges.

First: this makes an unwritable logs directory fatal to *every* invocation. The old code deliberately swallowed the `mkdirSync` failure. I'd rather solo degraded to console-only logging with a loud warning than refused to run at all — a broken log destination shouldn't stop `solo deployment create`.

Second, and more concretely: the issue this closes is `solo --version`. On 0.83.0 the reporter *did* get the version printed, then got buried in boxes. After this change they get no version at all, because the `-v/--version` branch in `src/index.ts:78` runs after the container init that now throws from here. The version banner needs `getSoloVersion()` and stdout, nothing else.

Handling `--version`/`--help` before `Container.getInstance().init(...)` and writing them with `process.stdout.write` would make them work no matter what state `~/.solo/logs` is in. User experience first — the user's goal in #5370 was to read the version number.

src/core/logging/solo-pino-logger.ts
```diff
@@ -382,6 +377,35 @@ export class SoloPinoLogger implements SoloLogger {
...
+   *
+   * The stream implementations report this asynchronously, after the constructor has already returned, so
+   * it escapes as an uncaught exception instead of an actionable message — and because reporting it writes
+   * another log line, it can repeat without bound. Checking up front turns the whole class of failure into
+   * one coded error naming the offending path.
+   */
+  private static assertLogDestinationsWritable(logsDirectory: string, fileNames: string[]): void {
+    try {
+      mkdirSync(logsDirectory, {recursive: true});
+      accessSync(logsDirectory, fileSystemConstants.W_OK);
```
JeffreyDallas, line 456: this will not detect the failure on Windows. `fs.accessSync(path, W_OK)` on Windows only reflects the read-only file attribute — it does not consult ACLs — so a directory the user genuinely cannot write passes the preflight and the failure surfaces asynchronously from `rotating-file-stream` further down the constructor. Your own tests acknowledge it by skipping the denial cases with `canDenyWrites`.

That leaves Windows users (and anyone whose disk fills mid-run) back on SOLO-9014 "internal Solo error, file a bug report" — the exact message #5370 complained about. Neither `createStream(...)` nor `pino.destination(...)` in this constructor gets an `'error'` listener, so nothing translates the async failure.

Could we attach one to both rotating streams and route it through the same coded error? Something like:

```ts
ndjsonStream.on('error', (streamError: Error): void => {
  FatalErrorReporter.reportWithoutLogger(
    new SoloErrors.system.soloLogsDirectoryNotWritable(PathEx.join(logsDirectory, ndjsonFileName), streamError),
  );
});
```

That gives every platform the actionable SOLO-5083 rather than only the ones where `accessSync` happens to be meaningful, and it keeps the preflight as the fast path rather than the only path.

src/index.ts
```diff
@@ -35,15 +41,23 @@ export async function main(argv: string[], context?: {logger: SoloLogger}): Prom
...
+    if (context?.logger) {
       context.logger.showUserError(error);
     } else {
-      console.error(`Error initializing container: ${error?.message}`, error);
+      // Initialization builds the logger, so a failure here often means there is nothing to log through.
+      FatalErrorReporter.reportWithoutLogger(error);
     }
-    throw error;
+    // The failure is rendered above; a SilentBreak keeps the entrypoint from rendering it a second time.
+    throw new SilentBreak(error.message);
```
JeffreyDallas: `SilentBreak`'s constructor is `(message: string)` and calls `super(message)`, so this throws away `code`, `cause` and `troubleshootingSteps` and keeps only the string. The PR description says the real error is kept as `cause` rather than discarded — true for `UncaughtFatalErrorSoloError`, not here. `main()` is also called programmatically (the entrypoint and the tests), and those callers used to receive the coded `SoloError`.

Widening `SilentBreak` to carry the cause keeps the break semantics intact — `ErrorHandler.extractBreak` tests the top-level instance first, so it still short-circuits — while preserving the code for anything that inspects the chain:

```ts
    throw new SilentBreak(error.message, error);
```

(`SoloError`'s constructor only re-throws the cause when message *and* `name` match, and the names differ here, so this is safe.)

src/core/fatal-error-reporter.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * Renders the first escaped error through the logger and routes any subsequent one to stderr.
+   * Always marks the process as failed so a caller cannot mistake this for a successful run.
+   *
+   * @param logger - the logger to render the first failure through
+   * @param kind - which handler fired: `uncaughtException` or `unhandledRejection`
+   * @param escapedError - the value the handler received; normalized to an `Error` when it is not one
+   */
+  public static report(logger: SoloLogger, kind: string, escapedError: unknown): void {
```
JeffreyDallas: `kind: string` accepts anything, and it ends up verbatim in the user-visible message (`Unhandled ${kind}`) and in `meta`. There are exactly two legal values. A named union — `FatalErrorKind = 'uncaughtException' | 'unhandledRejection'` in its own file per §3.5 — used here and on `UncaughtFatalErrorSoloError`'s constructor makes a typo a compile error instead of a rendered typo.

src/core/fatal-error-reporter.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Reports errors that escaped every handler (`uncaughtException` and `unhandledRejection`).
+ *
+ * Reporting means logging, and logging is itself a thing that can be broken: when the log destination
+ * cannot be written, every write raises a fresh error that lands straight back here, and reporting it
+ * raises another. Reporting is therefore latched — the first failure renders normally and every later one
+ * goes to stderr — so a failing logger cannot drive an unbounded report loop.
+ */
+export class FatalErrorReporter {
+  private static reported: boolean = false;
```
JeffreyDallas: the latch is set after the first *successful* render, not after the logger is proven broken. So a second, entirely unrelated fatal error later in a long `solo network deploy` run is reduced to `solo: unhandled uncaughtException: <message>` on stderr — no code, no stack, no troubleshooting steps — even when the logger is healthy.

I understand why it can't be conditional on `showUserError` throwing: in #5370 the stream error was asynchronous and `showUserError` returned normally, so "latch when the logger throws" wouldn't have bounded that loop. Would a small bound rather than 1 (say 3) get both properties — the loop still cut off, genuine multiple failures still rendered?

src/core/zippy.ts
```diff
@@ -13,18 +13,17 @@ import {InjectTokens} from './dependency-injection/inject-tokens.js';
...
   }
 
   /**
    * Zip a file or directory
-   * @param srcPath - path to a file or directory
-   * @param destPath - path to the output zip file
-   * @param [verbose] - if true, log the progress
    * @returns path to the output zip file
+   * @param sourcePath
+   * @param destinationPath
```
JeffreyDallas:
```suggestion
   * @param destinationPath - path to the output zip file
```
the descriptions were there before and got replaced by bare tags — same for `@param sourcePath` on the line above. Empty `@param` tags document less than no tag at all.

Stepping back: why is `zippy.ts` in a logging-initialisation PR? Dropping the unused `_verbose` and the `as any` casts is good cleanup, but it makes this diff harder to reason about and it isn't covered by the issue. I'd split it into its own PR.

solo.ts
```diff
@@ -17,7 +17,15 @@ await fnm
     context.logger?.info('Solo CLI completed, via entrypoint');
   })
   .catch((error): void => {
-    const errorHandler: ErrorHandler = container.resolve(InjectTokens.ErrorHandler);
+    let errorHandler: ErrorHandler;
+    try {
+      errorHandler = container.resolve(InjectTokens.ErrorHandler);
+    } catch {
```
JeffreyDallas, line 24: the comment explains the intended case well. My worry is the unintended one: this also swallows any *other* DI failure (a bad token, a constructor throwing for an unrelated reason), and the user then gets exit 1 with zero output and nothing in the log, which is harder to diagnose than the stack would have been.

One line before the `return` — printed only when the incoming error is not a `SilentBreak`, so the already-reported SOLO-5083 path stays quiet — would close that hole without reintroducing double reporting.

test/unit/core/logging/solo-pino-logger.test.ts
```diff
@@ -352,3 +355,108 @@ describe('SoloPinoLogger stream configuration', (): void => {
     expect(internalsOf(logger).rotatingStreams).to.have.lengthOf(0);
   });
 });
+
+// Typed view over the writability preflight, which runs before any stream is constructed.
+type LoggerPreflight = {
+  assertLogDestinationsWritable: (logsDirectory: string, fileNames: string[]) => void;
+};
+
+function preflightOf(): LoggerPreflight {
```
JeffreyDallas, line 371: these five cases are good, and testing the private static through a cast is consistent with `internalsOf` above. But every one of them calls `assertLogDestinationsWritable` directly, so nothing asserts that the *constructor* actually runs it, or runs it before the streams are built. Reorder the two statements at line 66-67 and this suite still passes green.

One case constructing `new SoloPinoLogger(...)` against a locked `SOLO_LOGS_DIR` and asserting SOLO-5083 escapes would pin the contract that matters. Same gap for `main()`'s new init-failure behaviour (exit code 1 + `SilentBreak`) and `solo.ts`'s resolve fallback — both are pure logic and cheap to unit test.

src/core/logging/solo-pino-logger.ts
jan-milenkov: updated

src/core/logging/solo-pino-logger.ts
jan-milenkov: updated

src/index.ts
jan-milenkov: updated

src/core/fatal-error-reporter.ts
jan-milenkov: updated

src/core/fatal-error-reporter.ts
jan-milenkov: updated

solo.ts
jan-milenkov: updated

test/unit/core/logging/solo-pino-logger.test.ts
jan-milenkov: updated

src/core/zippy.ts
```diff
@@ -36,24 +35,24 @@
...
-      const zip = new AdmZip('', {});
+      const zip: AdmZip = new AdmZip('', {});
 
-      const stat = fs.statSync(sourcePath);
+      const stat: fs.Stats = fs.statSync(sourcePath);
       if (stat.isDirectory()) {
         zip.addLocalFolder(sourcePath, '');
       } else {
-        zip.addFile(path.basename(sourcePath), fs.readFileSync(sourcePath), '', stat as any);
+        zip.addFile(path.basename(sourcePath), fs.readFileSync(sourcePath), '', stat);
```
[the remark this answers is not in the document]
jan-milenkov: updated

src/index.ts
```diff
@@ -52,69 +74,15 @@ export async function main(argv: string[], context?: {logger: SoloLogger}): Prom
...
     context.logger = logger;
   }
-  process.on('unhandledRejection', (reason: {error?: Error; target?: {url?: string}}, promise): void => {
-    logger.showUserError(
-      new SoloErrors.internal.commandReturnedFalse(
-        `Unhandled Rejection at: ${JSON.stringify(promise)}`,
-        `reason: ${JSON.stringify(reason)}`,
-      ),
-    );
+  process.on('unhandledRejection', (reason: unknown): void => {
```
JeffreyDallas: the `reset()` comment above reasons specifically about the E2E tests calling `main()` many times in one process — but these two handlers are registered on every call and never removed, so those same repeat invocations accumulate listeners.

After N calls, one `uncaughtException` fans out to N `report()` calls: the first 3 render in full and the remaining N-3 print a bare stderr line. So the failure mode the reset was meant to prevent ("reduced to a bare stderr line") arrives anyway, just from the other direction — and it's the *same* error reported N times rather than N distinct ones.

This predates the PR, so I'm not asking you to fix it here. But since the comment is new and explicitly reasons about that scenario, either registering once (module scope, or guarded) or a note that listener accumulation is a separate known issue would keep the comment from promising more than it delivers.

src/core/logging/solo-pino-logger.ts
```diff
@@ -419,6 +445,53 @@ export class SoloPinoLogger implements SoloLogger {
...
+  }
+
+  /**
+   * Routes an asynchronous write failure on a log stream through the same coded error as the preflight.
+   *
+   * Without this, a destination that only fails once writing starts — a Windows ACL the preflight cannot
+   * see, or a disk that fills mid-run — surfaces as an unhandled stream error rather than as SOLO-5083.
+   */
+  private static reportStreamFailures(stream: NodeJS.EventEmitter, filePath: string): void {
+    stream.on('error', (streamError: Error): void => {
```
JeffreyDallas: closing the Windows/ENOSPC gap here is right, but this listener is the one path that can render without any cap. `report()` got `MAX_RENDERED_REPORTS = 3`; `renderToStandardError` has no counter, and each call writes the message plus five troubleshooting steps plus the doc URL — roughly eight lines. `sonic-boom` (what `pino.destination` returns in the `CI === 'true'` branch above) emits `'error'` from its per-write path, so a destination that fails on one flush usually fails on all of them. That's the #5370 flood again, moved from the log file to stderr — and CI is where 271 MB of it hurt last time.

**Not `once`, though.** I tried that first and it's worse: removing the listener means the next failed write hits an `EventEmitter` with no `'error'` handler, which throws — so a full disk would turn a warning into a crash. I confirmed it both ways against this file; with `once` the second `emit('error', …)` throws `ENOSPC: no space left on device, write` straight out of the emitter.

Keep the listener attached and guard the render instead:

```suggestion
  private static reportStreamFailures(stream: NodeJS.EventEmitter, filePath: string): void {
    let reported: boolean = false;
    stream.on('error', (streamError: Error): void => {
      // The listener stays attached for the life of the stream: detaching it after the first failure
      // would make the next failed write an unhandled 'error' event, which crashes the process. Once
      // the destination is known bad every later failure says the same thing, so only the first is
      // rendered — otherwise a stream that fails on every flush reproduces the report flood of #5370.
      if (reported) {
        return;
      }
      reported = true;
      FatalErrorReporter.renderToStandardError(
        new SoloErrors.system.soloLogsDirectoryNotWritable(filePath, streamError),
        'WARNING',
      );
    });
  }
```

This compiles and leaves all 30 existing `solo-pino-logger.test.ts` cases green.

Nothing currently exercises this listener, and the two-emit case is exactly what distinguishes the three candidate implementations. This one passes against the guard above and fails against `once`:

```ts
it('renders one warning per stream and survives a repeated failure', (): void => {
  const stream: EventEmitter = new EventEmitter();
  reporterOf().reportStreamFailures(stream, '/tmp/solo/logs/solo.ndjson');

  stream.emit('error', new Error('ENOSPC: no space left on device, write'));
  const afterFirst: number = standardErrorWrite.callCount;

  // A destination that fails once tends to fail on every later flush; the second must not throw.
  expect((): boolean => stream.emit('error', new Error('ENOSPC: no space left on device, write'))).to.not.throw();

  expect(afterFirst).to.be.greaterThan(0);
  expect(standardErrorWrite.callCount).to.equal(afterFirst);
  expect(String(standardErrorWrite.firstCall.args[0])).to.include('SOLO-5083');
});
```

(`reporterOf()` being the same private-static cast shape as `preflightOf()` at line 369.)

src/core/logging/solo-pino-logger.ts
```diff
@@ -419,6 +445,53 @@ export class SoloPinoLogger implements SoloLogger {
...
+   * written, and `undefined` when the destination is usable.
+   *
+   * The stream implementations report this asynchronously, after the constructor has already returned, so
+   * it escapes as an unactionable internal error — and because reporting it writes another log line, it can
+   * repeat without bound. Probing up front turns the whole class of failure into one coded error naming the
+   * offending path, which the constructor renders before falling back to console-only logging.
+   */
+  private static findLogDestinationFailure(logsDirectory: string, fileNames: string[]): SoloError | undefined {
+    try {
+      mkdirSync(logsDirectory, {recursive: true});
```
JeffreyDallas: duplicate of the `mkdirSync` at line 101, and this one drops `mode: LOG_DIRECTORY_MODE`. In production line 101 runs first so the directory still ends up `0700`, which makes this call dead — but it's the one the preflight tests actually exercise (`creates the logs directory when it does not exist yet`), so deleting it isn't the fix.

Carrying the mode here instead lets lines 99-103 go away entirely, which also removes a swallowed catch:

```suggestion
      mkdirSync(logsDirectory, {recursive: true, mode: SoloPinoLogger.LOG_DIRECTORY_MODE});
```

With that applied I deleted the constructor's `try { mkdirSync(...) } catch {}` block; `tsc` is clean and all 30 logger tests still pass. See CLAUDE.md §Dead Code Removal.

src/core/errors/error-code-registry.ts
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
-export const ErrorCodeRegistry: Record<string, string> = {
+export const ErrorCodeRegistry = {
```
JeffreyDallas: dropping the annotation trades a clean file for a new `@typescript-eslint/typedef` warning ("Expected ErrorCodeRegistry to have a type annotation") — it's the only new lint output the PR adds. I put the annotation back and ran `npx tsc --noEmit`: clean, so nothing in the PR needs the literal types that `as const` alone provides.

```suggestion
export const ErrorCodeRegistry: Record<string, string> = {
```

If you *do* want the literal types repo-wide that's a worthwhile change, just not one that belongs in a `--version` fix — it widens the blast radius of this PR to every error class.

src/index.ts
JeffreyDallas: the `reset()` comment above reasons specifically about the E2E tests calling `main()` many times in one process — but these two handlers are registered on every call and never removed, so those same repeat invocations accumulate listeners.

After N calls, one `uncaughtException` fans out to N `report()` calls: the first 3 render in full and the remaining N-3 print a bare stderr line. So the failure mode the reset was meant to prevent ("reduced to a bare stderr line") arrives anyway, just from the other direction — and it's the *same* error reported N times rather than N distinct ones.

This predates the PR and I don't think it should block it. But since the comment is new and explicitly reasons about that scenario, either registering once (module scope, or guarded) or a note that listener accumulation is a separate known issue would keep the comment from promising more than it delivers.

src/core/logging/solo-pino-logger.ts
JeffreyDallas, line 485: closing the Windows/ENOSPC gap here is right, but this listener is the one path that can render without any cap. `report()` got `MAX_RENDERED_REPORTS = 3`; `renderToStandardError` has no counter, and each call writes the message plus five troubleshooting steps plus the doc URL — roughly eight lines. `sonic-boom` (what `pino.destination` returns in the `CI === 'true'` branch above) emits `'error'` from its per-write path, so a destination that fails on one flush usually fails on all of them. That's the #5370 flood again, moved from the log file to stderr — and CI is where 271 MB of it hurt last time.

**Not `once`, though.** I tried that first and it's worse: removing the listener means the next failed write hits an `EventEmitter` with no `'error'` handler, which throws — so a full disk would turn a warning into a crash. I confirmed it both ways against this file; with `once` the second `emit('error', …)` throws `ENOSPC: no space left on device, write` straight out of the emitter.

Keep the listener attached and guard the render instead:

```suggestion
  private static reportStreamFailures(stream: NodeJS.EventEmitter, filePath: string): void {
    let reported: boolean = false;
    stream.on('error', (streamError: Error): void => {
      // The listener stays attached for the life of the stream: detaching it after the first failure
      // would make the next failed write an unhandled 'error' event, which crashes the process. Once
      // the destination is known bad every later failure says the same thing, so only the first is
      // rendered — otherwise a stream that fails on every flush reproduces the report flood of #5370.
      if (reported) {
        return;
      }
      reported = true;
      FatalErrorReporter.renderToStandardError(
        new SoloErrors.system.soloLogsDirectoryNotWritable(filePath, streamError),
        'WARNING',
      );
    });
  }
```

This compiles and leaves all 30 existing `solo-pino-logger.test.ts` cases green.

Nothing currently exercises this listener, and the two-emit case is exactly what distinguishes the three candidate implementations. Dropping this at the end of `solo-pino-logger.test.ts` needs no new imports — `sinon`, `SinonStub`, `EventEmitter` and `SoloPinoLogger` are all already imported there. It passes against the guard above and fails against `once`:

```ts
// Typed view over the stream-failure reporter, which is a private static like the preflight above.
type StreamFailureReporter = {
  reportStreamFailures: (stream: NodeJS.EventEmitter, filePath: string) => void;
};

function reporterOf(): StreamFailureReporter {
  return SoloPinoLogger as unknown as StreamFailureReporter;
}

describe('SoloPinoLogger stream failure reporting', (): void => {
  let standardErrorWrite: SinonStub;

  beforeEach((): void => {
    standardErrorWrite = sinon.stub(process.stderr, 'write').returns(true);
  });

  afterEach((): void => {
    sinon.restore();
  });

  it('renders one warning per stream and survives a repeated failure', (): void => {
    // eslint-disable-next-line unicorn/prefer-event-target
    const stream: EventEmitter = new EventEmitter();
    reporterOf().reportStreamFailures(stream, '/tmp/solo/logs/solo.ndjson');

    stream.emit('error', new Error('ENOSPC: no space left on device, write'));
    const afterFirst: number = standardErrorWrite.callCount;

    // A destination that fails once tends to fail on every later flush; the second must not throw.
    expect((): boolean => stream.emit('error', new Error('ENOSPC: no space left on device, write'))).to.not.throw();

    expect(afterFirst).to.be.greaterThan(0);
    expect(standardErrorWrite.callCount).to.equal(afterFirst);
    expect(String(standardErrorWrite.firstCall.args[0])).to.include('SOLO-5083');
  });
});
```

The `eslint-disable` on `new EventEmitter()` is required, not decorative — `unicorn/prefer-event-target` is an error here otherwise, same as on `createFakeStream` at line 29. I appended this block verbatim to `solo-pino-logger.test.ts`: `tsc` clean, `eslint` clean, 31 passing.

src/core/logging/solo-pino-logger.ts
JeffreyDallas: duplicate of the `mkdirSync` at line 101, and this one drops `mode: LOG_DIRECTORY_MODE`. In production line 101 runs first so the directory still ends up `0700`, which makes this call dead — but it's the one the preflight tests actually exercise (`creates the logs directory when it does not exist yet`), so deleting it isn't the fix.

Carrying the mode here instead lets lines 99-103 go away entirely, which also removes a swallowed catch:

```suggestion
      mkdirSync(logsDirectory, {recursive: true, mode: SoloPinoLogger.LOG_DIRECTORY_MODE});
```

With that applied I deleted the constructor's `try { mkdirSync(...) } catch {}` block; `tsc` is clean and all 30 logger tests still pass. See CLAUDE.md §Dead Code Removal.

test/unit/core/version-banner.test.ts
```diff
@@ -0,0 +1,78 @@
...
+  afterEach((): void => {
+    sinon.restore();
+  });
+
+  function written(): string {
+    return standardOutputWrite
+      .getCalls()
+      .map((call): string => String(call.args[0]))
+      .join('');
+  }
```
JeffreyDallas, line 34: stubbing `process.stdout.write` for the whole test also stubs the stream mocha's spec reporter writes to, and the reporter emits a passing test's title *before* `afterEach` restores it. `npx mocha 'test/unit/core/version-banner.test.ts'` prints the `VersionBanner` heading, then nothing, then `11 passing` — no case names at all.

It's the only place in `test/` that stubs stdout; `main-init-failure.test.ts` and `fatal-error-reporter.test.ts` stub `process.stderr`, which the reporter doesn't touch, so they're unaffected.

Restoring as part of reading gets the titles back:

```suggestion
  function written(): string {
    const output: string = standardOutputWrite
      .getCalls()
      .map((call): string => String(call.args[0]))
      .join('');
    standardOutputWrite.restore();
    return output;
  }
```

I ran this: 9 of the 11 titles come back. The two that assert on the stub directly rather than through `written()` — line 29 and line 44 — need a `standardOutputWrite.restore()` after their assertion to match.

src/core/errors/error-code-registry.ts
JeffreyDallas: dropping the annotation trades a clean file for a new `@typescript-eslint/typedef` warning ("Expected ErrorCodeRegistry to have a type annotation") — it's the only new lint output the PR adds. I put the annotation back and ran `npx tsc --noEmit`: clean, so nothing in the PR needs the literal types that `as const` alone provides.

```suggestion
export const ErrorCodeRegistry: Record<string, string> = {
```

If you *do* want the literal types repo-wide that's a worthwhile change, just not one that belongs in a `--version` fix — it widens the blast radius of this PR to every error class.

src/core/version-banner.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    const outputArgument: string = argv[outputFlagIndex];
+    if (outputArgument.startsWith('--output=')) {
+      return outputArgument.split('=', 2)[1] ?? '';
+    }
+    return outputFlagIndex + 1 < argv.length ? argv[outputFlagIndex + 1] : '';
+  }
+
+  private static write(line: string): void {
```
JeffreyDallas, line 74: behavior change worth naming: the old path went `logger.showUser()` → `writeUser()`, which returns early when `constants.SOLO_SILENT_MODE` is set (`solo-pino-logger.ts:276`). So `SOLO_SILENT_MODE=true solo --version` printed nothing before and prints the banner now.

I think the new behavior is the correct one — silencing a value the user explicitly asked for is worse than the alternative, and `--output=json` consumers were getting empty output. But it's observable for anyone scripting around it, so it should be called out in the PR description rather than discovered. If you'd rather preserve it, it's a guard here.

(The other lost side effect — `showUser` also did `this.info(formatted)`, persisting the version to the log — is clearly right to drop, since the whole point is not touching the logger.)

src/index.ts
JeffreyDallas: the `reset()` comment above reasons specifically about the E2E tests calling `main()` many times in one process — but these two handlers are registered on every call and never removed, so those same repeat invocations accumulate listeners.

After N calls, one `uncaughtException` fans out to N `report()` calls: the first 3 render in full and the remaining N-3 print a bare stderr line. So the failure mode the reset was meant to prevent ("reduced to a bare stderr line") arrives anyway, just from the other direction — and it's the *same* error reported N times rather than N distinct ones.

This predates the PR and I don't think it should block it. But since the comment is new and explicitly reasons about that scenario, either registering once (module scope, or guarded) or a note that listener accumulation is a separate known issue would keep the comment from promising more than it delivers.

JeffreyDallas (dismissed):

src/core/logging/solo-pino-logger.ts
jan-milenkov: updated

src/core/logging/solo-pino-logger.ts
jan-milenkov: updated

test/unit/core/version-banner.test.ts
jan-milenkov: updated

src/core/errors/error-code-registry.ts
jan-milenkov: updated

src/core/version-banner.ts
jan-milenkov: updated

src/index.ts
jan-milenkov: updated

JeffreyDallas (approved):

Issue #5731, opened 2026-08-06
test: E2E installer validation — Linux



jeromy-cannon:
On a fresh Linux CI runner:
1. Install Solo using the Linux installer
2. Verify the post-install `solo cache image pull` completed successfully
3. Run `solo one-shot` (deploy → validate → destroy) end-to-end
4. Confirm the network lifecycle completes without errors from a clean install state

Part of #5714.

### Dependencies
- [ ] requires #5725
- [ ] requires #5727

Issue #5730, opened 2026-08-06
test: E2E installer validation — macOS



jeromy-cannon:
On a fresh macOS CI runner:
1. Install Solo using the DMG installer
2. Verify the post-install `solo cache image pull` completed successfully
3. Run `solo one-shot` (deploy → validate → destroy) end-to-end
4. Confirm the network lifecycle completes without errors from a clean install state

Part of #5714.

### Dependencies
- [ ] requires #5724
- [ ] requires #5727

Issue #5729, opened 2026-08-06
test: E2E installer validation — Windows



jeromy-cannon:
On a fresh Windows CI runner:
1. Install Solo using the NSIS installer
2. Verify the post-install `solo cache image pull` completed successfully
3. Run `solo one-shot` (deploy → validate → destroy) end-to-end
4. Confirm the network lifecycle completes without errors from a clean install state

Part of #5714.

### Dependencies
- [ ] requires #5723
- [ ] requires #5727

PR #5810, opened 2026-08-11
feat: add node.js sea build pipeline

### LINKED ISSUE #5716: feat: implement Node.js SEA build pipeline
Configure and build the Node.js [Single Executable Application](https://nodejs.org/api/single-executable-applications.html) blob. Integrate the SEA build step into `task build` and validate that the resulting binary runs as a fully standalone executable on Linux, macOS, and Windows.

This is the critical-path task — all OS-specific installer work depends on it.

Part of #5714.

### Dependencies
- [ ] required by #5723
- [ ] required by #5724
- [ ] required by #5725
- [ ] required by #5728

Ivo-Yankov:
## Description

Adds a Node.js Single Executable Application (SEA) build pipeline for the solo CLI.

- sea/build.ts — new build script that bundles dist/src/index.js with esbuild (CJS), generates a SEA blob via node --experimental-sea-config, injects it into a copy of the
current Node binary via postject, and re-signs on macOS. Output: sea/dist/solo-<platform>-<arch>[.exe]
- Taskfile.yml — new sea:build task, hooked into task build
- .github/workflows/flow-build-sea.yaml — CI matrix that builds and smoke-tests the binary on Linux, macOS, and Windows

### Related Issues

* Closes #5716 
* Related to #

### Testing

* TBD

The following was not tested:

* TBD

<details>
<summary>
Commit message guidelines
</summary>

</details>


Taskfile.yml
```diff
@@ -92,15 +92,16 @@ tasks:
...
     deps: [npm:install]
     sources:
       - src/**/*.ts
-      - test/**/*.tss 
+      - test/**/*.tss
     cmds:
       - task: "build:compile"
       - task: "build:lint"
       - task: "check:deprecations"
+      - task: "sea:build"
```
ddeskov-limechain, line 112: The release workflow runs `task build` and then publishes the npm package from the repo root. With this change, `task build` also creates the SEA binary and blob under `sea/dist`. `.npmignore` does not exclude `sea/`, so these large files will end up in the npm tarball. Please add `sea/` to `.npmignore`:

```
# SEA build output
sea/
```

sea/build.ts
```diff
@@ -0,0 +1,254 @@
...
+
+  if (needsExtraction) {
+    for (const key of ${JSON.stringify(assetKeys)}) {
+      let data;
+      try { data = sea.getAsset(key); } catch { continue; }
+      const destPath = path.join(seaRoot, key);
+      fs.mkdirSync(path.dirname(destPath), { recursive: true });
+      fs.writeFileSync(destPath, Buffer.from(data));
+    }
+    fs.writeFileSync(markerPath, SOLO_SEA_VERSION);
```
ddeskov-limechain: The marker only stores the version. Two builds from different commits with the same version (for example the CI artifacts from this workflow) share the same cache directory, so the second binary keeps using the old extracted bundle. Suggestion: include a build id in the marker.

```suggestion
  // Skip extraction when the marker already records this exact build.
  // The id includes a build timestamp so a rebuild with the same version
  // (for example a CI artifact from another commit) still re-extracts.
  const SOLO_SEA_BUILD_ID = ${JSON.stringify(version + '-' + Date.now())};
  const markerPath = path.join(seaRoot, '.sea-extracted');
  let needsExtraction = true;
  try { needsExtraction = fs.readFileSync(markerPath, 'utf8').trim() !== SOLO_SEA_BUILD_ID; }
  catch { /* not yet extracted */ }

  if (needsExtraction) {
    for (const key of ${JSON.stringify(assetKeys)}) {
      let data;
      try { data = sea.getAsset(key); } catch { continue; }
      const destPath = path.join(seaRoot, key);
      fs.mkdirSync(path.dirname(destPath), { recursive: true });
      fs.writeFileSync(destPath, Buffer.from(data));
    }
    fs.writeFileSync(markerPath, SOLO_SEA_BUILD_ID);
```

src/integration/kube/k8-client/resources/pod/k8-client-pod.ts
```diff
@@ -234,18 +234,22 @@ export class K8ClientPod implements Pod {
...
         const isTsx: boolean = __filename.endsWith('.ts');
         const persistScriptExtension: string = isTsx ? '.ts' : '.js';
-        const useDirectNodeRuntime: boolean = isWindows;
-        let persistCmd: string = isTsx ? 'tsx' : 'node';
+        const seaRootDirectory: string | undefined = constants.getEnvironmentVariable('SOLO_SEA_ROOT_DIR');
+
+        // SEA binary cannot run external JS files (process.execPath points to the SEA itself),
+        // so fall back to 'node' from PATH. Requires node to be available on PATH in SEA mode.
+        const useDirectNodeRuntime: boolean = isWindows && !seaRootDirectory;
+        let persistCmd: string = !seaRootDirectory && isTsx ? 'tsx' : 'node';
```
ddeskov-limechain, line 245: In SEA mode the persisted port-forward spawns `node` from PATH, so the standalone binary still needs a Node.js installation for this feature. Is this acceptable for the installer use case? If yes, can we track it in a follow-up issue?

ddeskov-limechain (changes requested):
The SEA pipeline looks good and the CI matrix proves it on all three platforms. Requesting changes for the npm packaging issue on `Taskfile.yml`. The other two comments are smaller.

src/integration/kube/k8-client/resources/pod/k8-client-pod.ts
Ivo-Yankov: Issue created: https://github.com/hiero-ledger/solo/issues/5844

ddeskov-limechain (dismissed):
LGTM

jan-milenkov (commented):
LGTM Please review @JeffreyDallas @jeromy-cannon

.github/workflows/flow-build-sea.yaml
```diff
@@ -0,0 +1,112 @@
...
+        run: |
+          if [[ "${{ matrix.platform }}" == "win32" ]]; then
+            BIN=$(find sea/dist -name "solo-win32-*.exe" | head -1)
+          else
+            BIN=$(find sea/dist -name "solo-${{ matrix.platform }}-*" ! -name "*.blob" ! -name "*.cjs" ! -name "*.json" | head -1)
+            chmod +x "$BIN"
+          fi
+          echo "Running: $BIN --help"
+          "$BIN" --help
+          echo "✓ Smoke test passed (exit code $?)"
```
JeffreyDallas: I suggest the smoke test should be expanded to include one shot deployment, some transactions, lastly query mirror node to to make sure it is working end to end.

JeffreyDallas (dismissed):

.github/workflows/flow-build-sea.yaml
Ivo-Yankov: This will be done at a later point with those issues:
- #5729
- #5730
- #5731

.github/workflows/flow-build-sea.yaml
```diff
@@ -0,0 +1,112 @@
...
+  cancel-in-progress: true
+
+jobs:
+  build-sea:
+    name: "SEA / ${{ matrix.os }}"
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
```
jeromy-cannon: instead of ubuntu-latest we should probably just use our self-hosted linux medium.  Unless you have a good reason to use the github hosted runner.

.github/workflows/flow-build-sea.yaml
```diff
@@ -0,0 +1,112 @@
...
+
+      - name: Build SEA Binary
+        run: npx tsx sea/build.ts
+
+      - name: Smoke Test — solo --help
+        # Verify the binary starts, resolves its embedded resources, and exits cleanly.
+        shell: bash
+        run: |
+          if [[ "${{ matrix.platform }}" == "win32" ]]; then
+            BIN=$(find sea/dist -name "solo-win32-*.exe" | head -1)
```
jeromy-cannon, line 109: I know that win32 means both x86 (32 bit) and x64 (64 bit), but it might not be intuitive and we might get questions from users about a 64 bit option.

Question, does the SEA installer need any configuration to work on 32 bit vs 64 bit?  will this single installer work on both architectures?

.github/workflows/flow-build-sea.yaml
```diff
@@ -0,0 +1,112 @@
...
+          else
+            BIN=$(find sea/dist -name "solo-${{ matrix.platform }}-*" ! -name "*.blob" ! -name "*.cjs" ! -name "*.json" | head -1)
```
jeromy-cannon, line 111: does the SEA binary for Darwin support the Intel and Silicon chipsets?

sea/build.ts
```diff
@@ -0,0 +1,258 @@
...
+// Step 3: generate the CJS SEA entry script.
+// sea-main.cjs is what Node.js executes from the SEA blob. It:
+//   - Bootstraps synchronously (sets env vars, extracts assets)
+//   - Requires solo-src-bundle.cjs from the extracted seaRoot
+//   - Calls soloModule.main() inside an async IIFE (CJS-compatible, no top-level await)
+// Note: Node.js SEA overrides require() to allow only built-in modules (embedderRequire).
+// Any non-built-in path passed to require() throws ERR_UNKNOWN_BUILTIN_MODULE.
+// Dynamic import() goes through the regular module loader and CAN load filesystem files,
+// so we use import() + pathToFileURL to load solo-src-bundle.cjs after extraction.
+const seaMainContent: string = `'use strict';
```
jeromy-cannon: I would prefer this just be a separate js or cjs that can be read in, that way when working with the file individually we could lint it, have syntax recognition, and other IDE features.

sea/build.ts
```diff
@@ -0,0 +1,258 @@
...
+// SEA bootstrap — synchronous CJS; sets env vars and extracts assets before loading solo.
+// NOTE: require() in SEA mode is restricted to Node.js built-ins; use import() for files.
+const sea = require('node:sea');
+const os = require('os');
+const path = require('path');
+const fs = require('fs');
+const url = require('url');
+
+const SOLO_SEA_VERSION = ${JSON.stringify(version)};
```
jeromy-cannon: I see you are wanting to do a string substitution, but you could also just run it through some simple templating (without trying to over-engineer it).  We might even be able to re-use some of other templating solutions, not sure.

sea/build.ts
```diff
@@ -0,0 +1,258 @@
...
+copyFileSync(process.execPath, binaryPath);
+console.log(`\n▶ Copied node binary → ${binaryName}`);
+
+if (platform === 'darwin') {
+  run(`codesign --remove-signature ${quotePath(binaryPath)}`, 'Removing existing macOS signature');
+}
+
+const machoFlag: string = platform === 'darwin' ? '--macho-segment-name NODE_SEA' : '';
+run(
+  `npx postject ${quotePath(binaryPath)} NODE_SEA_BLOB ${quotePath(blobPath)} --sentinel-fuse NODE_SEA_FUSE_fce680ab2cc467b6e072b8b5df1996b2 ${machoFlag}`,
```
jeromy-cannon: might be worth having a comment on what this `--sentinel-fuse NODE_SEA_FUSE_fce680ab2cc467b6e072b8b5df1996b2` does and means

sea/build.ts
Ivo-Yankov: Added a template file and I'm using a simple string replace to apply the correct values.

sea/build.ts
Ivo-Yankov: Added a `sea-main.template.cjs` file

.github/workflows/flow-build-sea.yaml
Ivo-Yankov: It is only for the Apple chipsets, so I added another workflow responsible for building the SEA binary for `darwin-x64`

.github/workflows/flow-build-sea.yaml
Ivo-Yankov: Currently it only generates a binary for windows 64 bit, we'll need to do further testing to verify a 32-bit version.

Taskfile.yml
JeffreyDallas, line 112: `CLAUDE.md` tells every contributor to run `task build` before pushing, and `flow-deploy-release-artifact.yaml:285,433` runs `task build` in the release publishing pipeline.

I confirmed this hard-fails on Homebrew Node (very common on macOS):

```
$ node -v && node -p "process.config.variables.single_executable_application"
v26.7.0
false
$ npx tsx sea/build.ts
▶ Generating SEA blob
Single executable application is disabled.
Error: Command failed: node --experimental-sea-config .../sea-config.json
```

Homebrew builds node with `--shared`, which disables SEA. The same run succeeds on official node v22.22.1.

I first wrote this up as "move `sea:build` out of `task build`", but #5716 explicitly asks you to "Integrate the SEA build step into `task build`" — so that would be a scope change, not a fix. The option that preserves #5716's intent: have `sea:build` check `process.config.variables.single_executable_application` up front and **skip with a warning** when the running Node can't produce a SEA.

```ts
// single_executable_application isn't declared on @types/node's ProcessConfig,
// so narrow it explicitly rather than reaching through an untyped property.
const seaSupported: boolean = (process.config.variables as {single_executable_application?: boolean})
  .single_executable_application === true;
if (!seaSupported) {
  console.warn('Skipping SEA build: this Node.js build has SEA disabled (e.g. Homebrew --shared). Use an official Node.js build to produce a binary.');
  process.exit(0);
}
```

(The direct `process.config.variables.single_executable_application` read fails `tsc` with TS2339 — `sea/` is in `tsconfig.json`'s `include`, so it must type-check. I verified both forms.)

Contributors on Homebrew Node keep a working `task build`; CI on official Node still produces the binary. Worth deciding separately whether the release-packaging path (`flow-deploy-release-artifact.yaml:285,433`) should build a 128 MB binary that `.npmignore` then discards.

sea/build.ts
```diff
@@ -0,0 +1,199 @@
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Builds the Solo Node.js Single Executable Application (SEA) for the current platform.
+ *
+ * Pipeline:
+ *   1. esbuild bundles dist/src/index.js (exports main(), no top-level await) into a CJS file
+ *   2. A generated CJS SEA entry (sea-main.cjs) bootstraps synchronously, then requires the bundle
```
JeffreyDallas: This says the entry **requires** the bundle, and line 24 repeats it ("Calls `require(seaRoot/solo-src-bundle.cjs)` — synchronous CJS"). The implementation uses dynamic `import()`, and the comment further down in this same file explicitly explains that `require()` *cannot* work in SEA mode:

> Node.js SEA overrides require() to allow only built-in modules (embedderRequire). Any non-built-in path passed to require() throws ERR_UNKNOWN_BUILTIN_MODULE.

Looks left over from an earlier design. Please update both lines so the docblock doesn't teach the next reader the opposite of what the code does.

sea/build.ts
```diff
@@ -0,0 +1,199 @@
...
+ *   constants.ts reads SOLO_SEA_ROOT_DIR (set before require()) for ROOT_DIR.
+ *   version.ts reads SOLO_SEA_VERSION (set before require()) for getSoloVersion().
+ *
+ * Output: sea/dist/solo-<platform>-<arch>[.exe]
+ *
+ * Run with: npx tsx sea/build.ts
+ * Or via:   task sea:build   (which ensures build:compile ran first)
+ */
+
+// eslint-disable-next-line n/no-unpublished-import
```
JeffreyDallas:
```suggestion
```
`npx eslint sea/` reports this as an unused directive — `n/no-unpublished-import` doesn't fire here:

```
sea/build.ts
  35:1  warning  Unused eslint-disable directive (no problems were reported from 'n/no-unpublished-import')
```

sea/build.ts
```diff
@@ -0,0 +1,199 @@
...
+const bundlePath: string = path.join(BUILD_DIR, 'solo-src-bundle.cjs');
+const seaMainPath: string = path.join(BUILD_DIR, 'sea-main.cjs');
+const blobPath: string = path.join(BUILD_DIR, 'sea-prep.blob');
+const configPath: string = path.join(BUILD_DIR, 'sea-config.json');
+const binaryPath: string = path.join(BUILD_DIR, binaryName);
+
+/** Wraps a path in double quotes for safe shell interpolation on all platforms. */
+const quotePath: (filePath: string) => string = (filePath: string): string => `"${filePath}"`;
+
+function run(cmd: string, label: string): void {
```
JeffreyDallas: `execSync` + the hand-rolled `quotePath` on line 63 breaks on any repo path containing a `"` or (on Unix) a `$`. `execFileSync` with an argument array needs no quoting and is immune to this:

```ts
function run(command: string, commandArguments: string[], label: string): void {
  console.log(`\n▶ ${label}`);
  execFileSync(command, commandArguments, {stdio: 'inherit'});
}
```

That also lets `quotePath` go away entirely.

sea/sea-main.template.cjs
```diff
@@ -0,0 +1,92 @@
...
+    /* not yet extracted */
+  }
+
+  if (needsExtraction) {
+    for (const key of SOLO_SEA_ASSET_KEYS) {
+      let data;
+      try {
+        data = sea.getAsset(key);
+      } catch {
+        continue;
```
JeffreyDallas: Two issues in this extraction block.

**Silent skip + unconditional marker.** `catch { continue; }` swallows a failed `sea.getAsset()`, and line 51 writes the marker regardless. If the blob is missing or corrupt for `solo-src-bundle.cjs`, the marker records the build as extracted and *every subsequent invocation* fails at `import()`, with no recovery short of the user deleting `~/.solo/sea-resources`. A missing asset here isn't recoverable — let it throw.

To be precise about the blast radius: the `try` wraps only `getAsset()`, so `mkdirSync`/`writeFileSync` failures (disk full, EPERM) and interrupted runs propagate and abort *before* line 51 — those cases correctly re-extract on the next run. The wedge is specific to the swallowed `getAsset()` path.

**Concurrency.** Two `solo` invocations racing (common with port-forwards) both see `needsExtraction` and write the same files. Worth noting a temp-dir-plus-`renameSync` isn't sufficient by itself here, since renaming onto an already-populated `seaRoot` fails — this needs a lock file or a versioned directory that's swapped atomically.

Also, per CLAUDE.md "Catch Blocks — Comment Required When Swallowing Errors", a non-rethrowing catch needs a comment explaining the intent.

sea/sea-main.template.cjs
```diff
@@ -0,0 +1,92 @@
...
+  .then(function (mod) {
+    const soloModule = mod.default || mod;
+
+    // Run the solo CLI. async IIFE keeps this CJS-compatible (no top-level await).
+    void (async function () {
+      const context = {logger: undefined};
+
+      await soloModule.main(process.argv, context).catch(function (error) {
+        // SilentBreak / UserBreak are solo's clean-exit signals (--help, --version, user ^C).
+        // The DI ErrorHandler swallows them; replicate that here so the exit code stays 0.
```
JeffreyDallas: Two things here.

**The comment is inaccurate.** `ErrorHandler` does not swallow `UserBreak` — `handleUserBreak()` calls `this.logger.showUser(userBreak.message)` (`src/core/error-handler.ts:32-34`). The bootstrap genuinely does swallow it, so this comment presents a behavior change as parity.

**The bigger problem: this whole block re-implements `solo.ts` and loses most of it.** I ran the same failing command both ways.

`npx tsx solo.ts consensus network deploy --deployment does-not-exist-xyz`:
```
╭─ ERROR ─────────────────────────────────────────────────────────────────╮
│ [SOLO-3009] Error installing chart solo-deployment: Deployment ... not found │
│   → List available deployments: solo deployment config list             │
│   → Create a deployment if needed: solo deployment config create        │
│ Learn more: https://solo.hiero.org/docs/troubleshooting/errors/...      │
╰─────────────────────────────────────────────────────────────────────────╯
💡 Tip: To collect diagnostic information ... solo deployment diagnostics logs
```

`./solo-darwin-arm64` with the same arguments:
```
  meta: {},
  statusCode: 'SOLO-2003',
  code: 'SOLO-3009',
  troubleshootingSteps: [ 'Check Helm release status: helm list -n <namespace>', ... ],
  retryable: true,
  ownership: 'infrastructure'
}
```

Lost: the formatted box, the docs link, the diagnostics tip, and all four translators (`KubeErrorTranslator`/`CacheErrorTranslator`/`HelmErrorTranslator`/`SdkErrorTranslator`). Additionally `ErrorHandler.extractBreak()` recurses through `error.cause`, while this only checks the top-level `name` — so a `SilentBreak` wrapped as a cause exits 1 with a stack dump instead of 0.

Rather than maintaining two entry points that drift, extract the `solo.ts` tail (resolve `ErrorHandler`, `handle()`, flush the logger, exit) into a `static` method on a small class in `src/` and call it from both. Checklist §1 (DRY) — "this seems like a duplicate of `solo.ts`, perhaps you should just enhance that so both entry points share it."

src/core/constants.ts
```diff
@@ -36,7 +36,12 @@ export function getEnvironmentVariable(name: string): string | undefined {
   return undefined;
 }
 
-export const ROOT_DIR: string = PathEx.joinWithRealPath(path.dirname(url.fileURLToPath(import.meta.url)), '..', '..');
+// In SEA mode the bootstrap (sea/build.ts banner) sets SOLO_SEA_ROOT_DIR to
+// ~/.solo/sea-resources/<version>/ before any module initializes, so all
+// RESOURCES_DIR-based paths continue to resolve without code changes at each read site.
+export const ROOT_DIR: string =
+  process.env['SOLO_SEA_ROOT_DIR'] ??
```
JeffreyDallas: Two things.

**Stale comment.** There is no esbuild `banner` in `sea/build.ts` — the env var is set by `sea/sea-main.template.cjs:28`. Please fix the reference so the next reader can find it.

**`ROOT_DIR` is now overridable in every install, not just SEA.** `RESOURCES_DIR` is derived from this, and that is where solo reads Helm values files, CRDs, and `resources/config` from (`container-init.ts:295,302`). After this change a plain `npm i -g @hiero-ledger/solo` install will silently read all of that from wherever `SOLO_SEA_ROOT_DIR` points. Please gate it on actually being a SEA. Note the API is `node:sea`, not `process.sea` — my earlier phrasing was wrong, and `process.sea?.isSea?.()` would silently evaluate to `false` forever:

```ts
import {isSea} from 'node:sea';
// ...
export const ROOT_DIR: string =
  (isSea() && process.env['SOLO_SEA_ROOT_DIR']) ||
  PathEx.joinWithRealPath(path.dirname(url.fileURLToPath(import.meta.url)), '..', '..');
```

`@types/node/sea.d.ts` declares only the `node:sea` module; there is no `process.sea`. The comment at `sea/build.ts:188` says `process.sea.isSea()` too and should be corrected alongside this.

Note also the SEA branch skips `PathEx.joinWithRealPath`, so unlike the normal path it isn't real-path resolved or validated.

version.ts
```diff
@@ -90,6 +90,12 @@ export const MINIMUM_SOLO_CHART_VERSION: string = '0.64.0';
 export const MINIMUM_HIERO_BLOCK_NODE_VERSION_FOR_DEDICATED_HEALTH_PORT: string = 'v0.39.0-0';
 
 export function getSoloVersion(): Version {
+  // In SEA mode the bootstrap embeds package.json and sets SOLO_SEA_VERSION before any
+  // module initializes, so we can return the version without a filesystem read.
+  if (process.env['SOLO_SEA_VERSION']) {
```
JeffreyDallas: Same gating concern as `constants.ts`: this makes `solo --version` report whatever an env var says in any install, not just a SEA. Please gate it on `isSea()` from `node:sea` (not `process.sea`, which doesn't exist — see the note on `constants.ts`).

(I'm deliberately *not* asking you to route this through `getEnvironmentVariable()` — CLAUDE.md scopes that rule to `src/**`, and importing `src/core/constants.js` from root `version.ts` would risk a cycle. And once this is gated to SEA mode it's an internal bootstrap sentinel, so it needs no env var documentation either — docs only apply if you intend it to stay a public, ungated knob.)

Minor: the comment says "the bootstrap embeds `package.json` and sets `SOLO_SEA_VERSION` ... so we can return the version without a filesystem read" — the embedded `package.json` is never consulted on this path, so the two clauses aren't related.

src/integration/kube/k8-client/resources/pod/k8-client-pod.ts
```diff
@@ -234,18 +234,22 @@ export class K8ClientPod implements Pod {
...
         // In a compiled build it ends in .js; use node to run the compiled .js.
+        // In a SEA binary, the script is extracted to SOLO_SEA_ROOT_DIR/scripts/ at startup.
         const isTsx: boolean = __filename.endsWith('.ts');
         const persistScriptExtension: string = isTsx ? '.ts' : '.js';
-        const useDirectNodeRuntime: boolean = isWindows;
-        let persistCmd: string = isTsx ? 'tsx' : 'node';
+        const seaRootDirectory: string | undefined = constants.getEnvironmentVariable('SOLO_SEA_ROOT_DIR');
+
+        // SEA binary cannot run external JS files (process.execPath points to the SEA itself),
+        // so fall back to 'node' from PATH. Requires node to be available on PATH in SEA mode.
```
JeffreyDallas, line 243: "Requires node to be available on PATH in SEA mode" defeats the point of shipping a SEA — a user who installed the binary precisely so they wouldn't need Node gets a spawn failure the first time they use `--persist`.

This is already tracked as #5844, and step 3 there ("spawn `process.execPath --internal-sea-port-forward` when `SOLO_SEA_ROOT_DIR` is set") covers this branch on every platform, including the Windows `useDirectNodeRuntime` path on line 244. So nothing to change here — I've left the implementation discussion on that issue rather than duplicating it.

One ask for this PR: please link #5844 from the description, because "runs without a Node.js installation" isn't accurate for `--persist`.

.github/workflows/flow-build-sea.yaml
```diff
@@ -0,0 +1,122 @@
...
+            chmod +x "$BIN"
+          fi
+          echo "Running: $BIN --help"
+          "$BIN" --help
+          echo "✓ Smoke test passed (exit code $?)"
```
JeffreyDallas: Under `bash -e` this line only runs when `"$BIN" --help` already succeeded, so it always prints `exit code 0` — the value is noise. And `BIN=$(find ... | head -1)` isn't checked, so a missing binary produces `: --help: command not found` rather than a clear failure.

Also, `--help` passes even if the version embedding or the `ROOT_DIR` redirection is wrong. Worth strengthening:

```bash
[[ -n "$BIN" ]] || { echo "no SEA binary found in sea/dist"; exit 1; }
"$BIN" --help
EXPECTED=$(node -p "require('./package.json').version")
"$BIN" --version | grep -q "$EXPECTED" || { echo "version mismatch, expected $EXPECTED"; exit 1; }
```

Taskfile.yml
```diff
@@ -92,15 +92,16 @@ tasks:
         {{end}}
 
   build:
-    desc: "Build the project (compile and lint)"
+    desc: "Build the project (compile, lint, and SEA binary)"
     deps: [npm:install]
     sources:
       - src/**/*.ts
-      - test/**/*.tss 
+      - test/**/*.tss
```
JeffreyDallas: The `build` task fingerprints only `src/**/*.ts` (plus the pre-existing `test/**/*.tss` typo — worth fixing while you're here). Task skips the whole task when declared sources are unchanged, so with `sea:build` wired in as a cmd, edits confined to `sea/`, `resources/`, `package.json`, or `version.ts` never rebuild the binary.

Reproduced with a minimal Taskfile of the same shape:

```
--- run 1:                       task: [sea:build] echo "SEA BUILD RAN"
--- run 2 (no changes):          task: Task "build" is up to date
--- run 3 (only sea/ changed):   task: Task "build" is up to date   <-- stale
--- run 4 (src changed):         task: [sea:build] echo "SEA BUILD RAN"
```

`resources/**` and the `package.json` version are both embedded in the binary, so this silently yields stale assets or a wrong `--version`. CI can't catch it — a fresh checkout has no `.task` fingerprint and always runs cold.

Fixing this task's `sources` alone isn't sufficient. `build:compile` (lines 115-117) declares the same two globs, so editing `version.ts` or `tsconfig.json` leaves `dist/version.js` stale even when the parent does rerun — and the SEA bundle is built from `dist/`. Both need widening:

```yaml
    sources:
      - src/**/*.ts
      - test/**/*.ts
      - version.ts
      - solo.ts
      - tsconfig.json
```

plus `sea/**`, `resources/**`, `package.json`, and `package-lock.json` on whichever task owns the SEA step. Giving `sea:build` its own `sources`/`generates` is worth doing too, but doesn't help on its own: Task skips the parent before it ever evaluates the child.

sea/sea-main.template.cjs
```diff
@@ -0,0 +1,92 @@
+'use strict';
```
JeffreyDallas, line 2:
```suggestion
// SPDX-License-Identifier: Apache-2.0
'use strict';
```
§3.1 requires an SPDX identifier in all source files — `sea/build.ts` has one, this doesn't.

Worth knowing why it slipped through: the docblock a few lines down says keeping this as a real `.cjs` file gives it "normal editor syntax highlighting, linting, and IDE navigation". Highlighting and navigation yes, linting no — `eslint.config.mjs:123` globally ignores `'**/*.*js'`, so this file is never linted and `headers/header-format` never fired. Please either correct that sentence or add a targeted override so `sea/*.cjs` is actually linted.

sea/build.ts
```diff
@@ -0,0 +1,199 @@
...
+writeFileSync(seaMainPath, seaMainContent);
+
+// Step 4: write sea-config.json and generate the SEA blob from sea-main.cjs.
+const seaConfig: Record<string, unknown> = {
+  main: seaMainPath,
+  output: blobPath,
+  assets: seaAssets,
+  disableExperimentalSEAWarning: true,
+  useSnapshot: false,
+  useCodeCache: true,
```
JeffreyDallas: `useCodeCache: true` combined with the bootstrap's dynamic `import()` at `sea-main.template.cjs:66` is a combination the Node v22 SEA docs call out explicitly:

> **Note:** `import()` does not work when `useCodeCache` is `true`.

Being straight about the evidence: it works today. My local darwin/arm64 binary ran `--version`, `--help`, and a full command through that `import()`, and all four CI legs pass. So I'm not claiming it's broken — I'm flagging that the build rests on a combination Node documents as unsupported, which can regress silently on a Node upgrade.

There's also a good chance you're not getting what the flag is for: `useCodeCache` caches the **main script** only — here that's the ~90-line `sea-main.cjs`, not the 16 MB bundle where startup cost actually lives. So the upside is close to nil.

```suggestion
  useCodeCache: false,
```

If you'd rather keep the code cache, the alternative is to load the bundle with `createRequire()` from the extracted path instead of `import()` — but given the main script is a thin bootstrap, turning it off looks like the better trade.

sea/build.ts
```diff
@@ -0,0 +1,199 @@
...
+
+if (platform === 'darwin') {
+  run(`codesign --remove-signature ${quotePath(binaryPath)}`, 'Removing existing macOS signature');
+}
+
+const machoFlag: string = platform === 'darwin' ? '--macho-segment-name NODE_SEA' : '';
+// --sentinel-fuse is the fixed marker string Node.js embeds in its own binary at build time
+// (see NODE_SEA_FUSE in Node's deps/v8 fuse.h). postject searches the copied node binary for
+// this exact byte sequence and flips a bit next to it once the blob is injected — that bit is
+// what makes process.sea.isSea() (and Node's SEA bootstrap) return true at runtime. The value
```
JeffreyDallas: There is no `process.sea`. The documented detection API is the `node:sea` builtin — `@types/node/sea.d.ts` declares only `declare module "node:sea"` with `isSea()`, `getAsset()`, etc., and nothing on `process`.

```suggestion
// what makes sea.isSea() (and Node's SEA bootstrap) return true at runtime. The value
```

Worth correcting because this is the line a reader would copy when adding the SEA gate that Major #4 asks for — and `process.sea?.isSea?.()` fails open (always `false`) rather than erroring, so the mistake would be silent.

Taskfile.yml
```diff
@@ -127,6 +139,19 @@ tasks:
     deps: [npm:install]
     sources:
       - src/**/*.ts
-      - test/**/*.tss 
+      - test/**/*.ts
```
jan-milenkov, line 142: nice

Fantomasa (approved):
LGTM

jan-milenkov (approved):
LGTM! @JeffreyDallas please review too!

JeffreyDallas (approved):

Issue #5432, opened 2026-07-29
Performance Test > crane pull failed for minio operator image, process killed (exit 143)



jeromy-cannon:
## Failure Summary

The `one-shot single deploy` command failed during the Performance Test job. The `crane` tool failed to pull the `quay.io/minio/operator:v7.1.1` image, which triggered a cascade of `podman` and `sudo` errors. The task was ultimately terminated with exit status 143 (SIGTERM).

## Error Details

```
[07:05:23.803] ERROR: Failed to save image archive: Error: Executing command: /home/runner/.solo/bin/crane pull --format oci --annotate-ref --platform linux/amd64 quay.io/minio/operator:v7.1.1 /home/runner/.solo/cache/images/quay.io__minio__operator__v7.1.1.tar.oci-layout
```

## Failed Command

```
Current Command: one-shot single deploy --debug --quiet-mode --deployment performance-tests-deployment --metrics-server --namespace rss-hiero-solo-primary-linux-large --relay-version 0.77.0 --block-node-version 0.38.0
```

## Workflow Information

- **Run**: [#106](https://github.com/hiero-ledger/solo/actions/runs/30430094892) (attempt 1)
- **Job**: [Extended Test (flow-performance-test) / Performance Test](https://github.com/hiero-ledger/solo/actions/runs/30430094892/job/90505348800)
- **Branch**: `main`

## Log Archives

- [CI Logs (secret gist)](https://gist.github.com/jeromy-cannon/6dc5ac0d4922495e86abe12a6126ac76) — 4 files (job log + complete `one-shot-timing-logs` artifact)

<details>
<summary>All files</summary>

- [hashgraph-sdk.log](https://gist.githubusercontent.com/jeromy-cannon/6dc5ac0d4922495e86abe12a6126ac76/raw/hashgraph-sdk.log)
- [job-90505348800.log](https://gist.githubusercontent.com/jeromy-cannon/6dc5ac0d4922495e86abe12a6126ac76/raw/job-90505348800.log)
- [solo.log](https://gist.githubusercontent.com/jeromy-cannon/6dc5ac0d4922495e86abe12a6126ac76/raw/solo.log)
- [solo.ndjson](https://gist.githubusercontent.com/jeromy-cannon/6dc5ac0d4922495e86abe12a6126ac76/raw/solo.ndjson)

</details>

PR #5814, opened 2026-08-13
fix(ci): poll run status so the performance test stops when cancelled

### LINKED ISSUE #5813: repeatedly the Performance Test (aka Performance Test Solo Deployment) does not stop when cancelled
I've noticed that consistently when I go in and cancel workflows, the [Performance Test Solo Deployment](https://github.com/hiero-ledger/solo/actions/workflows/flow-performance-test.yaml) does not respond and keeps running.  It should automatically cancel using the concurrency rules for the workflow, and it should properly respond to cancel requests.

Fantomasa:
## Problem

The **Performance Test Solo Deployment** workflow does not stop when its run is
cancelled — whether the cancel comes from the `cancel-in-progress` concurrency
rule (a new push supersedes an older run) or a manual "Cancel workflow" click.
The run is marked `cancelled` at the GitHub level, but the job keeps running the
full ~38-minute performance test to completion.

This is reproducible across every recent cancelled run. For example run
`31689291513`: the run's `conclusion` is `cancelled`, yet the **job** started at
`10:03:43` and ran every step to `10:42:09` with `conclusion: success`. The step
log shows the cancellation poller started but its "signalling main shell" branch
never executed.

## Root cause

Self-hosted (ARC) runners do not terminate the running step on cancellation, so
the workflow relies on an in-step API poller to detect the cancel and stop the
task. That poller (added in #5442) counted **this job's own** `in_progress`
status:

```
[.jobs[] | select(.runner_name == $runner and .status == "in_progress")] | length
```

The job hosting the poller is itself always `in_progress` while its step runs, so
the count is permanently ≥ 1, never reaches 0, and the poller can never fire. It
is a self-referential check that is logically incapable of detecting the
cancellation of its own job.

(The prior approach, #5432, polled the run-level `updated_at` timestamp, which
false-fired because any sibling-job transition in a multi-job run — the nightly
extended tests — bumps it.)

## Fix

Poll the **run's** `status`/`conclusion` instead of the job's own in-progress
count:

```
if (.status == "completed" or .conclusion == "cancelled") then "stop" else "run" end
```

- These flip to `completed`/`cancelled` as soon as a cancel is requested, so the
  poller actually fires (SIGTERM → `cleanup` → the task's process group is
  killed).
- They are immune to the sibling-job churn that broke the `updated_at` approach:
  a run stays `in_progress` until **all** its jobs finish, so a sibling
  transition never changes them. Verified against a live in-progress run
  (`status: in_progress`, `conclusion: null` → `run`) and cancelled runs
  (→ `stop`).
- Works under `workflow_call` (nightly extended tests): `GITHUB_RUN_ID` is the
  parent run, whose `conclusion` becomes `cancelled` only when the whole nightly
  run is cancelled — the correct time to stop.

The two-consecutive-reading guard and 30 s cadence are preserved, so a single
stale/failed API read cannot trigger a false stop (detection latency ≤ ~90 s vs.
the current 38 min).

This single fix addresses both symptoms in the issue: because concurrency cancels
and manual cancels both manifest as "run marked cancelled, job keeps running", the
concurrency group itself is already correct — the job just never reacted to the
cancel it was being sent.

Closes #5813

### DIFF
--- a/.github/workflows/flow-performance-test.yaml
+++ b/.github/workflows/flow-performance-test.yaml
@@ -202,30 +202,33 @@ jobs:
           }
           trap cleanup SIGTERM SIGINT
 
-          # API poller: polls this job's own status every 30 s and sends SIGTERM
-          # to the main shell when GitHub no longer reports it in_progress.
-          # Self-hosted runners do not deliver SIGTERM to step scripts directly,
-          # so this is the only reliable way to detect workflow cancellation.
-          # The run-level updated_at field is NOT a cancellation signal: any
-          # sibling job transition in a multi-job run (e.g. the nightly extended
-          # tests) bumps it, which made this job SIGTERM its own healthy task
-          # (see hiero-ledger/solo#5432). The job is identified by RUNNER_NAME,
-          # which is unique per ephemeral ARC runner pod.
+          # API poller: polls this workflow RUN's status every 30 s and sends
+          # SIGTERM to the main shell when GitHub reports the run cancelled or
+          # completed. Self-hosted (ARC) runners do not terminate the running
+          # step on cancellation — the job otherwise runs every step to full
+          # completion even after the run is cancelled (hiero-ledger/solo#5813) —
+          # so this API poll is the only reliable way to react to cancellation.
+          #
+          # The signal is the RUN's status/conclusion, NOT the job's own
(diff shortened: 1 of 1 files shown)


ddeskov-limechain (approved):
LGTM

jeromy-cannon (approved):

PR #5815, opened 2026-08-13
fix: `cluster-ref config setup` fails on a cluster created by one-shot

### LINKED ISSUE #5374: cluster-ref config setup fails with a cluster created by one-shot
Steps to reproduce:

1. Start from scratch or cleanup previous setup:
  1.1. `docker ps` - no containers running
  1.2. `kubectl get ns` - can't connect to a cluster
  1.3. `~/.kube/config` is empty or no active context
2. Follow Quickstart instructions in the docs: https://solo.hiero.org/docs/simple-solo-setup/quickstart/ to set up a one-shot solo deployment
3. Then clean up the deployment as proposed in the Cleanup section: `solo deployment single destroy --deployment ${SOLO_DEPLOYMENT}`
4. The cluster is still alive, `kubectl get ns` works just fine, the docker container is running
5. Now follow to `Advanced Solo Setup` section in the docs to set up a custom deployment
6. My assumption is it will reuse the same Kubernetes cluster as was created by the one-shot commands, same docker container with same Kubernetes config in `~/.kube/config`
7. It fails at the following step: `solo cluster-ref config setup`
8. The error message is:

```shell
╭─ ERROR ────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ [SOLO-2016] Cluster setup failed: Container operation 'cluster-ref config setup' failed: MinIO Operator chart      │
│ installation failed: Failed to install Helm chart 'operator': Helm command failed with exit code 1. Command: 'helm │
│ install --create-namespace --output json --kube-context kind-solo-cluster --namespace solo-cluster-setup --version │
│ 7.1.1 --set operator.replicaCount=1 operator operator/operator'. Error: Error: INSTALLATION FAILED: Unable to      │
│ continue with install: CustomResourceDefinition "tenants.minio.min.io" in namespace "" exists and cannot be        │
│ imported into the current release: invalid ownership metadata; annotation validation error: key                    │
│ "meta.helm.sh/release-namespace" must equal "solo-cluster-setup": current value is "solo-setup"                    │
│   → Check solo logs: tail -n 100 ~/.solo/logs/solo.log                                                             │
│   → List installed Helm releases: helm list -A                                                                     │
│   → Inspect cluster pods: kubectl get pods -A                                                                      │
│   → Re-run cluster setup: solo cluster-ref config setup                                                            │
│                                                                                                                    │
│ Learn more: https://solo.hiero.org/docs/troubleshooting/errors/deployment/SOLO-2016/                               │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

I noticed `solo one-shot` ignores `SOLO_xyz` env vars. For example, I set `SOLO_CLUSTER_SETUP_NAMESPACE` to `solo-cluster-setup`, but `one-shot` still uses the `solo-setup` namespace. Maybe this leads to discrepancy in the CRD metadata.

Note that solo doesn't provide (or I couldn't find) a way to just create a kind cluster. `one-shot` creates one, but it ignores the env vars. `deployment` assumes there is already a cluster running, but it can be used only with all defaults like cluster setup namespace to be `solo-setup` and nothing different.

jan-milenkov:
## Description

Solo decided whether the MinIO Operator was already installed by looking for a running operator pod.
After a destroy the pod is gone but the operator's cluster-scoped CRDs remain, so `cluster-ref config
setup` concluded MinIO was absent, tried to install it again, and Helm refused to adopt the existing
CRDs — the SOLO-2016 error in the issue.

Setup now looks for those CRDs directly and reuses them when they are present, the same way it already
handles cert-manager's.

The leftovers themselves came from `cluster-ref config reset`, which uninstalled the operator from the
namespace passed on the flag rather than the one it was actually installed in. That reported success
while leaving the release and its CRDs in place. Reset now resolves the operator's real namespace
first, so nothing is left behind for a later setup to collide with.

### Related Issues

* Closes # https://github.com/hiero-ledger/solo/issues/5374


src/commands/cluster/tasks.ts
```diff
@@ -501,11 +514,17 @@ export class ClusterCommandTasks {
   public uninstallMinioOperator(): SoloListrTask<ClusterReferenceResetContext> {
     return {
       title: 'Uninstall MinIO Operator chart',
-      task: async ({config: {clusterSetupNamespace: namespace, context}}): Promise<void> => {
-        const {exists: isMinioInstalled, releaseName}: ReleaseNameData = await this.findMinioOperator(context);
+      task: async ({config: {context}}): Promise<void> => {
+        // Looked up across all namespaces: the operator is shared, so it need not live in the namespace
+        // this reset was invoked with.
+        const release: ReleaseItem | undefined = await this.chartManager.getInstalledRelease(
```
jeromy-cannon, line 555: the `namespaceName` parameter of `getInstalledRelease` is typed as `NamespaceName`, not `NamespaceName | undefined`. Passing `undefined` is fine at runtime (the implementation already handles it with `!namespaceName` and `namespaceName?.name`), but `strict: false` is the only reason the compiler lets this through — see the tsconfig TODO. `chart-manager.ts` should be updated:

```suggestion
        const release: ReleaseItem | undefined = await this.chartManager.getInstalledRelease(
          undefined,
```

…and in `chart-manager.ts`:
```typescript
public async getInstalledRelease(
  namespaceName: NamespaceName | undefined,
  chartReleaseName: string,
  kubeContext?: string,
): Promise<ReleaseItem | undefined>
```

src/commands/cluster/tasks.ts
```diff
@@ -61,19 +61,32 @@ export class ClusterCommandTasks {
     this.oneShotState = patchInject(oneShotState, InjectTokens.OneShotState, this.constructor.name);
   }
 
-  public findMinioOperator(context: Context): Promise<ReleaseNameData> {
-    return findMinioOperator(context, this.k8Factory);
+  /**
+   * The operator's CRDs are cluster-scoped, so they are what a second install collides with — and they
+   * remain after the operator's pods and namespace are gone.
+   */
+  public async findMinioOperatorCrds(context: Context): Promise<string[]> {
```
jeromy-cannon: consider making this `private` — it is only called from `installMinioOperatorChart` on the same class. The outer `beforeEach` stub (`ClusterCommandTasks.prototype, 'findMinioOperatorCrds' as any`) works for private methods too.

jan-milenkov: updated

src/commands/cluster/tasks.ts
jan-milenkov: updated

.github/workflows/zxc-e2e-test.yaml
```diff
@@ -221,12 +221,8 @@ defaults:
     shell: bash
 
 permissions:
-  id-token: write
   contents: read
-  actions: write
-  pull-requests: write
-  checks: write
-  statuses: write
+  actions: read
```
JeffreyDallas: nothing in this workflow reads the Actions API — the steps are harden-runner, checkout, setup-node/kubectl/helm/kind/task/go/java/gradle, the e2e run, and three `upload-artifact` calls. `upload-artifact` uses the runtime token, not `GITHUB_TOKEN`, so this can go to zero:

```suggestion
```

Dropping it here is what lets the caller grants in `flow-build-application.yaml` and `flow-pull-request-checks.yaml` shrink to `contents: read`. Change both sides in the same commit.

.github/workflows/flow-gcs-test.yaml
```diff
@@ -34,16 +34,19 @@ defaults:
     shell: bash
 
 permissions:
-  id-token: write
   contents: read
-  actions: write
+  actions: read # gha-jobid-action reads the jobs of the current workflow run
```
JeffreyDallas: this top-level grant is dead — `gcs-storage-test` is the only job and it declares its own `permissions` block on line 46, which replaces the workflow-level one wholesale. Worse, this file exposes `workflow_call`, so a top-level request forces every future caller to grant `actions: read` at the job level even though no job would use it.

```suggestion
```

.github/workflows/zxf-stale-issue-pr-closer.yaml
```diff
@@ -11,15 +11,15 @@ defaults:
...
-  contents: write
+  contents: read
 
 jobs:
   stale-pull-requests:
     name: Stale Pull Requests
     runs-on: hiero-solo-linux-medium
+    permissions:
+      contents: write # actions/stale deletes the head branch (delete-branch: true)
+      pull-requests: write # actions/stale labels, comments on and closes pull requests
```
JeffreyDallas: `actions/stale`'s README asks for `issues: write` in addition to `pull-requests: write`, and label mutation goes through the `POST /repos/{owner}/{repo}/issues/{number}/labels` endpoint. I think you're fine because `days-before-issue-stale: -1` and `days-before-issue-close: -1` mean issues are never processed and that endpoint accepts `pull-requests: write` when the target is a PR — but this workflow only runs on a `15 0 * * *` cron, so if I'm wrong nobody notices for weeks.

did you get a green `workflow_dispatch` run of this on the branch? If so, could you link it?

.github/workflows/zxc-unit-test.yaml
```diff
@@ -39,15 +39,16 @@ defaults:
...
-  id-token: write
   contents: read
   actions: read
-  pull-requests: write
-  checks: write
-  statuses: write
 
 jobs:
   unit-test:
+    permissions:
```
JeffreyDallas: moving the write scopes into a reusable workflow's job creates a contract that isn't visible from either side: any caller that forgets a job-level grant no longer just loses the comment at runtime, the whole run fails to start with `The workflow is requesting 'checks: write', but is only allowed 'checks: none'`. I checked and both current callers (`flow-build-application.yaml`, `flow-pull-request-checks.yaml`) are covered, but consider a one-line header comment above the block spelling out the requirement, e.g.:

```yaml
    # Callers must grant checks: write and pull-requests: write on the calling job.
```

Applies the same way to `zxc-code-analysis.yaml:82` and `zxc-hugo-publish.yaml:50` — one comment each, no need to repeat the reasoning.

src/commands/cluster/tasks.ts
```diff
@@ -61,19 +61,32 @@ export class ClusterCommandTasks {
...
+
+    return present;
   }
 
   public async installMinioOperatorChart(clusterSetupNamespace: NamespaceName, context: Context): Promise<void> {
-    const existingMinioOperator: ReleaseNameData = await this.findMinioOperator(context);
+    const existingCrds: string[] = await this.findMinioOperatorCrds(context);
 
-    if (existingMinioOperator.exists) {
+    if (existingCrds.length > 0) {
```
JeffreyDallas: keying "already installed" off CRD presence answers a different question than the one we need: the CRDs prove that *some* release once owned them, not that an operator is running now.

in the exact state this PR is aimed at — namespace deleted, so the Helm release secret went with it, CRDs orphaned because they are cluster-scoped — we take this branch, print "Reusing pre-existing MinIO Operator", and return. `network deploy` then creates a `Tenant` CR that nothing reconciles, so a loud SOLO-2016 at `cluster-ref config setup` turns into a silent hang much later and much further from the cause.

the precedent I would follow is the prometheus-operator CRD handling in `src/commands/network.ts:1400-1427`: it probes the CRDs, reports what it found, sets `<key>.enabled=false` for each one it already has, and **still installs the chart**. Same shape applies here — reuse the CRDs, install the operator.

if the MinIO chart can't be told to leave the CRDs alone, then require evidence of an owner before skipping (CRDs present **and** `getInstalledRelease(undefined, MINIO_OPERATOR_RELEASE_NAME, context)` returns a release, or the operator pod is up), and when they are orphaned fail with a registered `SoloErrors` subclass that names the CRDs to delete. That keeps the failure loud and actionable instead of moving it downstream.

src/commands/cluster/tasks.ts
```diff
@@ -61,19 +61,32 @@ export class ClusterCommandTasks {
     this.oneShotState = patchInject(oneShotState, InjectTokens.OneShotState, this.constructor.name);
   }
 
-  public findMinioOperator(context: Context): Promise<ReleaseNameData> {
-    return findMinioOperator(context, this.k8Factory);
+  /**
+   * The operator's CRDs are cluster-scoped, so they are what a second install collides with — and they
+   * remain after the operator's pods and namespace are gone.
+   */
+  private async findMinioOperatorCrds(context: Context): Promise<string[]> {
```
JeffreyDallas: this is the fourth copy of the "walk a CRD name list and ask the cluster which ones are present" loop:

- `src/commands/explorer.ts:321-336` (cert-manager)
- `src/commands/network.ts:1300-1305` (PodLogs)
- `src/commands/network.ts:1405-1417` (prometheus-operator)
- this one

Checklist §1 — three or more occurrences means extract. A static helper (e.g. `SharedClusterResourceReport.probeCrds(k8Factory, context, crdNames): Promise<Map<string, Record<string, string>>>`, or its own class with static methods — not an exported function, §10.3.1) called from all four drops ~50 lines and leaves one place to change the semantics.

while extracting: the other three use `crds().readLabels(crd)` rather than `ifExists(crd)` so they can feed `SharedClusterResourceReport.versionFromLabels(...)`. `ifExists` is literally `(await this.readLabels(crdName)) !== undefined`, so switching costs nothing and gets the version back into the report.

src/commands/cluster/tasks.ts
```diff
@@ -61,19 +61,32 @@ export class ClusterCommandTasks {
...
+    const existingCrds: string[] = await this.findMinioOperatorCrds(context);
 
-    if (existingMinioOperator.exists) {
+    if (existingCrds.length > 0) {
       SharedClusterResourceReport.show(
         this.logger,
         'MinIO Operator',
         context,
-        `Helm release '${existingMinioOperator.releaseName}' (${SharedClusterResourceReport.formatVersion(existingMinioOperator.version)})`,
+        `CRDs ${existingCrds.join(', ')} already present`,
```
JeffreyDallas: this drops the version out of the report, which was the point of `SharedClusterResourceReport` — the old line told the user it was reusing `operator` at version 7.1.1, the new one just names the CRDs. And because the `expected` slot below is unchanged, the message now reads:

> found CRDs tenants.minio.min.io, policybindings.sts.min.io already present, expected version 7.1.1 as release 'operator'

`versionFromLabels()` exists for exactly this and `crds().readLabels()` gives you the labels (see the comment above on `findMinioOperatorCrds`), so the found-string can stay comparable to the expected-string the way `explorer.ts:340-343` and `network.ts:1418-1426` do it.

src/commands/cluster/tasks.ts
```diff
@@ -501,11 +514,17 @@ export class ClusterCommandTasks {
...
+        // this reset was invoked with.
+        const release: ReleaseItem | undefined = await this.chartManager.getInstalledRelease(
+          undefined,
+          constants.MINIO_OPERATOR_RELEASE_NAME,
+          context,
+        );
 
-        if (isMinioInstalled) {
-          await this.chartManager.uninstall(namespace, releaseName, context);
+        if (release) {
```
JeffreyDallas: this widens the blast radius more than I think we want. `getInstalledRelease(undefined, 'operator', context)` is `helm list -A` filtered on the release name alone, and `MINIO_OPERATOR_RELEASE_NAME` is the literal string `operator` — which is also the release name in MinIO's own documented install (`helm install operator minio-operator/operator -n minio-operator`). So a user who runs their own MinIO Operator (or anything else installed as release `operator`) in an unrelated namespace now has it uninstalled by `solo cluster-ref config reset`.

before this change that was a no-op: `findMinioOperator` located the pod anywhere, but the uninstall targeted `clusterSetupNamespace`, so a foreign release in another namespace was never touched. This turns a no-op into a destructive one.

the `isRemoteConfigPresentInAnyNamespace` guard in `uninstallClusterChart` (line 570) doesn't cover it either — it is short-circuited by `!argv.force`, and `DestroyArgvBuilders.buildClusterResetArgv` (`src/commands/one-shot/orchestrator/destroy/destroy-argv-builders.ts:87-97`) pushes `--force` unconditionally. So every `one-shot` / `deployment single destroy` reaches this line with the shared-cluster guard off: two Solo deployments co-resident in one cluster, destroy one, the other loses its MinIO Operator.

at minimum gate on the chart rather than the release name:

```suggestion
        if (release?.chart?.startsWith(`${constants.MINIO_OPERATOR_CHART}-`)) {
```

better still, search only the namespaces Solo manages (the cluster-setup namespaces of the known cluster refs) instead of `-A`, so reset can only remove what Solo installed.

src/core/constants.ts
```diff
@@ -155,6 +155,8 @@ export const MINIO_OPERATOR_CHART_URL: string =
   getEnvironmentVariable('MINIO_OPERATOR_CHART_URL') ?? 'https://operator.min.io/';
 export const MINIO_OPERATOR_CHART: string = 'operator';
 export const MINIO_OPERATOR_RELEASE_NAME: string = 'operator';
+/** Cluster-scoped, so they outlive the namespace the operator was installed into. */
+export const MINIO_OPERATOR_CRDS: string[] = ['tenants.minio.min.io', 'policybindings.sts.min.io'];
```
JeffreyDallas, line 171: two things worth pinning down here:

1. this list has to track `MINIO_OPERATOR_VERSION`. `policybindings.sts.min.io` only appears from operator v5 onward, so against an older pre-existing operator only `tenants.minio.min.io` is found — and the `existingCrds.length > 0` test in `tasks.ts:84` still calls that "reuse". A partial or older install is indistinguishable from a current complete one.
2. any-one-present is the odd semantics out. `CERT_MANAGER_CRDS` (line 612) is consumed all-must-be-present (`explorer.ts:324-336` breaks on the first missing CRD), and the prometheus list in `network.ts:1405-1426` counts N of N. If the intent here is "reuse only a complete install", make it `existingCrds.length === constants.MINIO_OPERATOR_CRDS.length` and say so in the doc comment; if the intent really is "any", the comment should explain why MinIO differs from the other two.

src/types/aliases.ts
```diff
@@ -9,18 +8,26 @@ export type NodeId = number;
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type CommandBuilder = (yargs: any) => any;
 
 export type TarCreateFilter = (path: string, entry: Stats | ReadEntry) => boolean;
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type SkipCheck = (context_: any) => Promise<boolean> | boolean;
 
-export type TaskFunction = (
+export type ConfigBuilder = (
```
JeffreyDallas: what does the `ConfigBuilder` / `TaskFunction` rework have to do with the MinIO fix? It reads like it rode in on the `lint-fixes` commit. Two notes if it stays:

- `@typescript-eslint/no-explicit-any` is a `warn` in `eslint.config.mjs:251` with a standing `// TODO remove (406 errors)`. Seven new `eslint-disable-next-line` directives don't retire any of those 406 — they hide them from the count that TODO is tracking. Prefer typing the params, or leaving the warning visible.
- the `TaskFunction` you deleted typed `task` as `ListrTaskWrapper<any, any, any>`. If we're retyping `ConfigBuilder` anyway, `task: SoloListrTaskWrapper<AnyListrContext>` is strictly better than `task: any` and needs no disable directive at all.

I'd rather see this land as its own cleanup PR so the MinIO fix stays reviewable and revertable on its own.

src/core/certificate-manager.ts
```diff
@@ -45,9 +45,9 @@ export class CertificateManager {
     switch (type) {
       //? HAProxy
       case GrpcProxyTlsEnums.GRPC: {
-        const certData = fs.readFileSync(cert).toString();
-        const keyData = fs.readFileSync(key).toString();
-        const pem = `${certData}\n${keyData}`;
+        const certData: string = fs.readFileSync(cert).toString();
```
JeffreyDallas: same question as `src/types/aliases.ts` — this file has nothing to do with the MinIO fix. The annotations are right (and `catch (error)` in place of `catch (error: Error | any)` is an improvement), but it makes the diff harder to read as a bug fix and harder to revert if we need to.

if it stays: `buildSecret` on line 44 is the one declaration left in this file without a return type, so the §6.1 pass is incomplete.

test/unit/commands/cluster.test.ts
```diff
@@ -221,4 +218,76 @@ describe('ClusterCommand unit tests', (): void => {
...
+
+      chartManager = container.resolve(InjectTokens.ChartManager);
+      chartManager.install = sandbox.stub().resolves(true);
+      chartManager.uninstall = sandbox.stub().resolves(true);
+      chartManager.getInstalledRelease = sandbox.stub().resolves();
+
+      tasks = container.resolve(InjectTokens.ClusterCommandTasks);
+    });
+
+    it('reuses a pre-existing operator instead of installing over its CRDs', async (): Promise<void> => {
```
JeffreyDallas: the name promises the live-operator path, but the only signal stubbed is `crds().ifExists` — and CRD presence is the *only* thing `installMinioOperatorChart` looks at. `getInstalledRelease` is stubbed to `undefined` on line 243, so what this test actually pins is the orphan-CRD skip I flagged on `tasks.ts:84`, not "a pre-existing operator". Nothing here would fail in the broken case.

what I'd want to see instead:

- CRDs present **and** `getInstalledRelease` resolves a release → skip the install (the genuine reuse case).
- CRDs present and `getInstalledRelease` resolves `undefined` → whatever we decide the orphan behaviour is (install with the CRDs reused, or throw a `SoloErrors` subclass), asserted explicitly.

test/unit/commands/cluster.test.ts
```diff
@@ -221,4 +218,76 @@ describe('ClusterCommand unit tests', (): void => {
       expect(rbacStub.deleteClusterRole.notCalled).to.be.true;
     });
   });
+
+  describe('MinIO Operator, whose CRDs outlive the namespace it was installed into', (): void => {
+    const configuredNamespace: NamespaceName = NamespaceName.of('solo-cluster-setup');
+    const installedNamespace: string = 'solo-setup';
+    let tasks: ClusterCommandTasks;
+    let chartManager: any;
```
JeffreyDallas: §6.1 wants real types here — `SinonStubbedInstance<ChartManager>` for `chartManager` and `SinonStubbedInstance<Crds>` (or a small local type) for `crdsStub` — rather than `any`.

also, `chartManager.install = sandbox.stub()` on lines 241-243 assigns over the container singleton's methods, and `sandbox.restore()` does not undo a plain property assignment, so the stubs outlive this `describe`. `sandbox.stub(chartManager, 'install').resolves(true)` is restored properly. The older `describe` above does the same thing so this is pre-existing, but this block is the one adding new ones.

src/commands/cluster/tasks.ts
jan-milenkov: updated

src/commands/cluster/tasks.ts
jan-milenkov: updated

src/commands/cluster/tasks.ts
jan-milenkov: updated

src/commands/cluster/tasks.ts
jan-milenkov: updated

src/core/constants.ts
jan-milenkov: updated

test/unit/commands/cluster.test.ts
jan-milenkov: updated

test/unit/commands/cluster.test.ts
jan-milenkov: updated

JeffreyDallas (approved):

PR #5835, opened 2026-08-14
chore: remove deprecated solo init command

### LINKED ISSUE #5389: Remove deprecated command init

JeffreyDallas:
## Description

* Removes the deprecated `solo init` command (deprecated since v0.85.0, removal tracked in #5389). All functionality `solo init` provided is already handled automatically:
  - **Home directory / local config / template setup** → `Middlewares.initSystemFiles()` runs as a yargs middleware before every command
  - **Dependency checks / chart manager setup** → `Subcommand.installDependencies()` runs before every `CommandBuilder`-based subcommand
* Inlines `setupSystemFilesTasks` logic directly into `Middlewares.initSystemFiles()`
* Inlines `installDependenciesTasks` logic directly into `Subcommand.installDependencies()`
* Deletes `InitCommand`, `InitConfig`, `InitContext`, and `InitFailedSoloError`
* Removes `InjectTokens.InitCommand` DI token and container registration
* Removes the redundant `invokeSoloCommand('init', ...)` call in `backup-restore.ts`
* Removes `solo init` step from all 15 example Taskfiles and `scripts/Taskfile.scripts.yml`
* Removes all init-related unit/integration/e2e tests

**CI fixes required after removing `solo init` from workflows (all workflows previously relied on `solo init` to install helm and kind binaries):**

* Adds explicit `Install Helm` and `Install Kind` steps to `flow-examples-test.yaml`, pinned to the same versions used across all E2E tests (helm v3.14.2, kind v0.29.0, k8s v1.31.9)
* Adds explicit `Install Helm` and `Install Kind` steps to `flow-gcs-test.yaml` (the GCS test script calls `kind delete cluster` directly)
* Enables `helm-install: true` and `kind-install: true` for all 12 entries in `e2e-test-matrix.json` that previously had these set to `false`, relying on `solo init` to supply the binaries
* Adds `timeout-minutes: 20` to the `Install Validation (Podman Runtime)` job in `flow-install-validation-runtime.yaml` to prevent silent brew hangs from blocking CI indefinitely (root cause addressed separately in #5837)

### Related Issues

* Closes #5389
* Related to #5836, #5837

### Testing

* N/A — all changes verified through CI runs on this PR branch

The following was not tested:

* The end-to-end behavior of `solo init` removal is implicitly validated by every E2E test that runs without calling `solo init` first; individual removal of the command is covered by updated unit tests

<details>
<summary>
Commit message guidelines
</summary>

</details>


test/e2e/dual-cluster/README.md
```diff
@@ -143,7 +143,7 @@ Current Command		: init
 
 ***************************************************************************************
 Note: solo stores various artifacts (config, logs, keys etc.) in its home directory: /Users/user/.solo
-If a full reset is needed, delete the directory or relevant sub-directories before running 'solo init'.
+If a full reset is needed, delete the directory or relevant sub-directories before running solo again.
```
JeffreyDallas, line 146: this transcript still shows `dist/solo.js init` at lines 127–141, so changing only the final sentence leaves the documented workflow invoking a command this PR removes. Please remove or regenerate that obsolete block, and update the remaining runtime recovery messages that prescribe `solo init` (`dependency-*-solo-error.ts`, `homebrew-install-failed-solo-error.ts`, `init-system-files-failed-solo-error.ts`, `cluster-reference-undetermined-solo-error.ts`, and `unsupported-linux-distribution-solo-error.ts`). Users otherwise receive instructions guaranteed to fail. Style guide §10.4 treats default behavior and its user-facing contract as backwards compatibility.

src/core/middlewares.ts
```diff
@@ -62,9 +62,59 @@ export class Middlewares {
 
   public initSystemFiles(): (argv: ArgvStruct) => AnyObject {
```
JeffreyDallas, line 66: we deleted the focused init unit/integration tests, but there is no replacement coverage for this new execution path. Please add unit tests that invoke this middleware with a temporary home/cache and assert directory creation, local-config creation/loading, template copying, and owner-only permissions; add coverage for the equivalent dependency/chart/cluster sequencing in `Subcommand.installDependencies()`. This is exactly the cheap unit coverage expected before relying on E2E runs (style guide §10.6).

jeromy-cannon, line 66: No unit tests were added for `initSystemFiles()` even though it now runs on every command invocation (previously only under `solo init`). The deleted `test/unit/commands/init.test.ts` covered the dev-debug-list behavior that's now gone with no replacement test. Given the higher blast radius (every command vs. an explicit opt-in command), I'd want at least a unit test around directory setup + the copy-templates skip/error path before merging.

src/core/middlewares.ts
```diff
@@ -62,9 +62,59 @@ export class Middlewares {
...
-        // @ts-expect-error - TS2445: Property taskList is protected and only accessible within class BaseCommand and its subclasses.
-        this.initCommand.taskList.newTaskList(this.initCommand.setupSystemFilesTasks(argv), {renderer: 'silent'});
+      const cacheDirectory: string =
+        (this.configManager.getFlag<string>(flags.cacheDir) as string) || (constants.SOLO_CACHE_DIR as string);
+
+      const tasks: Listr<ListrContext, ListrRendererValue, ListrRendererValue> = this.taskList.newTaskList(
+        [
+          {
+            title: 'Setup home directory and cache',
+            task: (): void => {
```
jeromy-cannon, line 82: this duplicates `BaseCommand.setupHomeDirectory()` (same four directories: `SOLO_HOME_DIR`, `SOLO_LOGS_DIR`, cache dir, `SOLO_VALUES_DIR`) but drops two things that method had: the `try/catch` wrapping into `SoloErrors.system.directoryCreationFailed`, and the per-directory `this.logger.debug(...)` call. A raw `fs.mkdirSync` failure here (e.g. permission denied on `~/.solo`) now propagates as an unwrapped Node error instead of a registered `SoloError` with troubleshooting steps.

Since `Middlewares` isn't a `BaseCommand`, perhaps `setupHomeDirectory` should move to a `static` helper (or a shared utility) both classes can call, rather than re-implementing it here — same pattern as `checkDockerDesktopContainerdSetting` was made `public static` for reuse.

src/core/middlewares.ts
```diff
@@ -62,9 +62,59 @@ export class Middlewares {
...
+          {
+            title: 'Create local configuration',
+            skip: (): boolean => this.localConfig.configFileExists(),
+            task: async (): Promise<void> => {
+              await this.localConfig.load();
+            },
+          },
+          {
+            title: `Copy templates in '${cacheDirectory}'`,
+            task: (): void => {
```
jeromy-cannon, line 104: the old `InitCommand` version of this task tracked `directoryCreated` and, on first creation, printed the "Home Directories"/"Chart Repository" debug lists (`argv.debug`) plus the grey "Note: solo stores various artifacts..." message. None of that survived the inlining. `test/e2e/dual-cluster/README.md:143-147` still shows that block as expected output (only the last line's wording changed) — was removing this message intentional, or should it be preserved here?

src/core/command-path-builders/command-builder.ts
```diff
@@ -35,26 +38,68 @@ export class Subcommand {
...
     useSmallMemoryCluster: boolean = false,
     collapseTasks: boolean = false,
   ): Promise<void> {
+    if (!this.dependencies || this.dependencies.length === 0) {
+      return;
+    }
+
+    const taskItems: any[] = [
+      {
+        title: 'Pre-flight: check Docker Desktop containerd setting',
```
jeromy-cannon: this seems like a duplicate of `BaseCommand.dockerDesktopPreflightTask()` — same title (`'Pre-flight: check Docker Desktop containerd setting'`), same body (call `checkDockerDesktopContainerdSetting()`, warn if `containerdSnapshotterEnabled`). Since `Subcommand` isn't a `BaseCommand`, consider making `dockerDesktopPreflightTask` `static` (like `checkDockerDesktopContainerdSetting` already was in this PR) and calling `BaseCommand.dockerDesktopPreflightTask()` here instead of re-declaring the task inline — avoids the title string drifting out of sync between the two copies.

src/core/command-path-builders/command-builder.ts
```diff
@@ -35,26 +38,68 @@ export class Subcommand {
...
 
   public async installDependencies(
     useSmallMemoryCluster: boolean = false,
     collapseTasks: boolean = false,
   ): Promise<void> {
+    if (!this.dependencies || this.dependencies.length === 0) {
+      return;
+    }
+
+    const taskItems: any[] = [
```
jeromy-cannon: `taskItems: any[]`, and `task: (_: ListrContext, task: any): any` / `tasks: any` below, lose the typing the old `SoloListrTask<InitContext>[]` had. The sibling refactor in this same PR (`cluster-task-manager.ts`) correctly swapped `InitContext` for the existing `AnyObject` type instead of falling back to `any` — worth doing the same here for consistency (`SoloListrTask<AnyObject>[]`, etc.).

src/commands/init/init.ts
```diff
@@ -1,232 +0,0 @@
...
-@injectable()
-export class InitCommand extends BaseCommand {
-  public static readonly COMMAND_NAME: string = 'init';
-  public static readonly INIT_COMMAND_NAME: string = InitCommand.COMMAND_NAME;
-  private static hasShownDevSystemFileLists: boolean = false;
-  private static readonly DEPRECATION: Deprecation = {
-    since: '0.85.0',
-    removalIssue: 5389,
-    reason: 'Running it is no longer required.',
-  };
```
jeromy-cannon, line 35: It says we deprecated this recently in v0.85.0.  How, can we remove the logic but still leave the deprecation alert with a return code zero success (since it doesn't do anything) until the last supporting LTS is no longer supported?

JeffreyDallas (commented):
## Intent

Remove the deprecated `solo init` command and move its system-file and dependency setup into normal command execution, while updating callers and CI prerequisites.

## Verdict

Request changes

## Critical

- None.

## Major

- Removed-command guidance remains in the user-facing E2E transcript and in seven runtime error recovery paths. These now direct users to a command that always fails; update the transcript and replace each recovery instruction with the relevant automatic setup or rerun guidance. This is a backwards-compatibility/user-experience regression (style guide §10.4).
- This moves the setup contract into `Middlewares.initSystemFiles()` and `Subcommand.installDependencies()`, but deletes all focused init unit/integration tests without adding coverage for the new paths. Add cheap unit tests for directory/config/template setup and dependency/chart/cluster task sequencing (style guide §10.6).

## Minor

- None.

## Questions for the author

- None.

## Positive

- The workflow changes explicitly install Helm and Kind where those binaries had previously been an implicit side effect of `solo init`.

## Companion work

- solo-docs PR: missing / N/A
- solo-containers PR: N/A

test/e2e/dual-cluster/README.md
JeffreyDallas: Fixed. Removed the obsolete `dist/solo.js init` transcript block (and the trailing artifact-note block) from the README, since `setup-dual-e2e.sh` no longer invokes `init`. Also updated the runtime recovery messages that still told users to run `solo init` in `dependency-not-found`, `dependency-install-failed`, `dependency-version-check-failed`, `homebrew-install-failed`, `init-system-files-failed`, `unsupported-linux-distribution`, `cluster-reference-undetermined`, and `local-config-not-found` — they now point at re-running the failing command (dependencies/system files are now installed automatically) or, for the cluster-reference case, at `solo cluster-ref config connect`.

src/core/middlewares.ts
JeffreyDallas: Added unit coverage: `test/unit/core/middlewares.test.ts` now exercises `Middlewares.initSystemFiles()` end-to-end with a temporary home/cache directory — asserting directory setup is delegated to `BaseCommand.setupHomeDirectory` (itself unit tested separately with real temp dirs and the error-wrapping path), local-config create-vs-skip behavior, template copying with owner-only permissions, the missing-source skip path, and the `InitSystemFilesFailedSoloError` wrapping. `test/unit/core/command-path-builders/command-builder.test.ts` (new) covers the equivalent `Subcommand.installDependencies()` sequencing: dependency checks, chart manager setup, local cluster setup, the shared Docker Desktop preflight task, and the `DependencyInstallFailedSoloError` wrapping.

JeffreyDallas: see the new `Middlewares.initSystemFiles` and `Subcommand.installDependencies` unit test suites.

src/core/middlewares.ts
JeffreyDallas: Fixed. Extracted `BaseCommand.setupHomeDirectory(directories, logger)` as a `public static` helper that keeps the try/catch wrapping into `SoloErrors.system.directoryCreationFailed` and the per-directory `logger.debug(...)` call. `Middlewares.initSystemFiles()` now delegates to it instead of re-implementing the bare loop, and the existing instance method `BaseCommand.prototype.setupHomeDirectory` delegates to the same static helper.

src/core/middlewares.ts
JeffreyDallas: intentional, will go with full removal rather than preserving the debug-list/artifact-note output. The dual-cluster README transcript has been updated to drop the corresponding obsolete output block.

src/core/command-path-builders/command-builder.ts
JeffreyDallas: Fixed. `BaseCommand.dockerDesktopPreflightTask()` is now `public static` (takes a `logger` param), and `Subcommand.installDependencies()` calls `BaseCommand.dockerDesktopPreflightTask(this.logger)` directly instead of re-declaring the task inline. The existing protected instance method now delegates to the same static helper.

src/core/command-path-builders/command-builder.ts
JeffreyDallas: Fixed. `taskItems` is now typed `SoloListrTask<AnyObject>[]`, the 'Check dependencies' task uses `AnyObject`/`SoloListrTaskWrapper<AnyObject>` (matching the `Listr<AnyObject, any, any>` return of `task.newListr`), and `tasks` is typed `Listr<AnyObject, ListrRendererValue, ListrRendererValue>` — consistent with the `AnyObject` pattern already used in `cluster-task-manager.ts`.

src/commands/init/init.ts
JeffreyDallas: Removal this does not impact last supporting LTS (already released)

jeromy-cannon: we need to honor the deprecation timeline of 3 months.  If we deprecated it in v0.85.0, that means about 3 months from now (minus a week or two).  I tested, and if a user is using scripts in their CI, then this is a breaking change without a proper deprecation window:
```
 @hiero-ledger/solo@0.86.0 solo
> node --no-deprecation --no-warnings dist/solo.js init

Unknown argument: init
Usage:
solo <command> [options]

Commands:
  config       Backup and restore component configurations for Solo deployments. These commands display what would be backed up or restored without performing actual operations.
  block        Block Node operations for creating, modifying, and destroying resources. These commands require the presence of an existing deployment.
  cluster-ref  Manages the relationship between Kubernetes context names and Solo cluster references which are an alias for a kubernetes context.
  consensus    Consensus Node operations for creating, modifying, and destroying resources. These commands require the presence of an existing deployment.
  deployment   Create, modify, and delete deployment configurations. Deployments are required for most of the other commands.
  explorer     Explorer Node operations for creating, modifying, and destroying resources.These commands require the presence of an existing deployment.
  keys         Consensus key generation operations
  ledger       System, Account, and Crypto ledger-based management operations. These commands require an operational set of consensus nodes and may require an operational mirror node.
  mirror       Mirror Node operations for creating, modifying, and destroying resources. These commands require the presence of an existing deployment.
  relay        RPC Relay Node operations for creating, modifying, and destroying resources. These commands require the presence of an existing deployment.
  cache        Manage solo cached items.
  one-shot     One Shot commands for new and returning users who need a preset environment type. These commands use reasonable defaults to provide a single command out of box experience.
  rapid-fire   Commands for performing load tests a Solo deployment

Options:


     --debug               ,  --dev         Enable debug mode                 [boolean] [default: false]
     --force-port-forward  Force port forward to access the network services  [boolean] [default: true]
-v,  --version             Show version number                                [boolean]
```

JeffreyDallas: resumed, now solo init command will print depreciation message

jan-milenkov (approved):

PR #5839, opened 2026-08-17
feat: Update Solo for Network transplant process (node-overrides.yaml)

### LINKED ISSUE #860: Update Solo for Network transplant process (node-overrides.yaml)

jan-milenkov:
## Description

A saved state also stores the list of machines in the network it came from — their addresses and their
identity certificates. Start a *different* network from that state and it tries to be the old network:
it talks to addresses that no longer exist, using certificates that do not match its own keys, and
fails to start.

Solo now writes an `override-network.json` file into `data/config`, which is where the consensus node
looks for it. That file lists the machines of the network you are actually starting, so the node uses
those instead of the ones stored in the state.

Two things worth knowing:

* The list is built fresh from the current deployment, so it stays correct after nodes are added or
  updated.
* It is only written when you pass `--transplant`. Restoring a network's own state leaves the list
  alone, because replacing a list that was already correct stops the node from starting.

Verified on a live network: the node logs that it read the file, and the file must be owned by the
`hedera` user. A root-owned copy is silently ignored and the node quietly keeps the old list.

Also adds an example, `examples/state-transplant`, that runs the whole thing end to end and checks the
node really used the file. It runs in CI.

### Related Issues

* Closes #860


src/commands/node/tasks.ts
```diff
@@ -2072,14 +2110,20 @@ export class NodeCommandTasks {
    * @param gossipEndpointPortMapping - port overrides for the gossip endpoints
    * @param serviceEndpointPortMapping - port overrides for the gRPC service endpoints
    */
-  private async generateGenesisNetworkJson(
+  /**
```
JeffreyDallas: This adds a second JSDoc block directly under the pre-existing one (`src/commands/node/tasks.ts:2103-2112`) instead of replacing it. Only this second block is attached to `generateNetworkJson` — the first one is now a dangling comment, and it's also stale: it still says "Generate genesis network json file", documents a `@param keysDirectory` that isn't a parameter, and doesn't mention the new `fileName` parameter at all. Please delete the old block (or fold anything still true into the new one) rather than leaving both.

src/commands/node/handlers.ts
```diff
@@ -1182,6 +1182,11 @@ export class NodeCommandHandlers extends CommandHandler {
         this.validateAllNodePhases({acceptedPhases: [DeploymentPhase.CONFIGURED]}),
         this.tasks.identifyExistingNodes(),
         this.tasks.uploadStateFiles(({config}): boolean => config.stateFile.length === 0),
+        // Only a transplant wants its roster replaced. Restoring a network's own state must leave the roster
+        // in the state alone, or the platform hits a roster transition it cannot replay past.
+        this.tasks.installOverrideNetworkJson(
```
JeffreyDallas: What happens if a user passes `--transplant` but forgets `--state-file` (or the reverse)? Right now the skip predicate just silently no-ops `installOverrideNetworkJson` — the node starts as if `--transplant` was never set, with no warning that the flag had no effect. Given the failure mode this is guarding against (a node that gossips to the wrong endpoints with mismatched certs) is confusing to debug, should this combination raise an error instead of silently doing nothing?

src/commands/flags.ts
```diff
@@ -456,6 +456,20 @@ export class Flags {
     prompt: undefined,
   };
 
+  public static readonly transplant: CommandFlag = {
+    constName: 'transplant',
+    name: 'transplant',
+    definition: {
+      describe:
+        'Treat the supplied state file as captured on a different network.' +
+        '\nInstalls this network‘s address book as override-network.json so the consensus node adopts it' +
```
JeffreyDallas:
```suggestion
        `\nInstalls this network's address book as override-network.json so the consensus node adopts it` +
```
This (and line 466) use `‘` (U+2018, a left single quotation mark) as an apostrophe rather than a straight `'` — looks like a copy/paste artifact. A template literal (as other possessive descriptions in this file do, e.g. `owner's username` at line 2111) avoids needing to escape the apostrophe; line 466 needs the same fix (`a network's own state`).

.github/workflows/support/examples-reduced-matrix.json
```diff
@@ -24,6 +24,11 @@
     "directory": "state-save-and-restore",
     "local-java-build": false
   },
+  {
```
JeffreyDallas: Per the README this example takes roughly 10-15 minutes (two Kind clusters, two single-node networks, freeze + state download + restore). The reduced matrix runs on every PR and push, not just nightly/scheduled — was adding this much runtime to every contributor's PR CI intentional, or would `examples-test-matrix.json` (the nightly/full matrix) be the better home for it?

src/commands/node/tasks.ts
```diff
@@ -1684,6 +1684,46 @@ export class NodeCommandTasks {
...
+      title: 'Install override-network.json',
+      task: async ({config}): Promise<void> => {
+        // Written to the cache rather than the staging directory: `node start` does not take the flags that
+        // populate `config.stagingDir`, so it is undefined here. The file only has to exist long enough to
+        // be copied into each pod.
+        const overrideNetworkJson: string = await this.generateNetworkJson(
+          constants.OVERRIDE_NETWORK_FILE,
+          config.namespace,
+          config.consensusNodes,
+          constants.SOLO_CACHE_DIR,
```
JeffreyDallas: this always writes `${SOLO_CACHE_DIR}/override-network.json`. Two `--transplant` starts under the same `SOLO_HOME` can interleave `writeFileSync()` and `copyTo()`, so one deployment can receive the other deployment's roster. Use a unique per-invocation cache path (for example, `fs.mkdtempSync`, as `createGeneratedGossipLoadBalancerService` does) and remove it after the copies complete. This is a correctness/backwards-compatibility issue; see style guide §§10.1 and 10.4.

src/commands/node/tasks.ts
JeffreyDallas, line 2209: This adds a second JSDoc block directly under the pre-existing one (`src/commands/node/tasks.ts:2103-2112`) instead of replacing it. Only this second block is attached to `generateNetworkJson` — the first one is now a dangling comment, and it's also stale: it still says "Generate genesis network json file", documents a `@param keysDirectory` that isn't a parameter, and doesn't mention the new `fileName` parameter at all. Please delete the old block (or fold anything still true into the new one) rather than leaving both.

src/commands/node/handlers.ts
JeffreyDallas, line 1188: What happens if a user passes `--transplant` but forgets `--state-file` (or the reverse)? Right now the skip predicate just silently no-ops `installOverrideNetworkJson` — the node starts as if `--transplant` was never set, with no warning that the flag had no effect. Given the failure mode this is guarding against (a node that gossips to the wrong endpoints with mismatched certs) is confusing to debug, should this combination raise an error instead of silently doing nothing?

src/commands/flags.ts
JeffreyDallas:
```suggestion
        `\nInstalls this network's address book as override-network.json so the consensus node adopts it` +
```
This (and line 466) use `‘` (U+2018, a left single quotation mark) as an apostrophe rather than a straight `'` — looks like a copy/paste artifact. A template literal (as other possessive descriptions in this file do, e.g. `owner's username` at line 2111) avoids needing to escape the apostrophe; line 466 needs the same fix (`a network's own state`).

.github/workflows/support/examples-reduced-matrix.json
JeffreyDallas: Per the README this example takes roughly 10-15 minutes (two Kind clusters, two single-node networks, freeze + state download + restore). The reduced matrix runs on every PR and push, not just nightly/scheduled — was adding this much runtime to every contributor's PR CI intentional, or would `examples-test-matrix.json` (the nightly/full matrix) be the better home for it?

examples/state-transplant/Taskfile.yml
```diff
@@ -0,0 +1,150 @@
+version: 3
+
+env:
+  # Runs from source by default so the example exercises the working tree. `npm run solo` would run the
+  # compiled dist/, which is only as current as the last build — a stale dist silently tests old behaviour.
+  SOLO_COMMAND:
+    sh: |
```
JeffreyDallas: this user-facing Taskfile immediately executes POSIX `sh`; later tasks also rely on `rm`, `grep`, `head`, `find`, and a bash verifier. It will not work on Windows without a POSIX compatibility layer. Please provide a Windows-native/task-native fallback or explicitly gate and document the supported platform. Style guide §10.5 and review checklist §11 require cross-platform safety.

src/commands/node/tasks.ts
```diff
@@ -1759,6 +1759,54 @@ export class NodeCommandTasks {
...
+        // keeps two concurrent transplants under the same SOLO_HOME from reading each other's roster.
+        fs.mkdirSync(constants.SOLO_CACHE_DIR, {recursive: true});
+        const temporaryDirectory: string = fs.mkdtempSync(PathEx.join(constants.SOLO_CACHE_DIR, 'override-network-'));
+
+        try {
+          const overrideNetworkJson: string = await this.generateNetworkJson(
+            constants.OVERRIDE_NETWORK_FILE,
+            config.namespace,
+            config.consensusNodes,
+            temporaryDirectory,
```
JeffreyDallas, line 1782: this call omits `domainNamesMapping`, `gossipEndpointPortMapping` and `serviceEndpointPortMapping`, but the genesis call at `tasks.ts:1941` passes all three. That divergence has teeth:

- `SETUP_FLAGS` includes `flags.domainNames`, `flags.gossipEndpointPort`, `flags.serviceEndpointPort`, and `setupConfigBuilder` (`configs.ts:703-708`) turns them into the mappings that `setupNetworkNodes(..., isGenesis: true)` feeds into `genesis-network.json`.
- `START_FLAGS` includes none of them, so `node start --transplant` cannot supply them even if the user wanted to.
- In `genesis-network-data-constructor.ts:81-99` the gossip port falls back to `HEDERA_NODE_EXTERNAL_GOSSIP_PORT`, the service port to `GRPC_PORT`, and the service address to `networkNodeService.externalAddress` when `domainNamesMapping` is absent.

So for a target network deployed with `consensus node setup --gossip-endpoint-port ...`, this override *replaces a correct roster with one on the default gossip port* — the same class of failure the feature exists to prevent, just arriving from the other direction. The default path is unaffected, which is why the example doesn't catch it.

Either pull the mappings off the deployment so the override matches what setup wrote, or add the three flags to `START_FLAGS`. If neither is in scope right now, say so in the JSDoc on `installOverrideNetworkJson` — right now the doc reads as though the generated roster is unconditionally correct for the target.

test/unit/commands/node/install-override-network-json.test.ts
```diff
@@ -0,0 +1,155 @@
...
+  it('carries the skip decision through, so a plain start writes nothing', (): void => {
+    const skip: () => boolean = (): boolean => true;
+    expect(tasks.installOverrideNetworkJson(skip).skip).to.equal(skip);
+  });
+
+  // The predicate `consensus node start` supplies. Getting this wrong installed an override on every state
+  // restore, which broke the state-save-and-restore example: replacing the roster in a network's own state
+  // forces a roster transition the consensus node cannot replay past.
+  describe('the skip predicate used by consensus node start', (): void => {
+    const skipPredicate = ({config}: {config: {transplant?: boolean; stateFile: string}}): boolean =>
```
JeffreyDallas, line 138: this is a re-declaration of the predicate, not the one `consensus node start` actually uses, so the three tests below cannot fail for the reason the comment above describes.

I checked: I replaced the predicate at `handlers.ts:1189` with `({config}): boolean => false` — which installs an override on every `consensus node start`, precisely the regression this block says it guards against — and all 8 tests still passed.

Extract it so there is one definition and the test imports the real one. A `public static` on `NodeCommandHandlers` works (needs a `type NodeStartConfigClass` import added to `handlers.ts`):

```ts
public static skipOverrideNetworkJson({config}: {config: NodeStartConfigClass}): boolean {
  return !config.transplant;
}
```

then `this.tasks.installOverrideNetworkJson(NodeCommandHandlers.skipOverrideNetworkJson)` at the call site, and the test imports it instead of redefining it.

Two smaller things fall out of the same line: `skipPredicate` has no type annotation (§6.1 — eslint flags it as `@typescript-eslint/typedef`), and the `transplant: true, stateFile: ''` case at line 151 asserts a config that `startConfigBuilder` now rejects before the task list is ever built.

src/commands/node/handlers.ts
```diff
@@ -1183,6 +1183,11 @@ export class NodeCommandHandlers extends CommandHandler {
...
+          ({config}): boolean => !config.transplant || config.stateFile.length === 0,
```
JeffreyDallas, line 1189:
```suggestion
          ({config}): boolean => !config.transplant,
```
the second clause is unreachable now. If `transplant` is falsy, `!config.transplant` short-circuits; if it's true, `startConfigBuilder` has already thrown `transplantRequiresStateFile` for an empty `stateFile`. CLAUDE.md "Dead Code Removal".

(If you take the extraction suggested on the test file, this collapses into that static instead.)

src/core/consensus-node-path-templates.ts
```diff
@@ -41,6 +41,14 @@ export class ConsensusNodePathTemplates {
 
   public static readonly GENESIS_NETWORK_JSON: string = `${this.DATA_CONFIG}/genesis-network.json`;
 
+  /** Replaces the address book carried by a state restored from another network. */
+  public static readonly OVERRIDE_NETWORK_JSON: string = `${this.DATA_CONFIG}/override-network.json`;
```
JeffreyDallas, line 45:
```suggestion
  public static readonly OVERRIDE_NETWORK_JSON: string = `${this.DATA_CONFIG}/${constants.OVERRIDE_NETWORK_FILE}`;
```
the PR adds `OVERRIDE_NETWORK_FILE` / `GENESIS_NETWORK_FILE` to `constants.ts` but leaves the filenames hardcoded here, so each name now lives in two places. This file already imports `constants` (it uses `constants.APPLICATION_PROPERTIES`), and `GENESIS_NETWORK_JSON` on line 42 and `ARCHIVE_GENESIS_NETWORK_JSON` on line 50 want the same treatment. I type-checked all three together — clean.

src/commands/node/tasks.ts
```diff
@@ -1759,6 +1759,54 @@ export class NodeCommandTasks {
...
+              .containers()
+              .readByRef(ContainerReference.of(config.podRefs[nodeAlias], constants.ROOT_CONTAINER));
+
+            await container.copyTo(overrideNetworkJson, ConsensusNodePathTemplates.DATA_CONFIG);
+
+            // copyTo lands the file as root; the node runs as hedera and would otherwise be denied access.
+            await container.execContainer([
+              'bash',
+              '-c',
+              `chown hedera:hedera ${ConsensusNodePathTemplates.OVERRIDE_NETWORK_JSON}`,
```
JeffreyDallas, line 1798: this makes nine hand-rolled `chown hedera:hedera` shell strings in `src/` (two in `helpers.ts`, one in `network-nodes.ts`, one in `network.ts`, four elsewhere in this file). `PlatformInstaller.setPathPermission` already wraps the pattern and is injected here as `this.platformInstaller` (line 347) — worth reusing?

One caveat that may be the reason not to: `setPathPermission` ends both commands with `2>/dev/null || true`, so a failed `chown` is swallowed. Given the PR description's point that a root-owned file is *silently* ignored by the node, failing loud here is the right behaviour and I'd keep it. So maybe the answer is a shared helper that doesn't swallow, rather than reusing that one — either way, nine copies of this string is a lot.

examples/state-save-and-restore/scripts/verify-override-network.sh
```diff
@@ -0,0 +1,68 @@
...
+
+POD=$(kubectl get pods -n "$TARGET_NAMESPACE" -l solo.hedera.com/type=network-node \
+  -o jsonpath='{.items[0].metadata.name}' 2>/dev/null)
+if [ -z "$POD" ]; then
+  fail "no consensus node pod in $TARGET_NAMESPACE"
+  exit 1
+fi
+echo "verifying $TARGET_NAMESPACE/$POD"
+
+in_pod() { kubectl exec -n "$TARGET_NAMESPACE" "$POD" -c root-container -- sh -c "$1" 2>/dev/null; }
```
JeffreyDallas, line 26: `2>/dev/null` means a failed `kubectl exec` is indistinguishable from "the grep found nothing". That's fine for the assertions testing *presence*, but assertion 2 tests *absence* — if the exec fails outright, `in_pod` returns empty and the AccessDeniedException check reports PASS on a pod it never reached.

Assertion 1 would already have failed in that scenario so the script still exits non-zero, but the output would point at the wrong thing. Capturing the exec status once up front (`in_pod 'true' || fail "cannot exec into $POD"`) would keep the failure legible.

src/commands/node/tasks.ts
jan-milenkov: updated

test/unit/commands/node/install-override-network-json.test.ts
jan-milenkov: updated

src/commands/node/handlers.ts
jan-milenkov: updated

src/core/consensus-node-path-templates.ts
jan-milenkov: updated

examples/state-save-and-restore/scripts/verify-override-network.sh
jan-milenkov: updated

src/commands/node/tasks.ts
jan-milenkov: Left as-is, filing separately. Agreed setPathPermission swallows failures.

JeffreyDallas (approved):

Issue #5722, opened 2026-08-06
design: self-upgrade mechanism (version check + binary swap)



jeromy-cannon:
Design the Solo self-upgrade flow:
- How to check for a newer release (GitHub Releases API polling cadence and trigger)
- UX for prompting the user (and what happens on decline)
- Strategy for downloading and atomically replacing the running SEA binary
- Error handling when the upgrade fails mid-download

Part of #5714.

### Dependencies
- [ ] required by #5728

Issue #5717, opened 2026-08-06
chore: set up code-signing infrastructure with Release Engineering



jeromy-cannon:
Coordinate with Release Engineering to provision signing credentials:
- **macOS** — Apple Developer ID certificate
- **Windows** — Microsoft Authenticode (EV) certificate

Configure the keys as CI/CD secrets. Both the macOS DMG and Windows NSIS installer must be signed before they can be distributed without OS security warnings.

Part of #5714.

### Dependencies
- [ ] required by #5723
- [ ] required by #5724

Issue #5725, opened 2026-08-06
feat: build Linux installer



jeromy-cannon:
Build the Linux installer using the tooling selected in the research task:
- Embeds the Solo SEA binary
- Runs `solo cache image pull` as a post-install step
- Runs the uninstall script on removal
- Targets the primary distributions determined during the research task

Part of #5714.

### Dependencies
- [ ] requires #5716
- [ ] requires #5720
- [ ] requires #5721
- [ ] required by #5727
- [ ] required by #5731

Issue #5721, opened 2026-08-06
design: define uninstall script behavior per OS



jeromy-cannon:
Define what a Solo uninstall must do on each OS: which Kind clusters to delete, which image caches and Solo cache directories to purge, and which files/directories to remove. Produce a shared design that the three OS-specific installer implementation tasks can implement against.

Part of #5714.

### Dependencies
- [ ] required by #5723
- [ ] required by #5724
- [ ] required by #5725

PR #5841, opened 2026-08-17
docs: add design proposal for linux installer

### LINKED ISSUE #5720: research: evaluate Linux installer tooling (makeself vs alternatives)
Evaluate `makeself` (used by the NMT product), AppImage, `.deb`/`.rpm` packages, and FPM as Linux installer candidates. Produce a recommendation with justification covering: install/uninstall UX, CI build complexity, and distribution target breadth.

Part of #5714.

### Dependencies
- [ ] required by #5725

Ivo-Yankov:
## Description

This PR adds a design doc for Solo's Linux installer.

**Recommendation: a `curl | sh` install script as the primary distribution channel**, with
nfpm-built native packages (`deb`/`rpm`/`apk`/`archlinux`) kept as an optional secondary channel
for later, built only if there's real demand for repo-based installs.

The two epic requirements — warm the image cache, clean up on removal — turn out not to need a
package manager at all: `postinst`/`preremove` can't safely do either (root, non-interactive, no
TTY, wrong `$HOME`), so the real logic already has to live in Solo itself — a first-run check and
a `solo uninstall` subcommand. A package's lifecycle hooks end up doing nothing more than printing
a notice and invoking that subcommand, which is thin payoff for the hosting/signing infrastructure
four repo formats need. An install script gets the same outcome — because it runs as the
installing user by construction, with no root/no-TTY/wrong-`$HOME` boundary to design around in
the first place — for a fraction of the cost, with direct precedent in `rustup`, Deno, Bun, uv,
and the OpenAI Codex CLI.

Key decisions:

- **Install layout:** per-user, no `sudo` by default — `~/.solo/bin/solo`, added to `PATH` by the
  script (mirrors rustup's `~/.cargo/bin`, Deno's `~/.deno/bin`). This is what lets self-upgrade
  work without elevated privileges. A `--prefix`/`SOLO_INSTALL_DIR` override covers users who
  explicitly want a system-wide install instead.
- **Integrity, not OS-mandated signing:** Linux has no Gatekeeper/Authenticode equivalent, so
  signing here defends against a compromised release-hosting step, not an OS requirement. CI
  publishes a `SHA256SUMS` manifest plus a **keyless `cosign`/sigstore signature** bound to the
  GitHub Actions workflow's OIDC identity — no key to generate, rotate, or custody, so nothing for
  #5717 to scope for this channel. The one gap this doesn't close: the script itself, piped via
  `curl | sh`, executes before any verification — only the binary it subsequently downloads is
  checked.
- **Self-upgrade (`solo update`):** an atomic same-filesystem rename over the current binary
  (safe on Linux — a running process keeps its already-open inode). Must detect install channel
  first (extending `HomebrewDeprecationNotifier`'s path-sniffing) so it never touches a binary
  owned by npm or a native package manager instead of self-replacing it. Sketched here as input to
  #5722, not a final design.
- **Lifecycle hooks stay thin if the native-package channel is ever built:** `postinstall` prints
  a notice instead of calling `solo cache image pull` directly; `preremove` invokes `solo
  uninstall` instead of deleting clusters itself.
- **nfpm over FPM**, if/when the secondary channel is built: a single static Go binary (17 open
  issues vs. FPM's Ruby+`rpmbuild` dependency and 791), with first-class `apk`/`archlinux` support.

Ruled out: AppImage (no install/uninstall lifecycle at all), makeself (superseded by the install
script, which gets the same no-`sudo` outcome with more familiar UX), raw `.deb`/`.rpm` built
independently (two pipelines for the coverage nfpm gets from one).

Open questions: pipe-to-shell trust perception at organizations with policies against it (mitigated
with a documented "download, inspect, then run" alternative, not eliminated); whether
`cosign`/sigstore keyless signing is workable against Solo's actual release pipeline; and — only if
the native-package channel is ever built — JFrog's Arch/pacman signing coverage and nfpm's built-in
signing coverage.

### Related Issues

* Closes #5720 
* Related to #

### Testing

* TBD

The following was not tested:

* TBD

<details>
<summary>
Commit message guidelines
</summary>

</details>


docs/design/installers/linux-installer-evaluation.md
```diff
@@ -0,0 +1,287 @@
+# Linux Installer Tooling Evaluation
+
+**Issue:** [#5720](https://github.com/hiero-ledger/solo/issues/5720)  
+**Epic:** [#5714](https://github.com/hiero-ledger/solo/issues/5714) — OS-Specific Solo Installers  
+**Blocks:** [#5725](https://github.com/hiero-ledger/solo/issues/5725) — Build Linux Installer
+
+## Background
```
JeffreyDallas: The macOS sibling proposal (#5873) lands in this same new `docs/design/installers/` directory and uses a different shape: numbered sections with explicit **Context**, **Goals**, **Non-Goals**, **Risks / Open Questions**, and **References**. This one has none of those four.

Since both land together and this is a brand-new directory, please align on one house pattern. Specifically, this doc is missing:

- **Non-Goals** — the macOS doc explicitly scopes out the uninstall design (#5721), signing CI (#5717), and self-upgrade (#5722). This one silently absorbs all three.
- **Risks / Open Questions** — there are several genuine ones here (musl, arm64, JFrog repo types, whether `postinst` may touch the network) and nowhere to put them.
- **References** — the doc makes claims about `dpkg`/`rpm`/`apk` repo metadata formats with no links.

Also worth settling: filename convention (`linux-installer-evaluation.md` vs `macos-dmg-installer.md`) and whether headings are numbered.

docs/design/installers/linux-installer-evaluation.md
```diff
@@ -0,0 +1,287 @@
...
+
+Solo currently distributes exclusively as an npm package (`npm install -g @hiero-ledger/solo`), which requires Node.js ≥ 22 and npm ≥ 9.8.1 to be pre-installed. The broader initiative is to ship native OS installers so end users need no runtime pre-installed. The Linux leg requires choosing a packaging format before building the installer.
+
+The chosen format must satisfy three requirements from the parent epic:
+
+1. Run `solo cache image pull` as a post-install step.
```
JeffreyDallas: This requirement is carried through the whole doc as if a package `postinst` hook satisfies it, but `postinst` is a hostile place to pull container images:

- it runs as **root**, non-interactive, with no TTY — nothing to prompt or render progress to;
- it may run with **no network** (image bakes, `apt install` inside a Dockerfile, offline mirrors), and Debian policy discourages network access in maintainer scripts;
- a non-zero exit **fails the whole package install** and leaves `dpkg` in a half-configured state.

There is a second problem specific to our SEA build. #5810 has the SEA bootstrap extract assets to `$HOME/.solo/sea-resources/<version>/`, so a root `postinst` invoking `solo cache image pull` populates `/root/.solo` — not the cache of the user who will actually run `solo`.

Can we design this as a first-run step inside Solo (or an opt-in `--with-image-cache` style debconf/preset) rather than a package maintainer script? Worth stating explicitly either way, because #5725 will otherwise implement the naive version.

docs/design/installers/linux-installer-evaluation.md
```diff
@@ -0,0 +1,287 @@
...
+2. Run a cleanup script on uninstall (remove clusters, image caches, residual files).
```
JeffreyDallas: "remove clusters" from a package `prerm` is a big behavioural claim to make in passing. `apt remove solo` running as root and deleting a user's Kind clusters is surprising and unrecoverable, and it isn't clear whose clusters would even be found from a root context.

The macOS proposal (#5873 §7.3) reaches the opposite conclusion for the same requirement — it puts this in a `solo uninstall` subcommand and defers the behaviour to #5721. Please cross-reference #5721 here and align the two docs, rather than having Linux delete clusters from `prerm` while macOS asks the user to run `solo uninstall`.

docs/design/installers/linux-installer-evaluation.md
```diff
@@ -0,0 +1,287 @@
...
+
+**Distribution target breadth**
+
+Universal. Any POSIX shell + `tar` + the chosen compression binary (`gzip` ships in every base image). Covers all nine distributions in the matrix with one artifact and no package-manager coupling.
+
+**Upgrade path**
+
+No built-in upgrade mechanism. A new release requires the user to download and re-run the latest `.run` file, which overwrites `/opt/solo/` in place.
+
+Solo already ships `VersionUpdateNotifier` (`src/core/version-update-notifier.ts`), a post-command banner that detects when a newer version is available. It currently checks the npm registry (`https://registry.npmjs.org/@hashgraph/solo/latest`) and caches the result for 24 hours in `~/.solo/cache/update-check.json`. For a makeself-installed binary two small adaptations are needed:
```
JeffreyDallas: `PACKAGE_NAME` is `@hiero-ledger/solo` (`src/core/constants.ts:22`), and `VersionUpdateNotifier.REGISTRY_URL` is built from it — the `@hashgraph/solo` URL was the pre-rename name.

```suggestion
Solo already ships `VersionUpdateNotifier` (`src/core/version-update-notifier.ts`), a post-command banner that detects when a newer version is available. It currently checks the npm registry (`https://registry.npmjs.org/@hiero-ledger/solo/latest`) and caches the result for 24 hours in `~/.solo/cache/update-check.json`. For a makeself-installed binary two small adaptations are needed:
```

docs/design/installers/linux-installer-evaluation.md
```diff
@@ -0,0 +1,287 @@
...
+
+**Install/uninstall UX**
+
+Same as native .deb/.rpm: either package-manager-first (with a hosted repository) or local-file install. Lifecycle hooks (`--after-install`, `--before-remove`) satisfy both post-install and uninstall requirements.
+
+**CI build complexity**
+
+Medium. A single `gem install fpm` step adds Ruby to the CI environment. After that, one FPM invocation per output format covers all target distributions with consistent metadata. The CI surface is significantly smaller than maintaining separate `.deb` and `.rpm` build pipelines.
+
+FPM supports `deb`, `rpm`, `apk`, and `pacman` output types, meaning all nine distributions in the matrix can be covered from one tool.
```
JeffreyDallas: This is the claim the recommendation rests on and I don't think it survives. Emitting an `.apk` does not make the payload runnable on Alpine — our own workflow already documents why, at `.github/workflows/flow-install-validation.yaml:92`:

> Alpine is musl, so the glibc runtime actions/setup-node downloads cannot execute; the apk-provided Node (22.x in 3.21) is used instead.

The SEA binary from #5810 is built on `ubuntu-latest` against a glibc Node, so it hits exactly the constraint this doc correctly used to disqualify AppImage on line 108. Covering Alpine means a musl-linked SEA build, which is a #5716/#5810 question, not a packaging-format question — and it should be called out as such.

Arch has the same kind of problem from the other direction: `pacman` can install an FPM-produced package, but the channel Arch users actually expect is the **AUR** (a community-maintained `PKGBUILD`), which the doc never mentions.

Same claim to correct at lines 205, 219 ("All 9 (4 output types)"), 271, and 275 — the comparison table row in particular should stay consistent with the `8/9 (no Alpine)` it gives AppImage.

docs/design/installers/linux-installer-evaluation.md
```diff
@@ -0,0 +1,287 @@
...
+
+**Upgrade path**
+
+Same as native .deb/.rpm: `apt upgrade solo` / `dnf upgrade solo` with a hosted repository, or manual re-download without one. Because FPM also generates `apk` and `pacman` packages, Arch and Alpine users get the same package-manager upgrade story if those repositories are hosted.
+
+**Repository hosting**
+
+APT and YUM/DNF repositories are static file trees — `apt` expects a signed `Packages.gz` + `Release`/`InRelease`; `dnf` expects a `repodata/` directory with `repomd.xml`. Three options are viable for Solo:
```
JeffreyDallas: This section covers repo *hosting* but not repo *signing*, which is the harder half and the part users see first (the `gpg --dearmor` line at 197 is the only trace of it).

FPM signs nothing. Each format needs its own path:

- **apt**: `debsigs`/`dpkg-sig` for the package, plus a GPG-signed `Release`/`InRelease` for the index;
- **dnf**: `rpmsign` with `%_gpg_name`, plus `gpgkey=` in the `.repo` file;
- **apk**: `abuild-sign` with an RSA key that must be installed into `/etc/apk/keys` on the client;
- **pacman**: `repo-add --sign` plus `pacman-key` trust setup.

So it is four repository types to host and sign, not two — and JFrog natively supports Debian and RPM but I don't believe it covers `apk` or `pacman`, which undercuts the "no new infrastructure" line in the recommendation.

Where the signing key lives, who holds it, and how it rotates should reference #5717 (signing infrastructure), the same way #5873 does.

docs/design/installers/linux-installer-evaluation.md
```diff
@@ -0,0 +1,287 @@
...
+| **Distro coverage** | All 9 (one artifact) | 8/9 (no Alpine) | 7/9 (no Arch, no Alpine) | All 9 (4 output types) |
+| **Repository infrastructure** | Not needed | Not needed | Required for best UX | Required for best UX |
+| **Build dependencies** | Shell + tar | appimagetool | dpkg-deb + rpmbuild | Ruby + fpm gem |
+| **Upgrade path** | Re-run new `.run` + version-check banner | AppImageUpdate delta patch (incompatible with lifecycle) | `apt upgrade` / `dnf upgrade` (requires hosted repo) | Same as .deb/.rpm |
+
+---
+
+## Bundle contents
+
+The makeself payload contains everything Solo needs at runtime — not just the SEA binary. Solo reads the `resources/` directory from disk at runtime (Helm values files, config YAML, shell scripts executed inside pods, CRD manifests, and node configuration templates). The post-build script already copies `resources/` alongside the compiled JS in the npm distribution; the same files must travel with the SEA binary in the installer.
```
JeffreyDallas: This is the main blocker. Everything from here to line 263 is a **makeself** design — "the makeself payload", `install.sh # makeself entry point`, `sudo ./solo.run`, "The `.run` archive is uploaded as a GitHub Release asset" — but the Recommendation on line 269 is FPM. It reads like the recommendation was flipped without the body following.

As it stands #5725 has an FPM verdict and a makeself build plan, and can't proceed. This section needs to become the FPM design: the source directory → package file map, the `--after-install` / `--before-remove` script contents, what lands in the `.deb`/`.rpm`/`apk`/`pacman` payload, and which of those become release assets vs. repo uploads.

The makeself material is still worth keeping — but as part of the ruled-out alternative, not as the implementation plan.

Separately, the premise of the section is stale: per #5810 the installer does not ship `resources/` at all (see the comment on line 253).

docs/design/installers/linux-installer-evaluation.md
```diff
@@ -0,0 +1,287 @@
...
+│   ├── crds/              # CRD manifests applied to the cluster
+│   ├── templates/         # application.properties, log4j2.xml, podman configs, etc.
+│   └── *.yaml / *.sh      # Helm values overrides, utility scripts run inside pods
+├── install.sh             # makeself entry point
+└── uninstall.sh           # installed to a known path during setup
+```
+
+`install.sh` is responsible for:
+
+1. Copying `solo` → `/usr/local/bin/solo`
```
JeffreyDallas: `/usr/local/bin` + `/opt/solo` is a reasonable makeself layout but it is not a valid layout for a distro package. `/usr/local` is reserved for the local administrator under the FHS and Debian Policy, and both `lintian` and `rpmlint` flag packages that write there.

For the FPM path this should be `/usr/bin/solo` with anything else under `/usr/lib/solo/`, or `/opt/solo/` with a `/usr/bin/solo` symlink. Worth stating the chosen layout explicitly so #5725 doesn't have to guess.

docs/design/installers/linux-installer-evaluation.md
```diff
@@ -0,0 +1,287 @@
...
+2. Copying `resources/` → `/opt/solo/resources/`
+3. Installing `uninstall.sh` → `/opt/solo/uninstall.sh`
+4. Running `solo cache image pull`
+
+`uninstall.sh` reverses steps 1–3 and runs any cluster/image-cache cleanup.
+
+### Coordination with the SEA build (#5716)
+
+The SEA binary needs to resolve the `resources/` path at runtime. In the current npm distribution this is relative to `__dirname` (or `import.meta.url`). In the SEA build there is no script path to anchor from, so the binary must use a fixed install path (`/opt/solo/resources`) or read it from an environment variable (e.g., `SOLO_RESOURCES_DIR`). The installer must write to whichever path the binary expects — this decision needs to be made in #5716 and matched here.
```
JeffreyDallas: This decision has already been made — #5810 implements #5716 and took a third option neither branch here anticipates: `resources/` is embedded as **SEA assets inside the binary**, and the SEA bootstrap extracts them to `~/.solo/sea-resources/<version>/` and sets `SOLO_SEA_ROOT_DIR` itself before the bundle loads (see `sea/build.ts` and the `constants.ts` change in that PR).

So there is no `/opt/solo/resources`, no `SOLO_RESOURCES_DIR` for the installer to set, and nothing for the installer to place — the resources travel inside the single binary and land per-user at first run.

Please read #5810 and rewrite this subsection accordingly. It also removes a dependency from #5725, which is good news. Two knock-on effects worth capturing while you're in here: the per-user extraction is why a root `postinst` can't warm the right cache (line 13), and `~/.solo/sea-resources/<version>/` becomes something the uninstall path has to clean up per user.

docs/design/installers/linux-installer-evaluation.md
```diff
@@ -0,0 +1,287 @@
...
+
+### Distribution
+
+The `.run` archive is uploaded as a GitHub Release asset as part of the existing release workflow, alongside the npm tarball and example archives. Users download it manually from the releases page or via:
+
+```sh
+curl -fsSL -o solo.run \
+  https://github.com/hiero-ledger/solo/releases/latest/download/solo-linux-x86_64.run \
```
JeffreyDallas: `x86_64` is the only architecture that appears anywhere in the doc, and the SEA matrix in #5810 is `ubuntu-latest` only — so also x64 only. arm64 Linux is not niche for this audience (Graviton, Ampere, Linux VMs on Apple Silicon).

Either add `aarch64`/`arm64` to the plan (it multiplies the package build matrix, and `deb` calls it `arm64` while `rpm` calls it `aarch64`) or state in Non-Goals that v1 is x86_64-only. Right now it is silent, which reads as an oversight rather than a decision.

docs/design/installers/linux-installer-evaluation.md
```diff
@@ -0,0 +1,287 @@
...
+  && chmod +x solo.run && sudo ./solo.run
+```
+
+---
+
+## Recommendation
+
+**Use FPM.**
```
JeffreyDallas: Two things I'd want before signing off on FPM specifically:

**1. `nfpm` isn't evaluated.** [`nfpm`](https://github.com/goreleaser/nfpm) (GoReleaser) produces `deb`, `rpm`, `apk`, and `archlinux` packages from one declarative YAML, ships as a single static binary, and needs no Ruby in CI. It is the direct modern competitor to FPM for exactly this job, and its absence is a real gap in "alternatives considered" — especially since "one tool, four formats" is the entire argument for FPM here.

**2. No tool-health assessment.** #5873 rejected `appdmg` with specifics — last release date, open issue count, unverified APFS support. FPM gets none of that: no release cadence, no assessment of whether `apk` and `pacman` are first-class or best-effort output types (they are considerably less exercised than `deb`/`rpm`), and no note that `gem install fpm` pins us to a Ruby toolchain in the release pipeline. Please apply the same rigour the sibling doc applied.

Also missing from the end of the doc: an **implementation / rollout plan** and an explicit **v1 scope cut**. #5725 needs to know which formats ship first, whether v1 is release-assets-only or hosted-repos, who owns the JFrog repository decision, and what is deferred (arm64, AUR, musl/Alpine, self-upgrade).

docs/design/installers/linux-installer-evaluation.md
```diff
@@ -0,0 +1,287 @@
+# Linux Installer Tooling Evaluation
+
+**Issue:** [#5720](https://github.com/hiero-ledger/solo/issues/5720)  
+**Epic:** [#5714](https://github.com/hiero-ledger/solo/issues/5714) — OS-Specific Solo Installers  
+**Blocks:** [#5725](https://github.com/hiero-ledger/solo/issues/5725) — Build Linux Installer
+
+## Background
+
+Solo currently distributes exclusively as an npm package (`npm install -g @hiero-ledger/solo`), which requires Node.js ≥ 22 and npm ≥ 9.8.1 to be pre-installed. The broader initiative is to ship native OS installers so end users need no runtime pre-installed. The Linux leg requires choosing a packaging format before building the installer.
```
JeffreyDallas: The doc never says what happens to the existing install channels. #5873 is explicit about this for macOS ("keeping the npm package as-is for developers who prefer it", Homebrew being retired), and this one should match — Homebrew runs on Linux too (`HomebrewDeprecationNotifier` already matches `/home/linuxbrew/.linuxbrew`), so Linux users may be on that channel today.

Also worth a sentence: a user without `sudo` can `npm i -g` or use Homebrew today, but cannot `apt install`. Is a plain tarball retained as the unprivileged fallback, or is root now a hard requirement on Linux?

jan-milenkov (approved):

PR #5843, opened 2026-08-17
fix: surface PVC mount diagnostics on pod readiness timeout

JeffreyDallas:
## Description

This pull request changes the following:

* When a pod fails to reach the ready state before Solo's readiness-wait timeout, `K8ClientPods.waitForRunningPhase` now inspects the PersistentVolumeClaims the pod depends on (via `pod.spec.volumes[].persistentVolumeClaim.claimName`) and any `FailedMount`/`FailedAttachVolume` Kubernetes events, and surfaces the result as a `volumeMountDiagnostic` on `KubePodNotReadyError` and `PodNotReadySoloError`.
* This distinguishes a stuck or slow volume attach/mount (e.g. an mdadm-backed storage class with variable RAID resync latency, as seen intermittently on Latitude.sh bare-metal clusters) from a generic pod-readiness timeout, so the error message and troubleshooting steps point at the volume subsystem instead of leaving the user with an unexplained timeout.
* The lookup is best-effort: any PVC/event API failure is swallowed and the original readiness timeout still surfaces unmodified. When the pod has no PVC-backed volumes, no extra API calls are made and the diagnostic is `undefined`.

### Related Issues

* Closes #5842

### Pull request (PR) checklist

* [x] This PR added tests (unit, integration, and/or end-to-end)
* [ ] This PR updated documentation
* [x] This PR added no TODOs or commented out code
* [x] This PR has no breaking changes
* [ ] Any technical debt has been documented as a separate issue and linked to this PR
* [ ] Any `package.json` changes have been explained to and approved by a repository manager
* [x] All related issues have been linked to this PR
* [x] All changes in this PR are included in the description
* [x] When this PR merges the commits will be squashed and the title will be used as the commit message, the 'commit message guidelines' below have been followed

### Testing

* [x] This PR added unit tests
* [ ] This PR added integration/end-to-end tests
* [ ] These changes required manual testing that is documented below
* [x] Anything not tested is documented

The following manual testing was done:

* N/A — this is a diagnostics-only change to an existing timeout path; verified via unit tests in `test/unit/integration/kube/k8-client-pods-wait-timeout.test.ts` covering both the diagnostic-present and diagnostic-absent cases, plus `task build`/`task format`/full unit test suite.

The following was not tested:

* Manual reproduction against a real Latitude.sh cluster with an mdadm-backed storage class was not performed (no such cluster available in this session); the fix was validated against mocked Kubernetes API responses only.

<details>
<summary>
Commit message guidelines
</summary>
[...]

PR #5876, opened 2026-08-20
fix: detect PVC mounts backed by undersized storage

### LINKED ISSUE #5875: Silent PVC mount failures: claim mounted on undersized storage reports a successful deploy
## Overview

Solo cannot detect a PersistentVolumeClaim that binds and mounts successfully but is backed by
storage far smaller than it requested, and reports the deployment as fully successful. This was hit
on a bare-metal cluster where the provisioner's host directory (`/opt`) was not on the intended
device: the mdadm array (`/dev/md0`) was not mounted there, so every claim was provisioned onto the
small system NVMe disk instead. The claims bound, the pods ran, no Kubernetes event was emitted, and
solo printed success.

This is invisible to the Kubernetes API by construction. Provisioners that hand out directories on
an existing filesystem (hostPath, local-path, most bare-metal setups) do not enforce
`spec.resources.requests.storage`, so the claim binds and reports the requested size back. The
`solo-deployment` chart requests roughly 865Gi per consensus node across 13 claims (500Gi
`dataSaved`, 100Gi each for the stream directories), which a directory provisioner will grant against
any disk of any size. The only place the truth is observable is inside the pod, in the size of the
mounted filesystem.

#5843 added a volume diagnostic to the pod-readiness timeout path, but it cannot reach this failure:
the claim is `Bound`, no `FailedMount`/`FailedAttachVolume` event exists, and the pod reaches
`Running` so the timeout branch never executes. Reviewing that path turned up five further gaps
where volume problems are either mis-detected or silently discarded.

## Sub-Tasks

- [ ] Verify each claim-backed mount inside the pod (`df` against the claim's requested size) and
      report mounts that provide materially less than requested; warn by default, fail on an opt-in
      flag, since kind and any single-disk cluster are legitimately oversubscribed
- [ ] Report pods that have no claim-backed mounts at all, so a deploy that silently ran on
      ephemeral storage (the `--pvcs` default of `false`) is visible
- [ ] Broaden the volume event reason set beyond `FailedMount`/`FailedAttachVolume` to include
      `FailedScheduling` (message-qualified, since it also fires for CPU/memory/taints),
      `ProvisioningFailed`, `FailedBinding`, and `VolumeBindingFailed` — the reasons a node-pinned
      local volume or a failing provisioner actually emits
- [ ] Pick the newest volume event by timestamp instead of by list position; the API guarantees no
      ordering, so a stale event from an earlier deploy in the same namespace can be reported as
      current
- [ ] Give the pod-never-created path volume context: a StatefulSet whose claim cannot be
      provisioned never creates its pod, and `KubePodNotFoundError` currently says only
      "no pod found"
- [ ] Translate `KubePodReadinessFailedError` in `KubeErrorTranslator`; it has no branch there, so
      every `waitForReadyStatus` caller falls through to `undefined` and no pod diagnostics
      (including the volume diagnostic) reach the user
- [ ] Stop the mirror-node importer wait from swallowing not-ready errors as "importer disabled via
      custom values"; an importer that exists but never starts is currently reported as a successful
      deploy

## References

- [PR #5843 — surface PVC mount diagnostics on pod readiness timeout](https://github.com/hiero-ledger/solo/pull/5843) — the partial fix this issue extends
- [solo-charts volumeClaims defaults](https://github.com/hashgraph/solo-charts/blob/main/charts/solo-deployment/values.yaml) — per-node claim sizes that no directory provisioner enforces

JeffreyDallas:
## Description

* Adds `PvcMountVerifier`, which verifies that each consensus node's PersistentVolumeClaim-backed mounts are actually backed by the storage the claims requested. Provisioners that hand out directories on an existing filesystem (hostPath, local-path, most bare-metal setups) do not enforce `spec.resources.requests.storage`, so a claim can bind, mount, and run against a disk far too small to hold it. That condition is invisible to the Kubernetes API — the claim reports `Bound` at the requested size, the pod reaches `Running`, and no event is emitted — so the verifier probes each mount path with `df -B1 -P` inside the pod and compares the result against the claim's request.
* Wires the verification into `network deploy` after the pods-running wait, skipped when `--pvcs` is off. It **warns by default** and fails only with the new `--verify-pvc-mounts` flag: kind and any single-disk cluster are legitimately oversubscribed by the chart's per-node requests (~865Gi across 13 claims) and would report the same shortfall, so a hard default would break existing deployments and CI.
* Reports pods that have no claim-backed mounts at all, which makes a deploy that silently ran on ephemeral container storage visible.
* Broadens the volume event reason set beyond `FailedMount`/`FailedAttachVolume` to include `ProvisioningFailed`, `FailedBinding`, `VolumeBindingFailed`, and `FailedScheduling`. These come from five different controllers and a given storage backend only emits a subset — a node-pinned local volume produces `FailedScheduling: … had volume node affinity conflict`, which the previous set missed entirely. `FailedScheduling` is message-qualified against `/volume|persistentvolumeclaim/i` so that `Insufficient cpu` is not misreported as a volume problem.
* Picks the newest volume event by timestamp instead of by list position. The API guarantees no ordering and a namespace retains events from earlier deploys, so `.at(-1)` could report a stale failure as the current one.
* Gives the pod-never-created path volume context. A StatefulSet whose claim cannot be provisioned never creates its pod, so `KubePodNotFoundError` was reporting only "no pod found"; it now carries a namespace-wide sweep of unbound claims (capped at five named claims plus a count) and the newest volume event.
* Adds the missing `KubePodReadinessFailedError` branch to `KubeErrorTranslator`. Every `waitForReadyStatus` caller sees that wrapper rather than the underlying cause, and with no branch there `tryTranslate` returned `undefined`, so none of the pod diagnostics — container summary, phase, or the volume diagnostic added in #5843 — ever reached the user.
* Narrows the mirror-node importer wait's `catch` to `KubePodNotFoundError`. Only an absent pod means the importer was disabled via custom values; an importer pod that exists but never starts (image pull, or a volume that cannot be provisioned or mounted) was being swallowed and reported as a successful deploy.

* Pins `SOLO_CHART_VERSION` to `0.66.1`, the release from hashgraph/solo-charts#285, and makes `--verify-pvc-mounts` set `defaults.volumeClaims.capacityCheck.enabled` on the chart as well. That chart value adds an init container which fails the pod **before** the consensus node starts, so an undersized volume stops the node rather than being caught after it has begun writing state. Without this wiring the chart bump would have no observable effect, since `capacityCheck` defaults to `false`. The post-deploy task in this PR remains the backstop for what the chart-side guard cannot cover — an older chart, or a values file that overrides the setting.

Supporting additions: `Pod.persistentVolumeClaimMounts` (resolved by joining `spec.volumes` to every container's `volumeMounts`, init containers included), `Pvcs.readAll()` returning binding phase and requested size, and `KubernetesQuantity` for converting Kubernetes quantity strings to bytes — which keeps `1Gi` (1073741824) and `1G` (1000000000) correctly distinct.

Probing is best-effort throughout: a container without a shell, an unavailable exec, or unparseable `df` output is logged at debug and skipped rather than reported as a storage problem, because an unverifiable mount is not the same as a bad one.

### Also carried on this branch: `--wraps-copy-parallel`

The `--wraps-copy-parallel` changes were originally reviewed on #5883, which is now closed. They are carried here so this PR's diff is fully accounted for:

* Adds `--wraps-copy-parallel` to `consensus network deploy`, letting the user choose whether the "Copy wraps lib into consensus node" step copies the WRAPs library to all consensus nodes at once or to one node after another.
* Converts that step's per-node copies into Listr subtasks — the same shape as the neighbouring "Copy JFR config file to nodes" step — with the flag selecting `concurrent`. Each node now gets its own titled subtask, so progress and failures are attributable to a specific node instead of being hidden inside one opaque step. The shared prep work (download, extract, prune, stage the tarball) stays sequential, since it operates on one local cache directory.
* The library is hundreds of megabytes per node, so on a large network these transfers dominate deploy time and running them concurrently is considerably faster. But concurrent copies put every transfer on the same link at once and each one shells out to `kubectl cp`, which is the wrong trade on a constrained or metered connection. Neither choice suits every cluster, hence a flag rather than a change of behaviour.
* **Defaults to `false`**, preserving exactly the sequential behaviour that exists today. Deployments that do not pass the flag are unaffected.
* Follows the flag checklist in `CLAUDE.md`: defined as a `CommandFlag`, added to `Flags.allFlags`, added to `NetworkCommand.DEPLOY_FLAGS_LIST.optional`, and added to the deploy step's disabled-prompt list so a non-interactive deploy is never blocked on it. It belongs to no special registry (`nodeConfigFileFlags`, `integerFlags`).

### Related Issues

* Closes #5875
* Related to #5843
* Depends on hashgraph/solo-charts#285 — merged there as `0.66.1`
* Includes the wraps-copy-parallel changes formerly reviewed on #5883 (see above)

### Before merge

* [x] The wraps-copy-parallel changes are carried here; #5883 is closed.
* [x] The chart pin now uses the released `0.66.1`.

No documentation update is included: this adds no environment variables, and no command, subcommand, or operation, so neither `docs/site/content/en/docs/env.md` nor `presentation_layer_cli_architecture.md` has a trigger. Both `--verify-pvc-mounts` and `--wraps-copy-parallel` appear in the generated CLI help, which is not a tracked artifact.

### Testing

* 26 new unit tests across four files. `PvcMountVerifier` is covered for the undersized case (a 500Gi claim on a 440GiB filesystem, the shape of the reported failure), the satisfied case, the filesystem-overhead tolerance boundary, a pod with no claim-backed mounts, an exec failure, unusable `df` output, an unknown requested size, and the no-matching-pods short circuit. The diagnostic changes are covered for a volume-caused `FailedScheduling`, a non-volume `FailedScheduling` that must be ignored, out-of-order events, the namespace-wide sweep on a never-created pod, the claim-count cap, and both translator paths.
* 241 tests pass across every suite touching this change: `test/unit/integration/kube/**`, `test/unit/business/utils/**`, `test/unit/core/errors.test.ts`, `test/unit/core/pvc-mount-verifier.test.ts`, `test/unit/commands/flags.test.ts`, plus the flag-prompt and PVC-validation command tests.
* For `--wraps-copy-parallel`: `npx tsx solo.ts consensus network deploy --help` renders `--wraps-copy-parallel [boolean] [default: false]`, confirming the flag reaches the CLI surface and not just the registry. Four unit tests cover the wiring that is easy to get wrong — the flag is in `Flags.allFlags` and `Flags.allFlagsMap`, its type is `boolean` with a `false` default, it is offered by `consensus network deploy`, and it has no prompt.
* On the combined branch (PVC verification, chart pin, and the flag together): `tsc` is clean and 235 tests pass.
* `task build` and `task format` are clean — zero lint errors.
* Confirmed the pinned chart is published and carries the guard: `helm template oci://ghcr.io/hashgraph/solo-charts/solo-deployment --version 0.66.1 --set defaults.volumeClaims.enabled=true --set defaults.volumeClaims.capacityCheck.enabled=true` renders `init-verify-volume-capacity` with the expected per-claim thresholds (`data/saved` → 524288000 KiB).

The following was not tested:

* The full `task test` run could not be completed locally: many unit tests resolve `K8Factory` at import time and the development machine has no current kube context, which aborts the run at `test/unit/commands/cluster.test.ts`. Verified to be pre-existing and unrelated by reproducing it on a clean checkout of `main` with these changes stashed. CI will give the full-suite signal.
* No end-to-end reproduction against a real bare-metal cluster with a misconfigured storage mount point. The verifier was validated against mocked pod, claim, and `df` responses.
* For `--wraps-copy-parallel`: no test asserts the resulting Listr concurrency, and no deploy was run in either mode. The concurrency decision is a single value passed to `task.newListr`, and the surrounding step is covered by the existing consensus network deploy end-to-end tests; asserting it directly would mean exposing the private task builder or driving a full deploy. Worth exercising `--wraps-copy-parallel` on a multi-node cluster before relying on it for a large network.

Two notes for reviewers, neither introduced here:

* When the chart-side guard trips, the init container fails and the pod sits in `Init:CrashLoopBackOff`. `detectFatalContainerError` treats only image-pull reasons and `OOMKilled` as fatal, so solo waits out the full `PODS_RUNNING_MAX_ATTEMPTS` (15 minutes) before failing. The resulting error does name the failing init container and its waiting reason, so it is diagnosable, just slow. Making a persistently crash-looping init container fail fast is a broader change than this PR should make.
* `SOLO_CONTAINERS_VERSION` in `version.ts` is `0.46.0` while the chart's root image tag is `0.46.1`, on `0.66.0` as well as this release, so the image cache warms a tag the chart never requests. Pre-existing drift, left alone here because changing which image is prepulled deserves its own review.


jan-milenkov (commented):

src/business/utils/kubernetes-quantity.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (!quantity) {
+      return undefined;
+    }
+
+    const match: RegExpExecArray | null = KubernetesQuantity.QUANTITY_PATTERN.exec(quantity.trim());
+    if (!match) {
+      return undefined;
+    }
+
+    const amount: number = Number.parseFloat(match[1]);
```
JeffreyDallas:
```suggestion
    const amount: number = Number(match[1]);
```
The regexp has already validated the complete decimal token, so use `Number()` here. Style guide §4.7.3 says not to use `parseFloat()` for base-10 parsing.

src/commands/network.ts
```diff
@@ -216,6 +222,60 @@ export class NetworkCommand extends BaseCommand {
...
+   * clusters where the storage layout is expected to be correct.
+   *
+   * That flag also enables the chart's `volumeClaims.capacityCheck` init container, which fails the
+   * pod before the consensus node starts. This task stays as the backstop that still reports when
+   * the chart-side guard is unavailable — an older chart, or a values file that overrides it.
+   */
+  private verifyPersistentVolumeClaimMounts(): SoloListrTask<NetworkDeployContext> {
+    return {
+      title: 'Verify persistent volume claim mounts',
+      skip: (context_): boolean => !context_.config.persistentVolumeClaims,
```
JeffreyDallas: what should happen when a user passes `--verify-pvc-mounts` without `--pvcs`? This skip succeeds without a probe or message, even though the flag description says it will fail an undersized PVC mount. Should the strict flag require or imply `--pvcs`, or should Solo report that it had no effect?

src/commands/flags.ts
```diff
@@ -1192,6 +1192,18 @@ export class Flags {
     },
   };
 
+  public static readonly wrapsCopyParallel: CommandFlag = {
```
jan-milenkov: In my opinion a CLI flag is overkill here.

Once it ships, removing it is a breaking change - and by the description's own account this one is unproven: no test asserts the resulting concurrency, and neither mode has been run on a real deploy.

An env var like `EXPERIMENTAL_COPY_WRAPS_LIB_IN_PARALLEL` lets us battle-test it and drop it quietly if it does not pay off, the same way `ENABLE_IMAGE_CACHE` and `SKIP_NODE_PING` work today. If it proves itself on a large network, promoting it to a flag later is easy — the reverse is not.

JeffreyDallas (commented):
@copilot address 
jan-milenkov review

src/core/constants.ts
```diff
@@ -84,6 +84,8 @@ export const SOLO_CLUSTER_ROLE_LABELS: Record<string, string> = {'solo.hedera.co
 export const SOLO_REMOTE_CONFIG_MAX_COMMAND_IN_HISTORY: number = 50;
 export const SOLO_REMOTE_CONFIGMAP_LABEL_SELECTOR: string = 'solo.hedera.com/type=remote-config';
 export const NODE_COPY_CONCURRENT: number = Number(getEnvironmentVariable('NODE_COPY_CONCURRENT')) || 4;
+export const EXPERIMENTAL_COPY_WRAPS_LIB_IN_PARALLEL: boolean =
```
Fantomasa, line 87: Small docs follow-up: this new env var will need an entry on the solo-docs env page (env.md moved out of this repo). Mind linking a companion docs PR?

JeffreyDallas: solo-docs PR https://github.com/hiero-ledger/solo-docs/pull/312

Fantomasa (approved):

jan-milenkov (approved):

PR #5879, opened 2026-08-21
fix: windows deploy self-poisons its own cache file

### LINKED ISSUE #5302: bug: one-shot deploy self-poisons ~/.solo on Windows — restrictToOwnerWindows empties a cache file DACL (SOLO-3035 EPERM), unrecoverable via destroy / rm -rf

jan-milenkov:
## Description

* **Windows: the cached PodLogs CRD file no longer gets bricked with an empty ACL.** Solo hardened the cached PodLogs CRD with folder-only `(OI)(CI)F` flags. `icacls` drops that grant on a file but still exits 0, and `/inheritance:r` then leaves the file with no permissions, so every deploy failed with `EPERM` and even `rm -rf ~/.solo` could not clean it up. Now the real path type is passed, the grant is verified before inheritance is removed, the CRD is hardened once at creation instead of on every deploy, and an unreadable cached file reports a dedicated Solo error with repair steps instead of a bare `EPERM`. The downloaded CRD YAML is also validated against its expected shape before it is cached, and cache writes are atomic (write-to-staging-file then rename) so a crash mid-download cannot leave a corrupt file behind.
* **Windows: the CLI could crash on exit right after the update-check request.** `VersionUpdateNotifier` used the pooled global `fetch`, whose underlying connection can finish tearing down on the libuv thread pool after `process.exit()` has already begun, aborting the process with `Assertion failed: !(handle->flags & UV_HANDLE_CLOSING)`. It now issues an unpooled `node:https` request that always closes with the response. Log flushing was also waiting on the wrong stream to close (the pino-pretty transform, which emits `close` before its file-backed destination has finished writing), which could let the process exit before the tail of `solo.log` was actually on disk; flush now waits on the real destinations.
* **`cluster-ref config reset` no longer leaves orphaned MinIO Operator CRDs behind.** Helm never deletes CRDs on uninstall, so `tenants.minio.min.io` and `policybindings.sts.min.io` used to survive every reset. The next `cluster-ref config setup` — including the one `one-shot deploy` runs internally during its own failure rollback — then found them with no owning Helm release and refused to proceed (`SOLO-3035`), with no way to recover short of a manual `kubectl delete crd`: a `destroy` → `deploy` cycle on the same cluster could never fix itself. Reset now deletes these CRDs itself, both after uninstalling a release it owns and when no release is found at all (which is the orphaned state itself), so a cluster in this state self-heals on the next reset instead of failing forever.

### Related Issues

* Closes #5302
* Also fixes a distinct `SOLO-3035` trigger (orphaned MinIO Operator CRDs after `cluster-ref config reset`) found and reproduced during manual testing on this branch; not filed as a separate issue

* [ ] All related issues have been linked to this PR
* [ ] When this PR merges the commits will be squashed and the title will be used as the commit message, the 'commit message guidelines' below have been followed

### Testing

* [ ] Anything not tested is documented

* Reproduced the MinIO Operator CRD orphan state on a local kind cluster via repeated `one-shot single deploy` / `one-shot single destroy` cycles, confirmed `cluster-ref config reset` now deletes the leftover CRDs (both after uninstalling a release it owns and when no release exists at all), and confirmed a subsequent `one-shot single deploy` no longer fails with `SOLO-3035`.

The following was not tested:

* The Windows-specific fixes (DACL hardening, `icacls` grant verification, update-notifier exit crash) were not re-verified on a Windows machine as part of this round of testing.

<details>
<summary>
Commit message guidelines
</summary>

</details>


JeffreyDallas (commented):
Can we update the window standard runner to reproduce Keith original steps:
install
remove
reinstall

JeffreyDallas (dismissed):

jeromy-cannon (dismissed):

JeffreyDallas (dismissed):

JeffreyDallas:
First one shot deploy working but got error and end of one shot destroy
```
Assertion failed: !(handle->flags & UV_HANDLE_CLOSING), file src\win\async.c, line 94
```



Then next re deploy failed

```

  √ solo cache image load [29s]
× Prepare cluster and deployment
  √ solo cluster-ref config connect [2s]
  √ solo deployment config create [2s]
  √ solo deployment cluster attach [2s]
  × solo cluster-ref config setup
  › solo keys consensus generate
› Create remote config components
› Deploy Solo components
› Finish
⚠ Container engine resources look low for a Solo deployment:
  detected:    11.7GB memory, 8 CPU(s)
  recommended: >=12GB memory, >=6 CPU(s)
  The deployment may fail or run slowly. Increase the memory/CPU allocated to
  Docker/Podman (or free up resources) before continuing.
╭─ ERROR ─────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ [SOLO-3035] Deploy failed: MinIO Operator CRDs exist in context 'kind-solo-cluster' with no Helm release owning │
│ them: tenants.minio.min.io, policybindings.sts.min.io. Rollback skipped (--no-rollback).                        │
│   → Delete the leftover CRDs, then run the setup again: kubectl delete crd tenants.minio.min.io                 │
│   policybindings.sts.min.io                                                                                     │
│   → Deleting these CRDs also deletes any MinIO Tenant resources defined by them                                 │
│   → They are usually left behind when the operator namespace was deleted directly rather than through solo      │
│   cluster-ref config reset                                                                                      │
│                                                                                                                 │
│ Learn more: https://solo.hiero.org/docs/troubleshooting/errors/component/SOLO-3035/                             │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```
No error in solo.log
```
[19:28:23.159] INFO: ==== Running 'cluster-ref config disconnect' === [traceId="74710ec7-c1c6-4595-8a8a-09ad84219596"]
[19:28:23.160] INFO: ==== Finished running 'cluster-ref config disconnect'==== [traceId="74710ec7-c1c6-4595-8a8a-09ad84219596"]
[19:28:23.208] INFO: Executing command (Windows): npm list --global --depth=0 [traceId="74710ec7-c1c6-4595-8a8a-09ad84219596"]
[19:28:26.212] INFO: ==== Running 'deployment config delete' === [traceId="74710ec7-c1c6-4595-8a8a-09ad84219596"]
[19:28:26.213] INFO: ==== Finished running 'deployment config delete'==== [traceId="74710ec7-c1c6-4595-8a8a-09ad84219596"]
[19:28:26.231] INFO: Remote config loaded [traceId="74710ec7-c1c6-4595-8a8a-09ad84219596"]
[19:28:26.310] INFO: Removing one-shot output directory: C:\Users\JeffreyTang\.solo\one-shot-one-shot [traceId="74710ec7-c1c6-4595-8a8a-09ad84219596"]
[19:28:26.385] INFO: ==== Finished running 'one-shot single destroy'==== [traceId="74710ec7-c1c6-4595-8a8a-09ad84219596"]
[19:28:26.880] INFO:
================================================================================ [traceId="74710ec7-c1c6-4595-8a8a-09ad84219596"]
[19:28:26.884] INFO:   A new version of Solo is available: 0.88.0 â†’ 0.88.1 [traceId="74710ec7-c1c6-4595-8a8a-09ad84219596"]
[19:28:26.886] INFO:   Upgrade guide: https://solo.hiero.org/docs/simple-solo-setup/upgrading-solo/ [traceId="74710ec7-c1c6-4595-8a8a-09ad84219596"]
[19:28:26.888] INFO:   Release notes: https://github.com/hiero-ledger/solo/releases [traceId="74710ec7-c1c6-4595-8a8a-09ad84219596"]
[19:28:26.889] INFO: ================================================================================
 [traceId="74710ec7-c1c6-4595-8a8a-09ad84219596"]
[19:28:26.890] INFO: Solo CLI completed, via entrypoint [traceId="74710ec7-c1c6-4595-8a8a-09ad84219596"]
[19:28:26.890] INFO: Flushing logs and exiting... [traceId="74710ec7-c1c6-4595-8a8a-09ad84219596"]
```
Tried twice, the same error

JeffreyDallas (commented):
@copilot fix failed unit test

      AssertionError: expected 'SOLO-5084' to equal 'SOLO-5083'
      + expected - actual

      -SOLO-5084
      +SOLO-5083

JeffreyDallas:
## Windows: `one-shot single destroy` aborts at exit with a libuv assertion

While testing this branch I hit a crash at the very end of an otherwise **fully successful** command:

```
√ Destroy [49s]
...
  A new version of Solo is available: 0.88.0 → 0.88.1
================================================================================

Assertion failed: !(handle->flags & UV_HANDLE_CLOSING), file src\win\async.c, line 94
```

Every task succeeded and all output was printed — the process then aborted during shutdown, leaving a
non-zero exit code. **This is a pre-existing defect on `main`, not something this PR introduced**, and it is
unrelated to the cache-file DACL work here. Fixed it on this branch (commit `02aa6cb7`) since this is where
it surfaced; happy to move it to its own issue/PR if preferred.

### Root cause

The version-update check is the last thing a command does, and it used Node's global `fetch` (undici):

1. `src/index.ts` calls `VersionUpdateNotifier.notifyIfUpdateAvailable()` after the command completes
2. `solo.ts` then flushes logs and calls `process.exit()`

`fetch` is a **pooled** client — it finishes tearing its connection down on the libuv thread pool after the
response is read. A worker thread signals completion via
[`threadpool.c:129`](https://github.com/libuv/libuv/blob/v1.52.1/src/threadpool.c#L129):

```c
uv__queue_insert_tail(&w->loop->wq, &w->wq);
uv_async_send(&w->loop->wq_async);
```

`process.exit()` closes `loop->wq_async` (setting `UV_HANDLE_CLOSING`). If a worker finishes inside that
window, the `uv_async_send` above lands on a closing handle and hits
[`src/win/async.c:94`](https://github.com/libuv/libuv/blob/v1.52.1/src/win/async.c#L94):

```c
assert(!(handle->flags & UV_HANDLE_CLOSING));
```

### Why Windows only

libuv has two separate `uv_async_send` implementations, and **only the Win32 one validates handle state**.
[`src/unix/async.c`](https://github.com/libuv/libuv/blob/v1.52.1/src/unix/async.c) has no such check — it just
sets a pending flag and writes a byte to a descriptor owned by the *loop*, never inspecting `handle->flags`.

So the underlying defect (exiting with thread-pool work in flight) is present on **all three platforms**;
macOS and Linux silently tolerate it. Windows is the only one that reports it. The fix removes a real, if
invisible, use-after-close on Unix too.

### Evidence

Reproduced in a harness driving the real `VersionUpdateNotifier` + real `SoloPinoLogger.flush()` +
`process.exit()`, i.e. the exact shutdown sequence. The trigger isolates cleanly to `fetch` — spawning child
processes (helm/kubectl/kind) alone never reproduced it (0/12):

| Variant | Aborts |
| --- | --- |
| `fetch` (current code) | **7 / 100** |
| `node:https` with `agent: false` | **0 / 100** |
| final patch, via the real `VersionUpdateNotifier` | **0 / 100** |

The ~7% rate explains why this looks random — it needs the worker to finish inside a narrow window.

### Fix

`VersionUpdateNotifier.fetchLatestVersion()` now uses `node:https` with `agent: false`, so this one-shot
request gets its own socket that closes with the response, leaving nothing in flight at exit. The 2s deadline
is preserved as an `unref()`'d timer that destroys the request. `registry.npmjs.org/<pkg>/latest` returns
`200` with no redirects, so losing `fetch`'s redirect-following costs nothing.

### Secondary finding: `SoloPinoLogger.flush()` (different bug, same area)

While tracing this I also found that `flush()` waits for the **pino-pretty transform** to close, but that
transform emits `close` as soon as its own side ends — while the `rotating-file-stream` behind it is still
writing and has yet to close its file descriptor. Exiting there can truncate the tail of `solo.log`. Fixed by
tracking the file-backed destinations and waiting on those instead of the transforms in front of them.

To be explicit, since it was my first hypothesis: **this is not what caused the abort.** I measured it —
3/40 with that change versus the same rate without. It is a genuine bug, but a separate one.

### Checks

- Full unit suite: 2127 passing vs 2125 baseline, same 3 pre-existing failures in both (a `BaseCommand`
  containerd stub, a `where helm` PATH lookup, and a flaky temp-dir cleanup)
- Typecheck and ESLint clean on all touched files
- Notifier tests retargeted from `globalThis.fetch` to `https.get`, plus two regression tests: one asserting
  `agent: false`, one asserting `flush()` waits on the destination behind a transform

JeffreyDallas (dismissed):

JeffreyDallas (dismissed):

JeffreyDallas (dismissed):

JeffreyDallas (approved):

PR #4981, opened 2026-07-09
fix: pass minimal explicit env to external command

Ivo-Yankov:
## Description

This PR removes all instances where the full local process environment was being passed to external commands. It introduces a mechanism for exposing specific environment variables to specific command types.

### Related Issues

* Closes #4954 
* Related to #

### Pull request (PR) checklist

* \[ ] This PR added tests (unit, integration, and/or end-to-end)
* \[ ] This PR updated documentation
* \[ ] This PR added no TODOs or commented out code
* \[ ] This PR has no breaking changes
* \[ ] Any technical debt has been documented as a separate issue and linked to this PR
* \[ ] Any `package.json` changes have been explained to and approved by a repository manager
* \[ ] All related issues have been linked to this PR
* \[ ] All changes in this PR are included in the description
* \[ ] When this PR merges the commits will be squashed and the title will be used as the commit message, the 'commit message guidelines' below have been followed

### Testing

* \[x] This PR added unit tests
* \[ ] This PR added integration/end-to-end tests
* \[x] These changes required manual testing that is documented below
* \[x] Anything not tested is documented

The following manual testing was done:

The per-command allowlists were validated empirically by observing what each external command actually reads from the environment:

* Built a Linux container (Ubuntu + Docker-in-Docker) with solo, the pinned CLIs, and `strace`, then ran the real commands (`init`, `one-shot single deploy` / `destroy`) plus isolated tool invocations under `strace`.
* For each spawned command, traced `execve` (which vars were *passed*) and `openat`/`connect` (which vars were *actually consumed* — i.e. a var's value showing up as a file opened or socket dialed), attributing effects across the Go tools' worker threads.
* Verified minimization end-to-end with a non-allowlisted "leak canary": on this branch it reached **no** spawned child (vs. every child on `main`), while `init` and the mirror/relay/explorer deploy still succeeded on the reduced env.
* Coverage: `helm`, `kubectl`, `kind`, `docker`/`crane`, `npm`, `gh`, and podman-backed `kind` (`KIND_EXPERIMENTAL_PROVIDER=podman`).
* This process caught the gaps fixed here: `helm`/`kind` reading `DOCKER_CONFIG`, podman-backed `kind` reading `CONTAINERS_STORAGE_CONF`, and three spawn sites still inheriting the full env (`checkDockerImageExists`'s `docker images`, `ps`, `icacls`).

The following was not tested:

[...]

Issue #4954, opened 2026-07-07
security: pass minimal explicit env to external command execution


jan-milenkov:
## Description

Several execution paths still spread the full parent environment into child processes:

- `src/core/shell-runner.ts` — `env: {...process.env, ...environmentVariablesToAppend}`
- `src/integration/helm/execution/helm-execution-builder.ts` — `{...process.env}`
- `src/integration/kind/execution/kind-execution-builder.ts` — `{...process.env}`
- `src/integration/kube/k8-client/resources/container/k8-client-container.ts` — `{...process.env, PATH: ...}`
- `src/integration/kube/k8-client/resources/pod/persist-port-forward.ts` — `{...process.env, PATH: ...}`

Tasks:
- For each external command (kubectl, helm, kind, docker/podman, gh, brew),
  determine the minimal set of env vars it actually needs (PATH, KUBECONFIG,
  HOME, etc.).
- Build and pass that minimal set explicitly instead of spreading the whole
  process environment.
- Verify unit and E2E suites still pass on macOS, Linux, and Windows.

PR #5897, opened 2026-08-26
fix: pass EKS workload-identity variables to helm and kubectl

### LINKED ISSUE #5895: `AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE` are stripped from helm
### Describe the bug

solo is stripping env variables like **`AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE`** breaking EKS IRSA. this was introduced at some point as solo 0.76 doesn't have this issue 

helm should be able to use native EKS IRSA for authentification when executed from solo

### To Reproduce

1. Set env vars for solo like **`AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE`**
2. replace helm with stub to record env variables
3. see helm not getting these env vars 

### Additional Context

The following env vars should be whitelisted and passed to the helm and other tooling when running in AWS EKS cluster to utilize  EKS IRSA
```
`AWS_ROLE_ARN`, 
`AWS_WEB_IDENTITY_TOKEN_FILE`, 
`AWS_REGION`, 
`AWS_DEFAULT_REGION`, 
`AWS_STS_REGIONAL_ENDPOINTS`
```

JeffreyDallas:
## Description

EKS IRSA authentication is broken: Solo filters the environment it passes to external commands, and the variables the AWS credential plugin needs were never on the allowlist. This adds them, and closes the two gaps that made the failure expensive — it was silent, and it could only be fixed by a release.

* Adds the exact names the AWS credential plugin requires to the `KUBECTL` and `HELM` profiles.
* Reports withheld variable names in `solo.log` at `info`, so "was my variable filtered out?" is answerable from the log the user already has.
* Adds `subprocess.additionalEnvironmentVariables`, a per-command, config-file-only setting for forwarding extra names without waiting for a release — with the file validated before it is trusted.

### Root cause

`SubprocessEnvironment` was introduced in #4954 / #4981 (v0.83.0) to stop Solo spreading the full parent environment into every external command — which is why 0.76 is unaffected. On EKS, `kubectl`/`helm` authenticate through a kubeconfig `exec` credential plugin (`aws eks get-token`) that reads `AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE` from the environment. Neither was enumerated, so IRSA had nothing to authenticate with.

#4981 was validated carefully — `execve`/`openat`/`connect` traced under `strace` to observe what each tool actually consumed, which caught real gaps (`DOCKER_CONFIG` for helm/kind, `CONTAINERS_STORAGE_CONF` for podman-backed kind). The gap is that it ran against local **kind** clusters, where the `exec` credential-plugin path is never invoked: the kubeconfig carries embedded certs. No amount of care enumerating an allowlist protects against a code path the test environment never exercises.

### Why this stays an allowlist

An earlier revision of this PR inverted the filtering to a name-shaped denylist. **That was wrong and was reverted.** The argument was that a missed secret is marginal, since every profile already receives `HOME`, `PATH` and `KUBECONFIG` and a tool could read `~/.aws/credentials` off disk anyway.

That reasoning covers *confidentiality* only, and it assumes the threat is a compromised binary. The real threat is an **honest binary and a poisoned environment** — a compromised earlier CI step, a malicious `.envrc`, a tampered shell profile. An attacker who can set environment variables but cannot replace the binary still gets code execution and MITM through variables carrying no secret-shaped name at all:

| Vector | Effect |
| --- | --- |
| `LD_PRELOAD`, `DYLD_INSERT_LIBRARIES`, `NODE_OPTIONS=--require`, `BASH_ENV` | arbitrary code execution inside a process holding cluster-admin |
| `SSL_CERT_FILE`, `CURL_CA_BUNDLE`, `AWS_CA_BUNDLE`, `NODE_EXTRA_CA_CERTS` | trust an attacker CA, MITM the Kubernetes API |
| `AWS_ENDPOINT_URL`, `AWS_CONFIG_FILE` | redirect STS, exfiltrating the very IRSA credentials this PR enables |
| `GIT_SSH_COMMAND`, `GIT_EXTERNAL_DIFF`, `PS4` | command execution |

Measured against those 19 variables: the proposed denylist forwarded **19/19** to `helm`; this allowlist blocks **18/19**. A unit test pins this so a future attempt to loosen it fails loudly.

The one that still gets through is `HELM_DRIVER`, via the pre-existing `HELM_` *prefix* rule from #4981 — not introduced here, left alone as out of scope, and noted below.

### What is allowlisted, and what is not

Added to `KUBECTL` and `HELM` — each an identifier, a region selector, or a path to a token the platform placed on disk:

* `AWS_ROLE_ARN`, `AWS_WEB_IDENTITY_TOKEN_FILE`, `AWS_REGION`, `AWS_DEFAULT_REGION`, `AWS_STS_REGIONAL_ENDPOINTS`, `AWS_PROFILE`

**AWS only, deliberately.** An earlier revision also added GKE and AKS equivalents on the reasoning that they are the same bug. They were removed:

1. **They were never verified.** No GKE or AKS cluster was available, so those names came from knowledge rather than observation — repeating exactly the mistake that caused #5895, and implying support nobody has exercised.
2. **`AZURE_AUTHORITY_HOST` proved the risk.** Review identified it as an authority-endpoint override — the Azure twin of `AWS_ENDPOINT_URL`, which this PR excludes on precisely that reasoning. One of five speculative Azure names was a security defect.
3. **The config setting removed the justification.** The case for pre-adding was "otherwise a GKE operator waits for a release". They no longer do.

GKE and AKS are no worse off than before this PR — they were already broken, and the docs give them a worked opt-in example. When someone verifies the required set against a real cluster, it belongs in the built-in list.

**Excluded on purpose**, with the reasoning in code so it survives: `AWS_CA_BUNDLE` and `AWS_ENDPOINT_URL` (redirect TLS trust and the STS endpoint), `AWS_CONFIG_FILE` and `AWS_SHARED_CREDENTIALS_FILE` (an AWS config file may declare `credential_process`, i.e. arbitrary command execution — the default location under `HOME` is still honored), and `AZURE_AUTHORITY_HOST`.

### Making the failure visible

An allowlist fails closed **and silently**. That is what made this issue expensive: the reporter had to replace the `helm` binary with a recorder to discover which variable was missing.

`SubprocessEnvironment.configureWithheldReporter()` installs a single sink at startup which reports withheld variable **names** — never values. `SubprocessEnvironmentBootstrap` wires it to the logger at **`info`**.

* **`info`, not `debug`.** `solo.log` is written at `SOLO_LOG_LEVEL`, default `info`, so a `debug` line only appears if the user already suspects environment filtering. At `info` the answer is always already in the log.
* **A single sink, not a per-call callback.** An earlier revision threaded a callback through call sites and wired up two of nine — the ones missed were the direct `kubectl` spawns in `K8ClientContainer` and port forwarding, which filtered the environment with no diagnostic at all. A central sink covers every site by construction, within one process.
* **Unfiltered, but bounded.** A heuristic for "interesting" names would hide exactly the user-defined variable someone is searching for, so every conforming name is emitted — chunked at 100 per line rather than truncated, with a 2000-name ceiling and the residue reported as a count. Names are restricted to a conservative identifier shape, since an env var name may contain newlines on POSIX and could otherwise forge log lines.
* Emitted **once per command profile per run**; the withheld set does not change during a run.
* **Persistent port-forwarding crosses a process boundary, and the sink does not cross with it.** `persist-port-forward.ts` runs `kubectl port-forward` from a detached worker process so it survives the parent Solo process exiting. That worker is deliberately dependency-free (it cannot pull in the DI container), so it never runs `SubprocessEnvironmentBootstrap` and starts with no operator allowlist and no reporter of its own — `SubprocessEnvironment`'s vetted-names state is per-process. Left alone, this both dropped a `subprocess.additionalEnvironmentVariables.kubectl` entry silently at the worker's own re-filtering step and made it invisible, since nothing in that process reports withheld names. The parent now passes its already-vetted per-profile names to the worker as an argument, which re-seeds them via the same `configureOperatorAllowlist()` and installs its own console-based reporter before spawning `kubectl`. A process-level regression test (`test/unit/integration/kube/persist-port-forward.test.ts`) spawns the real worker against a recording `kubectl` to pin this.

### Extending the allowlist without a release

```yaml
# ~/.solo/solo-config.yaml  (optional; absent by default)
subprocess:
  additionalEnvironmentVariables:
    helm:
      - MY_PLATFORM_SETTING
    kubectl:
      - MY_PLATFORM_SETTING
```

Exact names only, **declared per command** — a variable added for a credential plugin has no business reaching `npm` or a container engine, and a flat list would discard the containment the allowlist exists to provide.

**The file is `solo-config.yaml`, not `solo.yaml`.** `~/.solo/solo.yaml` is a leftover from older Solo versions holding an unrelated `flags:` map, and the `solo:config:remove` task in hiero-consensus-node's CITR Taskfile deletes it automatically — an operator's settings would have been silently wiped by test tooling. A stale legacy file alongside the real one parses harmlessly; both facts are pinned by tests.

**Loaded at startup, not in the `ConfigProvider` factory.** Config sources load asynchronously, so a value read in that synchronous factory resolves to the schema default and the setting silently does nothing — an earlier revision had exactly that bug. Refreshing the merged config there is not an option either: `LayeredConfig.refresh()` refreshes *every* source, which would newly activate bundled resource files that currently never load. `OptionalDefaultConfigSource` lets the operator file join the cascade without breaking `refresh()` when it does not exist.

**Three safeguards, because a config file is a trust boundary rather than a guarantee.** Without them, "attacker can write `~/.solo/solo-config.yaml`" becomes "attacker gets code execution on the operator's machine inside a cluster-admin process", via a one-line edit adding `LD_PRELOAD`.

1. **A refusal set config cannot override.** Loader/interpreter hooks, TLS trust overrides and credential/endpoint redirection are refused unconditionally and case-insensitively, and each refusal is logged. Matched **by prefix as well as exact name** for `AWS_ENDPOINT_URL`, `LD_` and `DYLD_`: AWS honours `AWS_ENDPOINT_URL_<SERVICE>`, and `AWS_ENDPOINT_URL_STS` takes precedence over the global setting, so exact-name refusal alone left the EKS credential exchange redirectable by a one-line config entry.
2. **The one config key that cannot be set from the environment.** `EnvironmentConfigSource` drops `subprocess.*` and warns. A setting that relaxes environment filtering must not be controllable by the environment being filtered.
3. **The file is validated before it is trusted.** Solo does not own this file's permissions — the operator creates it — so it checks them. See below.

**Shape is validated at runtime, not just by TypeScript.** `helm: MY_VAR` is valid YAML that would be iterated character by character, quietly allowlisting `M`, `Y`, `_`; a mapping threw a raw `TypeError` outside the registered-error path. Both now raise `SubprocessConfigInvalidValueSoloError` (`SOLO-4084`) naming the offending key.

**Failures are surfaced, not swallowed.** Absence is silent (the common case). A present-but-unloadable file raises `SubprocessConfigLoadFailedSoloError` (`SOLO-4082`); an untrusted one raises `SubprocessConfigUnsafePermissionsSoloError` (`SOLO-4083`). An operator believing a security-relevant setting applied when it silently did not is the worse outcome.

### What the permission checks do and do not guarantee

Stated precisely, because it is easy to overstate:

* **Guaranteed, on POSIX** — the file's contents. It is opened with `O_NOFOLLOW`, `fstat`ed, validated, and read from that same descriptor, so the bytes parsed are the bytes checked.
* **Best-effort** — the directory checks. They catch a static misconfiguration such as a group-writable `SOLO_HOME`; they do **not** stop an attacker who already has write access on the path from swapping a component between check and open. Component-by-component opens (`openat`) are not available through Node's `fs` API.
* **POSIX** walks to the filesystem root, accepting directories owned by the user or root, plus sticky directories such as `/tmp`.
* **Windows** reads the DACL with `icacls` and refuses any write grant to a principal other than the user, `SYSTEM`, `Administrators` or `CREATOR OWNER`. Inherit-only `(IO)` and deny `(DENY)` ACEs are skipped. The walk stops before the volume root, since `C:\` legitimately carries broad write grants. There is no `O_NOFOLLOW`, so the symlink check is a check-then-open.

**No Windows ACL integration test exercised real `icacls`/filesystem behavior.** `windows-2022` CI runs the fixture-based unit suite and passed, but the fixtures are captured `icacls` *text*, not a live invocation against a real Windows filesystem — so the matching logic is verified, but not that `icacls` emits what is assumed or that the surrounding filesystem behavior matches. The docs say so and advise against relying on the setting there yet.

### Related Issues

* Closes #5895
* Related to #4954
* Companion docs PR: hiero-ledger/solo-docs#308

### Review follow-ups applied after the initial description

* **`OptionalDefaultConfigSource` now creates its `basePath` directory.** On a first run with a fresh `SOLO_HOME` the directory does not exist yet, which made the optional source throw instead of treating the absent file as the no-op it is meant to be. Covered by the new test above.
* **`icacls` is invoked with an explicit `shell: false`.** The `execFileSync` calls in the Windows ACL reader relied on the default rather than stating it, so a path containing shell metacharacters was one refactor away from being interpreted.
* **The ancestor trust walk uses `while (true)` rather than `for (;;)`,** per review — the intent reads more directly.

**A bad conflict resolution briefly reverted CI hardening, and has been undone.** Commit `fd6a2ce02` overwrote `flow-examples-test.yaml` and `zxc-e2e-test.yaml` with a stale copy, rolling back `persist-credentials: false` (#5953), the `$/` self-repository syntax (#5948), the Solo Bin Cache steps (#5851), the `env:` indirection keeping `inputs.*` out of `run:` blocks, the inline `zizmor: ignore[cache-poisoning]` annotations, three action version bumps, and the least-privilege `permissions:` blocks — while also unpinning `grpcurl` from its commit SHA, despite its own message claiming to pin it. Both files are now byte-identical to `main`. Neither needed changing for this PR: `one-shot-falcon` is already in `examples-test-matrix.json`, and the new `test-env-passthrough` task runs from the example's existing `test` task.

### Out of scope, flagged not fixed

* **`HELM_DRIVER`** reaches helm via the pre-existing `HELM_` prefix rule from #4981. `HELM_` and `HOMEBREW_` are the only remaining wildcards, and the "exact matches only" principle argues against both.
* **`PATH`** is inherited from the base allowlist. It was briefly on the refusal set; a test caught that as misleading, since it implies a protection that does not exist.
* **`resources/config/tss-config.yaml` disagrees with the schema defaults** — the file pins WRAPS `v0.2.0`, the schema defaults to `v1.0.0`, and since nothing awaits `refresh()` at startup the schema default silently wins. Found while establishing when config sources actually load.
* **`test/unit/commands/cluster.test.ts`** fails at module load with `unregistered dependency token "Symbol(K8Factory)"` when no kubectl current-context is set. Reproduced on `origin/main`, so it predates this PR, but it makes the unit suite environment-dependent.

* \[x] This PR updated documentation

No CLI or API surface changes. Subprocess environment contents change materially for `helm` and `kubectl` — that is the fix.

### Testing

* \[x] This PR added integration/end-to-end tests

**Unit — 2168 passing, 0 failing.** `task build` clean, `git diff --check` clean.

`test/unit/core/subprocess-environment.test.ts` — regression guard for #5895 (all five IRSA variables reach `helm` and `kubectl`); GKE/AKS names are *not* forwarded by default; trust/credential/endpoint redirects withheld from every profile; **the code-execution and MITM vectors above withheld from every profile**, which is what pins the allowlist decision; per-command operator entries; `AWS_ENDPOINT_URL_STS` and unlisted `LD_*`/`DYLD_*` refused by prefix; the diagnostic emits every conforming name across chunked lines with a ceiling test.

`test/unit/core/subprocess-environment-bootstrap.test.ts` — integration through real `Container.init()` and a real YAML file in a temporary home: a variable declared for `helm` reaches `helm` and not `npm`; refused entries dropped and warned; malformed YAML, unreadable file, group-writable file, symlinked file, group-writable directory and group-writable *ancestor* each refused; scalar/mapping/non-string shapes rejected; the reporter covers a direct `kubectl` spawn; absent file is a silent no-op.

`test/unit/business/utils/file-permissions-windows-acl.test.ts` — 12 fixture tests over captured `icacls` output, so the security-critical matching is covered on every platform: inherited `(OI)(CI)(M)` rejected (the fail-open regression), inherit-only `(IO)(M)` on `C:\` accepted, `WDAC`/`WO`/`D`/`DE` rejected, `(DENY)(W)` not read as a grant while a separate allow entry for the same principal still is, principals containing spaces matched, `(RX)` accepted.

`test/unit/integration/kube/persist-port-forward.test.ts` — a process-level regression test for the gap review caught: `subprocess.additionalEnvironmentVariables.kubectl` did not survive persistent port-forwarding, because the detached worker process re-filters its own environment with no operator allowlist of its own. Spawns the real `persist-port-forward.ts` worker via `tsx` against a recording fake `kubectl`, and asserts an operator-configured variable reaches the real `kubectl` spawn while an unconfigured one does not; fails against the pre-fix code (missing from the worker's environment) and passes once the parent forwards its vetted names to the worker. POSIX only — the fake `kubectl` relies on a shebang.

`test/unit/data/configuration/impl/environment-config-source.test.ts` — `subprocess.*` cannot be set from the environment: the source drops those keys and warns, so the setting that relaxes environment filtering is not itself controllable by the environment being filtered.

`test/unit/data/configuration/impl/optional-default-config-source.test.ts` — the optional source is a no-op when the file is absent, loads it when present, and creates its `basePath` directory rather than throwing when `SOLO_HOME` does not yet exist.

**Manual.** Verified end-to-end against a throwaway kind cluster using non-deprecated commands (`cluster-ref config connect`, `cluster-ref config setup`) with an isolated `SOLO_HOME`:

```
helm     ACCEPTED-withheld=0 CONTROL-withheld=1
kubectl  ACCEPTED-withheld=0 CONTROL-withheld=1
npm      ACCEPTED-withheld=2 CONTROL-withheld=2
generic  ACCEPTED-withheld=1 CONTROL-withheld=1
```

The declared variable is absent from the withheld list for exactly the two commands it was configured for. Re-running with the config file removed makes the assertions fail (`exit=1`), so the check is not vacuous. `examples/one-shot-falcon` carries this as a `test-env-passthrough` task using a throwaway `SOLO_HOME` under `.tmp` — verified by snapshotting `~/.solo` around a run that it never touches the operator's real config.

The following was not tested:

* **No live EKS cluster.** The credential-plugin path is covered by unit assertions on environment contents, not an end-to-end `aws eks get-token` exchange. A reviewer with EKS access confirming against an IRSA-backed cluster before merge would be worth more than any unit test here — that is precisely the coverage gap that let #5895 through.
* **No Windows ACL integration test.** `windows-2022` unit CI and the Windows Deploy/Destroy job both passed, but the ACL parser is covered only by fixtures against captured `icacls` text, which validates the matching logic but not that `icacls` emits what is assumed, nor the surrounding filesystem behaviour end-to-end.
* GKE and AKS opt-in is documented but unexercised.

<details>
<summary>
Commit message guidelines
</summary>

</details>


nathanklick (dismissed):
A catch all allow opens the door to attacks. We should only allow exact matches required by a given command to be passed through to the subprocess environment.

src/core/subprocess-environment.ts
```diff
@@ -68,13 +122,189 @@ export class SubprocessEnvironment {
...
+    // GCP / GKE (gke-gcloud-auth-plugin)
+    'GOOGLE_APPLICATION_CREDENTIALS',
+    'CLOUDSDK_CONFIG',
+    'CLOUDSDK_CORE_PROJECT',
+    'USE_GKE_GCLOUD_AUTH_PLUGIN',
+    // Azure / AKS workload identity
+    'AZURE_FEDERATED_TOKEN_FILE',
+    'AZURE_CLIENT_ID',
+    'AZURE_TENANT_ID',
+    'AZURE_AUTHORITY_HOST',
```
JeffreyDallas: `AZURE_AUTHORITY_HOST` is the Azure AD authority endpoint, so this is the same class of trust-routing input that we correctly exclude for `AWS_ENDPOINT_URL`. This needs a tighter boundary. `kubelogin` reads it in workload-identity mode and passes it into the Azure cloud configuration while consuming `AZURE_FEDERATED_TOKEN_FILE` ([source](https://github.com/Azure/kubelogin/blob/66784961ef44a34236a651d9b08bcdf0aa07b836/pkg/internal/token/options.go#L255-L264), [source](https://github.com/Azure/kubelogin/blob/66784961ef44a34236a651d9b08bcdf0aa07b836/pkg/internal/token/workloadidentitycredential.go#L47-L60)).

The current Azure SDK normally validates unknown authorities through instance discovery, but its own API says `DisableInstanceDiscovery` makes the application responsible for trusting the authority; `kubelogin` exposes that mode for disconnected/private clouds. In that supported configuration, an environment attacker can redirect the token exchange to an attacker-controlled HTTPS authority and receive the projected assertion. This conflicts with the exact-per-command, integrity-first rule this class documents and with TypeScript guide §10.2.1's service-boundary discipline.

Please either constrain this to a reviewed set of Azure authority hosts, or obtain an explicit private-cloud authority from a trusted, non-environment config source. Add a regression test for the private-cloud/disabled-discovery path; do not rely on a downstream default that the kubeconfig can switch off.

src/core/subprocess-environment.ts
```diff
@@ -155,13 +399,16 @@ export class SubprocessEnvironment {
         ...SubprocessEnvironment.COMMON_ALLOWLIST,
         ...(onWindows ? SubprocessEnvironment.WINDOWS_ALLOWLIST : []),
         ...SubprocessEnvironment.COMMAND_ALLOWLIST[profile],
+        // Operator-configured extras, already vetted by configureOperatorAllowlist().
+        ...SubprocessEnvironment.operatorAllowlist,
```
JeffreyDallas: This applies each operator-approved name to *every* profile. The new test intentionally proves that `MY_PLATFORM_SETTING` reaches all profiles, so a credential required only by a `kubectl`/`helm` exec plugin is also exposed to `npm`, `brew`, `gh`, and container-engine subprocesses when those run. That is a confidentiality and integrity regression from this class's documented minimal per-command environment, and violates the established exact-names-required-by-a-given-command convention (TypeScript guide §10.2.1).

Make the configuration profile-scoped (at minimum `kubectl` and `helm`) and test both sides: the extra name reaches only its declared profile and is absent from every other profile. Better still, add new cloud requirements to the reviewed built-in per-profile list after diagnosis rather than creating a global generic escape hatch. A generic refusal list cannot establish safety for arbitrary variables across every current and future child program.

JeffreyDallas (commented):
## Intent

Restore managed-cloud workload-identity authentication for `helm` and `kubectl` while retaining Solo's deny-by-default subprocess environment.

## Verdict

Request changes

## Critical

- None.

## Major

- `AZURE_AUTHORITY_HOST` is an authority-endpoint override, not a passive identifier — `src/core/subprocess-environment.ts:165`. Under the supported private-cloud / `--disable-instance-discovery` execution mode, a poisoned parent environment can direct the workload-identity plugin to an attacker authority while it reads the projected federated token.
- `subprocess.additionalEnvironmentVariables` is global rather than per-command — `src/core/subprocess-environment.ts:403`. A value needed only by an exec credential plugin is forwarded to every profile, including package managers and container engines, defeating the per-command containment that this allowlist exists to provide.

## Minor

- None.

## Questions for the author

- This adds an operator-facing configuration key, but the PR says documentation was not updated and I could not find a linked `solo-docs` PR. Where will users find the YAML syntax, scope, and security limitations?
- The info-level diagnostic emits arbitrary parent-environment names. Can we limit the logged representation to printable, bounded identifier names so control characters or secret-bearing names cannot forge or bloat `solo.log`?

## Positive

- The EKS IRSA fix is narrowly scoped to the `helm` and `kubectl` profiles and has focused regression coverage.
- Rejecting `subprocess.*` environment configuration correctly avoids allowing the environment filter to be disabled by the environment it filters.

## Companion work

- solo-docs PR: missing
- solo-containers PR: N/A

src/business/utils/file-permissions.ts
```diff
@@ -38,6 +38,284 @@ export class FilePermissions {
...
+   * for broad principals, so applying the POSIX-style walk there rejects every ordinary
+   * installation. Checking `C:\` is also of little value — anyone able to write it already
+   * controls the machine. This is a narrower guarantee than POSIX, and is stated as such.
+   *
+   * @param startDirectory - the directory to begin from; it must already exist
+   * @returns a human-readable reason a directory is untrusted, or `undefined` if all are acceptable
+   */
+  public static findUntrustedAncestorReason(startDirectory: string): string | undefined {
+    let current: string = PathEx.realPathSync(PathEx.resolve(startDirectory));
+    for (;;) {
```
jan-milenkov: use a `while()` loop if this is the intent with this (ex. `while(true) {`

JeffreyDallas: udpated

jan-milenkov (commented):
Comment blocks are way too big and unreadable, we should make sure that information is not hidden inside giant blocks on noisy text. Attempt to simplify the PR. Cleanup the description is too large

jan-milenkov (approved):

jan-milenkov (dismissed):

Ivo-Yankov (dismissed):

PR #5911, opened 2026-08-27
chore: OpenSSF — **Token-Permissions** improvements

### LINKED ISSUE #5904: OpenSSF — **Token-Permissions** improvements

jan-milenkov:
## Description

Moved workflow write permissions from whole files down to single jobs. Scorecard Token-Permissions check goes from 0 to 10.

### Related Issues

* Closes # https://github.com/hiero-ledger/solo/issues/5904


.github/workflows/flow-build-application.yaml
```diff
@@ -95,6 +106,9 @@ jobs:
   e2e-tests:
     name: E2E Tests (${{ matrix.e2e-test-type.name }})
     if: ${{ (github.event_name == 'push' || github.event.inputs.enable-e2e-tests == 'true') && !cancelled() }}
+    permissions:
+      contents: read
+      actions: read # the cancellation poller reads the current workflow run
```
JeffreyDallas: the cancellation poller is in `flow-performance-test.yaml`, not `zxc-e2e-test.yaml` — I enumerated every step in `zxc-e2e-test.yaml` and none of them touches the Actions API (no `gha-jobid-action`, no `GITHUB_RUN_ID` poll, no `gh run`). So this comment points at the wrong workflow, and the grant itself is only required because `zxc-e2e-test.yaml` still requests `actions: read` at its top level.

```suggestion
```

This has to be applied together with dropping `actions: read` from `zxc-e2e-test.yaml` (see my comment there) — if the callee keeps requesting it and the caller stops granting it, the whole workflow fails validation with `The workflow is requesting 'actions: read', but is only allowed 'actions: none'`.

Same applies to the `code-style` job on line 86: `zxc-code-style.yaml` only checks out, runs `task check`, and diffs eslint warning counts against the base branch — no Actions API either.

.github/workflows/zxc-e2e-test.yaml
```diff
@@ -221,12 +221,8 @@ defaults:
     shell: bash
 
 permissions:
-  id-token: write
   contents: read
-  actions: write
-  pull-requests: write
-  checks: write
-  statuses: write
+  actions: read
```
JeffreyDallas: nothing in this workflow reads the Actions API — the steps are harden-runner, checkout, setup-node/kubectl/helm/kind/task/go/java/gradle, the e2e run, and three `upload-artifact` calls. `upload-artifact` uses the runtime token, not `GITHUB_TOKEN`, so this can go to zero:

```suggestion
```

Dropping it here is what lets the caller grants in `flow-build-application.yaml` and `flow-pull-request-checks.yaml` shrink to `contents: read`. Change both sides in the same commit.

.github/workflows/flow-deploy-release-artifact.yaml
```diff
@@ -213,6 +208,8 @@ jobs:
   create-github-release:
     name: 'Create GitHub Release'
     runs-on: hiero-solo-linux-medium
+    permissions:
+      contents: write # semantic-release pushes tags/commits and gh release upload attaches assets
```
JeffreyDallas: is this grant actually load-bearing? Everything in this job authenticates with the PAT, not `GITHUB_TOKEN`:

- `Prepare Job` → `checkout-token: '${{ secrets.GH_ACCESS_TOKEN }}'`, so the persisted push credential is the PAT
- `Publish Semantic Release` → `GH_TOKEN`/`GITHUB_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}`
- `Upload Example Asset Archives to Release` → `GH_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}`

If that's right, this can be `contents: read` and the comment should say so, since the whole point of the PR is to not hand out write scopes that nothing consumes. If you'd rather keep it as a belt-and-braces guard against a future step that does use `GITHUB_TOKEN`, could you reword the comment to reflect that instead of naming semantic-release and `gh release upload`?

.github/workflows/flow-deploy-release-artifact.yaml
```diff
@@ -538,6 +535,9 @@ jobs:
   publish-npm-packages:
     name: Publish Release Packages
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write # OIDC token for npm publish provenance
```
JeffreyDallas: this one is required, but not for the reason the comment gives. There is no `NPM_TOKEN` / `NODE_AUTH_TOKEN` anywhere in this file, and `Install NPM latest` runs right before the two `npm publish` calls — so the OIDC token is how the publish *authenticates* (npm trusted publishing); provenance is a side effect. Worth being precise, because "provenance" reads like something you could drop.

```suggestion
      id-token: write # OIDC token for npm trusted publishing (no NPM_TOKEN is configured)
```

.github/workflows/flow-pull-request-checks.yaml
```diff
@@ -36,12 +36,8 @@ on:
       - ready_for_review
 
 permissions:
-  id-token: write
   contents: read
-  actions: write
-  pull-requests: write
-  checks: write
-  statuses: write
+  actions: read
```
JeffreyDallas: this is the only scope added in the PR without a `#` justification. It is load-bearing today — `code-style`, `e2e-tests` and `docs-automation` inherit the top level, and `zxc-code-style.yaml` / `zxc-e2e-test.yaml` both request `actions: read` at their top level, so the call fails validation without it. Please say that here so the next person doesn't delete it:

```suggestion
  actions: read # zxc-code-style.yaml and zxc-e2e-test.yaml request it at their top level
```

If you take the suggestion to drop `actions: read` from those two callees, this line goes away entirely instead.

.github/workflows/flow-nightly-extended-tests.yaml
```diff
@@ -37,8 +37,7 @@ defaults:
 
 permissions:
   contents: read
-  actions: write
-  pull-requests: write
+  actions: read # flow-performance-test.yaml polls the current workflow run
```
JeffreyDallas: the comment is accurate, but this grants `actions: read` to all six jobs in the workflow when only the `flow-performance-test` job needs it. That's the pattern you used correctly in `flow-build-application.yaml` — consider moving it down:

```yaml
  flow-performance-test:
    permissions:
      contents: read
      actions: read # flow-performance-test.yaml polls the current workflow run
    uses: ./.github/workflows/flow-performance-test.yaml
```

and leaving the top level as `contents: read` only. `flow-one-shot-timing.yaml` and `flow-ledger-reset-warm-start-timing.yaml` only request `contents: read`, so they don't need it.

.github/workflows/flow-gcs-test.yaml
```diff
@@ -34,16 +34,19 @@ defaults:
     shell: bash
 
 permissions:
-  id-token: write
   contents: read
-  actions: write
+  actions: read # gha-jobid-action reads the jobs of the current workflow run
```
JeffreyDallas: this top-level grant is dead — `gcs-storage-test` is the only job and it declares its own `permissions` block on line 46, which replaces the workflow-level one wholesale. Worse, this file exposes `workflow_call`, so a top-level request forces every future caller to grant `actions: read` at the job level even though no job would use it.

```suggestion
```

.github/workflows/zxf-stale-issue-pr-closer.yaml
```diff
@@ -11,15 +11,15 @@ defaults:
...
-  contents: write
+  contents: read
 
 jobs:
   stale-pull-requests:
     name: Stale Pull Requests
     runs-on: hiero-solo-linux-medium
+    permissions:
+      contents: write # actions/stale deletes the head branch (delete-branch: true)
+      pull-requests: write # actions/stale labels, comments on and closes pull requests
```
JeffreyDallas, line 22: `actions/stale`'s README asks for `issues: write` in addition to `pull-requests: write`, and label mutation goes through the `POST /repos/{owner}/{repo}/issues/{number}/labels` endpoint. I think you're fine because `days-before-issue-stale: -1` and `days-before-issue-close: -1` mean issues are never processed and that endpoint accepts `pull-requests: write` when the target is a PR — but this workflow only runs on a `15 0 * * *` cron, so if I'm wrong nobody notices for weeks.

did you get a green `workflow_dispatch` run of this on the branch? If so, could you link it?

.github/workflows/zxc-unit-test.yaml
```diff
@@ -39,15 +39,16 @@ defaults:
...
-  id-token: write
   contents: read
   actions: read
-  pull-requests: write
-  checks: write
-  statuses: write
 
 jobs:
   unit-test:
+    permissions:
```
JeffreyDallas, line 47: moving the write scopes into a reusable workflow's job creates a contract that isn't visible from either side: any caller that forgets a job-level grant no longer just loses the comment at runtime, the whole run fails to start with `The workflow is requesting 'checks: write', but is only allowed 'checks: none'`. I checked and both current callers (`flow-build-application.yaml`, `flow-pull-request-checks.yaml`) are covered, but consider a one-line header comment above the block spelling out the requirement, e.g.:

```yaml
    # Callers must grant checks: write and pull-requests: write on the calling job.
```

Applies the same way to `zxc-code-analysis.yaml:82` and `zxc-hugo-publish.yaml:50` — one comment each, no need to repeat the reasoning.

.github/workflows/flow-build-application.yaml
jan-milenkov: updated

.github/workflows/zxc-e2e-test.yaml
jan-milenkov: updated

.github/workflows/flow-deploy-release-artifact.yaml
jan-milenkov: updated

.github/workflows/flow-pull-request-checks.yaml
jan-milenkov: updated — dropped entirely rather than commented, per your note that it goes away if the callees stop requesting `actions: read`.

.github/workflows/flow-nightly-extended-tests.yaml
jan-milenkov: updated

.github/workflows/flow-gcs-test.yaml
jan-milenkov: updated

.github/workflows/zxc-unit-test.yaml
jan-milenkov: updated — applied to `zxc-code-analysis.yaml` and `zxc-hugo-publish.yaml` as well.

.github/workflows/flow-deploy-release-artifact.yaml
jan-milenkov: Confirmed — everything in this job authenticates with `GH_ACCESS_TOKEN`, not `GITHUB_TOKEN`: `checkout-token` L231, semantic-release L320-321, `gh release upload` L360 and L386. The only `GITHUB_TOKEN` in the job is `repo-token` for `setup-task`, which is a read. So the grant was dead.

Dropped to `contents: read` and reworded the comment to say why.

Worth flagging that this is the one change in the PR that isn't verifiable statically, and it sits on the release path. Doing a `dry-run-enabled: true` dispatch before merge — if `create-github-release` fails on a credential I'll revert this line to `contents: write` and keep only the corrected comment.

.github/workflows/zxf-stale-issue-pr-closer.yaml
jan-milenkov: No dispatch run, so I didn't want to leave it unverified on a daily cron. Put `issues: write` back at job level instead.

Job-level writes carry no Scorecard penalty, so this costs nothing against the check and removes the risk of the workflow failing silently for weeks if your reading is right.

If you'd rather have the answer than the guard, I can dispatch it on the branch and drop the scope if it comes back green.

.github/workflows/zxc-unit-test.yaml
```diff
@@ -39,15 +39,17 @@ defaults:
     shell: bash
 
 permissions:
-  id-token: write
   contents: read
   actions: read
-  pull-requests: write
-  checks: write
-  statuses: write
```
JeffreyDallas, line 43: Reminder from the re-review — the only thing left, and it's a cleanup rather than a break.

`actions: read` is dead in this workflow. Nothing here touches the Actions API:

- `actions/upload-artifact@v7` (x3) authenticates with `ACTIONS_RUNTIME_TOKEN`, not `GITHUB_TOKEN`
- `arduino/setup-task`'s `repo-token` is a **contents** read (it resolves a release asset)
- `step-security/publish-unit-test-result-action` asks for `checks: write` + `pull-requests: write`, and `contents: read` / `issues: read` on **private** repos only

This is the same argument you already accepted for `zxc-e2e-test.yaml`, just not carried through to its neighbours. The full set is 8 grants: here (top-level + job), `zxc-code-analysis.yaml` (top-level + job), and the 5 caller-side grants in `flow-pull-request-checks.yaml` and `flow-build-application.yaml`.

Separately, the top-level block itself can go. It's a strict subset of the job-level one, so it adds nothing — and dropping it is the pattern you applied to `flow-gcs-test.yaml` for exactly this reason (a `workflow_call` workflow shouldn't impose a caller contract it doesn't need). Suggestion removes the whole block:

```suggestion
```

.github/workflows/zxc-unit-test.yaml
```diff
@@ -39,15 +39,17 @@ defaults:
...
 
 jobs:
   unit-test:
+    # Callers must grant these scopes on the calling job; otherwise the run fails to start.
+    permissions:
+      contents: read
+      actions: read
```
JeffreyDallas: Dead grant — see the note on the top-level block above. Nothing in this job reads the Actions API.

```suggestion
```

.github/workflows/zxc-code-analysis.yaml
```diff
@@ -74,14 +74,17 @@ defaults:
 permissions:
   contents: read
   actions: read
-  pull-requests: write
-  checks: write
-  statuses: write
```
JeffreyDallas, line 76: Same as `zxc-unit-test.yaml`: `actions: read` is dead here, and the top-level block is a strict subset of the job-level one.

`actions/download-artifact@v8` uses the runtime token for **same-run** downloads; `actions: read` is required only when you pass `run-id` to reach across runs, which none of these three steps do. Codecov uses `CODECOV_TOKEN` (not OIDC), Codacy uses `CODACY_PROJECT_TOKEN`, Snyk uses `SNYK_TOKEN`, and there's no SARIF upload.

```suggestion
```

.github/workflows/zxc-code-analysis.yaml
```diff
@@ -74,14 +74,17 @@ defaults:
...
 
 jobs:
   analyze:
     name: ${{ inputs.custom-job-label || 'Analyze' }}
     runs-on: hiero-solo-linux-medium
+    # Callers must grant these scopes on the calling job; otherwise the run fails to start.
+    permissions:
+      contents: read
+      actions: read
```
JeffreyDallas: Dead grant — see above.

```suggestion
```

.github/workflows/flow-pull-request-checks.yaml
```diff
@@ -98,6 +93,11 @@ jobs:
 
   unit-tests:
     name: Unit Tests
+    permissions:
+      contents: read
+      actions: read
```
JeffreyDallas: Caller side of the same finding — droppable once `zxc-unit-test.yaml` stops requesting `actions: read`.

```suggestion
```

.github/workflows/flow-pull-request-checks.yaml
```diff
@@ -141,6 +141,11 @@ jobs:
 
   codecov:
     name: CodeCov
+    permissions:
+      contents: read
+      actions: read
```
JeffreyDallas: Caller side — droppable once `zxc-code-analysis.yaml` stops requesting `actions: read`.

```suggestion
```

.github/workflows/flow-pull-request-checks.yaml
```diff
@@ -157,6 +162,11 @@ jobs:
 
   codacy-coverage:
     name: Codacy
+    permissions:
+      contents: read
+      actions: read
```
JeffreyDallas: Caller side — droppable once `zxc-code-analysis.yaml` stops requesting `actions: read`.

```suggestion
```

.github/workflows/flow-build-application.yaml
```diff
@@ -84,6 +87,11 @@ jobs:
 
   unit-tests:
     name: Unit Tests
+    permissions:
+      contents: read
+      actions: read
```
JeffreyDallas: Caller side — droppable once `zxc-unit-test.yaml` stops requesting `actions: read`.

```suggestion
```

JeffreyDallas (commented):
Re-reviewed at `9f70dab1`. All nine findings from the previous round are addressed, and I re-verified the risky ones against the source rather than the comments:

- `create-github-release` → `contents: read`: confirmed every write authenticates with `GH_ACCESS_TOKEN` (checkout-token, semantic-release, `gh release upload`). The only `GITHUB_TOKEN` in the job is `setup-task`'s `repo-token`, a read.
- `packages: write` removal is safe — `npm publish` targets registry.npmjs.org, not npm.pkg.github.com.
- `zxc-hugo-publish.yaml`'s `contents: write` **is** load-bearing (`GH_TOKEN=${{ secrets.GITHUB_TOKEN }} task github:upload:release:assets`), so keeping it is correct.
- The `actions: write` removals in `flow-examples-test`, `flow-migration-test`, `flow-one-shot-*` and `flow-ledger-reset-warm-start-timing` are safe — those files have no Actions API consumer at all; the `/actions/runs/` hits are Slack permalink strings.
- Every caller of a reusable workflow that now declares job-level scopes grants them: `zxc-unit-test` (2 callers), `zxc-code-analysis` (3), `zxc-hugo-publish` (3). A green check suite on this commit is real evidence the contract validates.

One cleanup left, inline: `actions: read` looks dead in `zxc-unit-test.yaml` and `zxc-code-analysis.yaml` and in the 5 caller grants that exist only to satisfy them — the same reasoning you applied to `zxc-e2e-test.yaml`. Not blocking.

**Residual risk worth naming:** CI exercises the PR-checks, GCS and performance paths. It does *not* exercise `flow-deploy-release-artifact` (release-only), `flow-hugo-publish`'s publish jobs (`release/*` only), the two cron workflows, or the nightly. I read those through and the grants line up, but the first real release is where a mistake would surface.

.github/workflows/flow-build-application.yaml
```diff
@@ -125,6 +133,11 @@ jobs:
 
   analyze:
     name: Analyze
+    permissions:
+      contents: read
+      actions: read
```
JeffreyDallas: Caller side — droppable once `zxc-code-analysis.yaml` stops requesting `actions: read`.

```suggestion
```

.github/workflows/zxc-unit-test.yaml
jan-milenkov: updated

.github/workflows/zxc-code-analysis.yaml
jan-milenkov: updated

.github/workflows/flow-pull-request-checks.yaml
jan-milenkov: updated

.github/workflows/flow-pull-request-checks.yaml
jan-milenkov: updated

.github/workflows/flow-pull-request-checks.yaml
jan-milenkov: updated

.github/workflows/flow-build-application.yaml
jan-milenkov: updated

.github/workflows/flow-build-application.yaml
jan-milenkov: updated

.github/workflows/zxc-unit-test.yaml
jan-milenkov: Dropped `actions: read` here and in the other seven places — agreed it's dead. Confirmed nothing in either workflow reaches the Actions API: no `run-id` on any `download-artifact` (same-run downloads use the runtime token), no `gh run`/`gh api`/`actions/runs`, and `setup-task`'s `repo-token` is a contents read.

I did **not** take the second half — removing the whole top-level block. It costs a point on the check this PR exists to raise. Measured with scorecard v5.4.0 on two trees identical except for those two blocks:

- top-level `contents: read` kept → Token-Permissions **10**
- top-level block removed → Token-Permissions **9**

`checks/evaluation/permissions.go` (`updateScoreFromUndeclaredTop`) subtracts 0.5 for a workflow with no top-level block whether or not every job declares its own, so removing it from both reusable workflows costs 1.0.

The `flow-gcs-test.yaml` precedent is actually the same shape as what I've done here — there I removed the redundant `actions: read` from the top level but kept `contents: read`. So the `actions: read` goes, `contents: read` stays.

One related instance I left alone: `flow-standard-runner-one-shot.yaml:61` has a dead `actions: read` too, but it's pre-existing in `main` and not in this PR's diff, so I didn't pull it in. Happy to if you'd rather it went with the rest.

.github/workflows/zxc-code-analysis.yaml
jan-milenkov: Dropped `actions: read` at both levels here — agreed, and your reading of `download-artifact` matches what I found: none of the three steps passes `run-id`, so they're same-run and use the runtime token.

Kept the top-level `contents: read` block, for the reason in the `zxc-unit-test.yaml` thread: removing it measurably drops Token-Permissions from 10 to 9 (0.5 per workflow with an undeclared top level, even when every job declares its own).

JeffreyDallas (approved):

Issue #5903, opened 2026-08-27
Epic: OpenSSF Scorecard improvements


jan-milenkov:
## Description

This Epic is created to track progress towards improving solo's OpenSSF score.
The initial score based on 

- [Source](https://github.com/hiero-ledger/hiero/blob/main/scorecards.md)
- [Original Scoreboard](https://scorecard.dev/viewer/?uri=github.com/hiero-ledger/solo)

## How the OpenSSF Scoreboard works

Scorecard's aggregate is a weighted mean — `sum(score × weight) / sum(weight)` — with weights
Critical 10, High 7.5, Medium 5, Low 2.5. For Solo's 14 checks the denominator is **85**, which
reproduces the published 6.3 exactly. That lets each fix be ranked by the points it actually
delivers rather than by how bad the check looks.

## Priority — failing checks, ranked by value per unit of effort

| # | Check | Now | Weight | Effort | Δ overall if maxed |
| --- | --- | --- | --- | --- | --- |
| 1 | **Token-Permissions** | **0** | High 7.5 | Low — YAML only | **+0.88** |
| 2 | **Fuzzing** | **0** | Medium 5 | Low — add `fast-check` tests | **+0.59** |
| 3 | SAST | 5 | Medium 5 | Low — add a CodeQL workflow | +0.29 |
| 4 | Pinned-Dependencies | 8 | Medium 5 | Low — pin ~8 commands | +0.12 |
| 5 | **Signed-Releases** | **0** | High 7.5 | Medium — release pipeline | **+0.88** |
| 6 | Branch-Protection | 4 | High 7.5 | Low edit, real process change | +0.53 |
| 7 | Code-Review | 8 | High 7.5 | Process only | +0.18 |
| 8 | CII-Best-Practices | 2 | Low 2.5 | ~180 questionnaire answers | +0.09 → +0.24 |

PR #5912, opened 2026-08-27
test: add fast-check property-based tests and pin the grpcurl install

### LINKED ISSUE #5905: OpenSSF — **Fuzzing** improvements

mgarbs:
## Description

* Adds two property-based test suites built on [`fast-check`](https://github.com/dubzzz/fast-check), exercising pure parsing/encoding code with randomised input:
  * `test/unit/business/utils/semantic-version.property.test.ts` – `SemanticVersion` round-trips every well-formed version string, tolerates the `v` prefix and surrounding whitespace, compares/orders versions consistently and transitively, bumps to strictly greater versions, and only ever fails with `IllegalArgumentError` on arbitrary or corrupted input.
  * `test/unit/core/helpers-ipv4.property.test.ts` – `isIpV4Address`, `parseIpAddressToUint8Array` and `ipV4ToBase64` agree on every dotted quad, reject out-of-range octets and wrong part counts, and encode every accepted address into four decodable bytes.
* Adds `fast-check` `4.9.0` as a devDependency. Its only transitive dependency is `pure-rand` `8.4.2`. The `overrides` block was refreshed with `scripts/update-production-overrides.mjs` (two new entries) and `scripts/check-production-overrides.mjs` passes.
* Pins the `go install github.com/fullstorydev/grpcurl/cmd/grpcurl@latest` step in `flow-examples-test.yaml` and `zxc-e2e-test.yaml` to `v1.9.3` (the current release), so the tool version used by the E2E and example test runs is reproducible.

Motivation is the OpenSSF Scorecard work tracked in #5903. The Scorecard **Fuzzing** check recognises `fast-check` property tests in TypeScript sources, and **Pinned-Dependencies** counts pinned `go install` commands. This complements #5911 (Token-Permissions) and #5899 (CodeQL / SAST) and does not overlap with either: the only files shared with #5911 are the two workflows above, in different hunks.

| Check | Before | After (local `scorecard --local`) |
|---|---|---|
| Fuzzing | 0 | 10 |
| Pinned-Dependencies | 8 (go commands 0/2 pinned) | 8 (go commands 2/2 pinned) |

The Pinned-Dependencies score does not move because the remaining findings are global npm tool installs (`npm install -g …`, `npm i @hiero-ledger/solo@latest`) and two `curl … | bash` installers, which Scorecard never counts as pinned; those were intentionally left alone.

One observation from the fuzz run, left for a follow-up rather than changed here: build metadata containing a hyphen on a version that has no pre-release (for example `1.0.0+build-1`) is parsed by `SemanticVersion` as if it had a pre-release, so it does not round-trip. The build-metadata generator excludes hyphens for that reason (commented in the test); this is now tracked in #5932.

### Related Issues

* Closes #5905
* Related to #5903

* \[x] Any technical debt has been documented as a separate issue and linked to this PR – the `SemanticVersion` build-metadata parsing bug is tracked in #5932
* \[ ] Any `package.json` changes have been explained to and approved by a repository manager – explained above (`fast-check` devDependency plus the refreshed overrides); approval from a repository manager is requested as part of this review

### Testing

* `npx mocha test/unit/business/utils/semantic-version.property.test.ts test/unit/core/helpers-ipv4.property.test.ts` – 10 passing; also run with `fc.configureGlobal({numRuns: 5000})` to check for flaky properties (all passing).
* `npx mocha test/unit/business/utils/semantic-version.test.ts test/unit/core/helpers.test.ts` – 123 passing (sibling suites unaffected).
* `npx eslint` and `npx prettier --check` on the two new files – clean; `npx tsc --noEmit` – clean.
* `node scripts/check-production-overrides.mjs` – all 902 overrides current.
* `actionlint` on the two changed workflows – no new findings; `scorecard --local --checks Fuzzing,Pinned-Dependencies,Dangerous-Workflow,Binary-Artifacts` – Fuzzing 10, Dangerous-Workflow 10, Binary-Artifacts 10.

The following was not tested:

* The E2E / example test workflows were not executed locally; the only change there is the grpcurl version reference.

### DIFF
--- a/test/unit/business/utils/semantic-version.property.test.ts
+++ b/test/unit/business/utils/semantic-version.property.test.ts
@@ -33,7 +33,7 @@ const preReleaseArbitrary: fc.Arbitrary<string> = fc
   .map((parts: string[]): string => parts.join('.'));
 
 // Build metadata identifiers are kept to letters and digits: a hyphen inside the build metadata of a version
-// without a pre-release is currently read as the start of a pre-release, so it would not round-trip.
+// without a pre-release is read as the start of a pre-release, so it would not round-trip (tracked in #5932).
 const buildMetadataArbitrary: fc.Arbitrary<string> = fc
   .array(fc.string({unit: fc.oneof(digitCharacter, letterCharacter), minLength: 1, maxLength: 8}), {
     minLength: 1,
--- a/test/unit/business/utils/semantic-version.property.test.ts
+++ b/test/unit/business/utils/semantic-version.property.test.ts
@@ -1,0 +1,161 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import {expect} from 'chai';
+import fc from 'fast-check';
+import {SemanticVersion} from '../../../../src/business/utils/semantic-version.js';
+import {IllegalArgumentError} from '../../../../src/core/errors/classes/validation/illegal-argument-error.js';
+
+interface VersionParts {
+  major: number;
+  minor: number;
+  patch: number;
+  preRelease: string | undefined;
+  buildMetadata: string | undefined;
+}
+
+const digitCharacter: fc.Arbitrary<string> = fc.constantFrom(...'0123456789');
+const letterCharacter: fc.Arbitrary<string> = fc.constantFrom(
+  ...'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ',
+);
+const letterOrHyphenCharacter: fc.Arbitrary<string> = fc.oneof(letterCharacter, fc.constant('-'));
+const identifierCharacter: fc.Arbitrary<string> = fc.oneof(digitCharacter, letterOrHyphenCharacter);
+
+// A numeric identifier without leading zeros, as required by the SemVer specification.
+const numericIdentifier: fc.Arbitrary<string> = fc.nat({max: 9999}).map((value: number): string => `${value}`);
+
+// An alphanumeric identifier that starts with a letter or hyphen, so it can never be read as a number.
+const alphanumericIdentifier: fc.Arbitrary<string> = fc
+  .tuple(letterOrHyphenCharacter, fc.string({unit: identifierCharacter, maxLength: 7}))
+  .map(([first, rest]: [string, string]): string => `${first}${rest}`);
+
+const preReleaseArbitrary: fc.Arbitrary<string> = fc
+  .array(fc.oneof(numericIdentifier, alphanumericIdentifier), {minLength: 1, maxLength: 3})
+  .map((parts: string[]): string => parts.join('.'));
+
+// Build metadata identifiers are kept to letters and digits: a hyphen inside the build metadata of a version
+// without a pre-release is currently read as the start of a pre-release, so it would not round-trip.
+const buildMetadataArbitrary: fc.Arbitrary<string> = fc
+  .array(fc.string({unit: fc.oneof(digitCharacter, letterCharacter), minLength: 1, maxLength: 8}), {
+    minLength: 1,
+    maxLength: 3,
+  })
+  .map((parts: string[]): string => parts.join('.'));
+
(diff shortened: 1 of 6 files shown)


test/unit/core/helpers-ipv4.property.test.ts
```diff
@@ -0,0 +1,80 @@
...
+      fc.property(wrongLengthArbitrary, (parts: number[]): void => {
+        const address: string = toDottedQuad(parts);
+
+        expect(isIpV4Address(address)).to.be.false;
+        expect((): string => ipV4ToBase64(address)).to.throw(Error, 'Invalid IPv4 address');
+      }),
+    );
+  });
+
+  it('encodes every string it accepts as an address into four decodable bytes', (): void => {
```
JeffreyDallas: I probed `isIpV4Address` against 100,000 `fc.string()` samples locally and got **0 hits** — the regex needs an exact 4-group dotted-quad of digits, which generic unicode strings essentially never produce. With the default 100 runs, this property almost always just executes the early `return` on line 71 and never reaches the assertions on lines 75-76, so it reads as coverage but rarely exercises anything.

Suggest biasing the generator toward near-valid addresses (e.g., mutate/corrupt strings built from the `octetsArbitrary` used above) instead of relying on `fc.string()` to stumble into a match, or drop the test if this path is already covered by the first property.

test/unit/business/utils/semantic-version.property.test.ts
```diff
@@ -0,0 +1,161 @@
...
+// An alphanumeric identifier that starts with a letter or hyphen, so it can never be read as a number.
+const alphanumericIdentifier: fc.Arbitrary<string> = fc
+  .tuple(letterOrHyphenCharacter, fc.string({unit: identifierCharacter, maxLength: 7}))
+  .map(([first, rest]: [string, string]): string => `${first}${rest}`);
+
+const preReleaseArbitrary: fc.Arbitrary<string> = fc
+  .array(fc.oneof(numericIdentifier, alphanumericIdentifier), {minLength: 1, maxLength: 3})
+  .map((parts: string[]): string => parts.join('.'));
+
+// Build metadata identifiers are kept to letters and digits: a hyphen inside the build metadata of a version
```
JeffreyDallas, line 35: This documents a real parsing bug in production `SemanticVersion` code (build metadata with a hyphen and no pre-release is misread as a pre-release) but only as a comment the generator works around, not a tracked issue. The PR checklist leaves "technical debt documented as a separate issue" unchecked with an offer to file one — I'd take that offer rather than let a `currently` comment be the only record of it (comments anchored to "now"/"currently" tend to decay — see PR #4363 precedent).

test/unit/core/helpers-ipv4.property.test.ts
mgarbs: Thanks for probing this — you are right, that property was almost pure early-return. I replaced it in 3c69ffc with a weighted generator that starts from a valid dotted quad and applies one targeted corruption per case: an oversized octet (256–999), a zero-padded octet (two-digit `01` is accepted by the regex, three-digit `001` is not — both are asserted), three or five groups, an empty group, an injected non-address character, or leading/trailing garbage, each with a decidable expected result asserted against `isIpV4Address`; a small `fc.string()` share remains as noise checked only for accept-then-encode consistency. Sampling 100,000 inputs: ~93% now carry a known expected classification and ~21% are accepted addresses, versus the 0 matches you measured before. Still green with `numRuns: 5000`.

test/unit/business/utils/semantic-version.property.test.ts
mgarbs: Fair point — a `currently` comment is a poor system of record. Filed #5932 with a repro (`1.2.3+build-1` parses to `preRelease: "1"` / `buildMetadata: "build-1"` and compares below plain `1.2.3`, where SemVer 2.0.0 says build metadata must be ignored for precedence), the expected behavior, and a pointer to the pre-release regex in `semantic-version.ts`. The generator comment now references the issue and the technical-debt checklist item in the PR body is checked with the link.

PR #5924, opened 2026-08-31
fix: two kind flags miss their -- and process.env is mutated globally

### LINKED ISSUE #5872: bug: two kind flags miss their `--`, and `process.env` is mutated globally
## Description

Two kind options are passed as bare words, so kind treats them as positional arguments and the option
does nothing: `cluster-create-options.ts:66` (`flag('retain')`) and
`export-kubeconfig-options.ts:44` (`flag('internal')`). Correct examples are
`get-kubeconfig-options.ts:37` and `get-nodes-options.ts:37`. Consider making
`KindExecutionBuilder.flag()` reject a value that does not start with `-`.

Separately, a few places mutate `process.env` instead of passing overrides to the subprocess:
`cluster-task-manager.ts:156, 465`, `brew-package-manager.ts:49, 101`,
`podman-dependency-manager.ts:318`. Use `environmentVariablesToAppend` like the other call sites do.

### Related Issues

* Found while auditing # https://github.com/hiero-ledger/solo/issues/4005

Fantomasa:
### Kind flags passed as bare words

`ClusterCreateOptions` passed `retain` and `ExportKubeConfigOptions` passed `internal` without the `--` prefix, so kind treated them as positional arguments and the options did nothing. Both are now passed as `--retain` / `--internal`, and `KindExecutionBuilder.flag()` rejects any value that does not start with `-` (via `SoloErrors.validation.illegalArgument`), so this class of bug fails fast in the future. Neither option is currently set to `true` anywhere in `src/`, so the fix has no behavioral impact on existing flows.

### Global `process.env` mutations

Five places mutated `process.env` to influence later subprocess spawns (`cluster-task-manager.ts`, `brew-package-manager.ts`, `podman-dependency-manager.ts`). These mutations were load-bearing: the values flow to child processes through the `SubprocessEnvironment` per-profile allowlists, and `KIND_EXPERIMENTAL_PROVIDER` is additionally read back in-process by `PodmanClient` / `DockerClient` to decide which container engine backs kind.

Instead of threading overrides through every call path, `SubprocessEnvironment` now owns explicit session-scoped state:

* `setSessionVariable()` / `sessionVariable()` — runtime-established variables, merged over the parent environment **before** allowlist filtering, so they behave exactly like inherited variables without the global mutation.
* `prependSessionPath()` / `appendSessionPath()` / `currentPath()` — PATH additions (brew bin directories, brew-installed podman), applied to the PATH of every spawned command and available to the call sites that build explicit `sudo env PATH=...` prefixes.

All five mutation sites now register session state instead. The in-process reads were consolidated into `PodmanClient.kindProvider()`, which prefers the session value and falls back to a user-provided `KIND_EXPERIMENTAL_PROVIDER` environment variable, preserving existing behavior for users who set it themselves. PATH readers that fed spawned commands (`cluster-task-manager`, `podman-client`, `default-kind-client`) now use `SubprocessEnvironment.currentPath()` so they keep seeing the session additions.

### Testing

* New unit tests for the session state (allowlist filtering, precedence, PATH wrapping, dedupe, override precedence).
* New `BrewPackageManager` unit test covering the `brew shellenv` parsing into session state.
* New `setupConfig` virtual-machine-mode test asserting `CONTAINERS_CONF` lands in session state and `process.env` stays untouched.
* New `PodmanClient.kindProvider()` precedence tests, plus the flag-guard test; existing tests that locked in the bare `retain` flag updated to `--retain`.

### Related Issues

Closes #5872

### DIFF
--- a/src/core/package-managers/brew-package-manager.ts
+++ b/src/core/package-managers/brew-package-manager.ts
@@ -100,22 +100,22 @@ export class BrewPackageManager extends ShellRunner implements PackageManager {
         continue;
       }
       const [, key, rawValue]: string[] = match;
-      const value: string = BrewPackageManager.expandShellValue(rawValue);
       if (key === 'PATH') {
-        BrewPackageManager.registerPathAdditions(value);
+        BrewPackageManager.registerPathAdditions(rawValue);
       } else {
-        SubprocessEnvironment.setSessionVariable(key, value);
+        SubprocessEnvironment.setSessionVariable(key, BrewPackageManager.expandShellValue(rawValue));
       }
     }
   }
 
   /**
-   * Registers the new directories from a shellenv `PATH` value as session path prepends, preserving
-   * order. The brew value is split on the `:` brew hardcodes; this flow only runs on Linux.
+   * Registers the new directories from a raw shellenv `PATH` value as session path prepends, preserving order.
+   * References to the existing PATH are dropped, never expanded and re-split on the `:` brew hardcodes.
    */
-  private static registerPathAdditions(pathValue: string): void {
+  private static registerPathAdditions(rawPathValue: string): void {
+    const literalValue: string = rawPathValue.replaceAll(/\$(\{PATH[^}]*\}|PATH\b)/g, '');
     const currentSegments: Set<string> = new Set<string>(SubprocessEnvironment.currentPath().split(PathEx.delimiter));
-    const newSegments: string[] = pathValue
+    const newSegments: string[] = BrewPackageManager.expandShellValue(literalValue)
       .split(':')
       .filter((segment: string): boolean => Boolean(segment) && !currentSegments.has(segment));
     // prependSessionPath puts each directory in front, so iterate in reverse to preserve order.
--- a/src/core/package-managers/brew-package-manager.ts
+++ b/src/core/package-managers/brew-package-manager.ts
@@ -110,12 +110,11 @@ export class BrewPackageManager extends ShellRunner implements PackageManager {
   }
 
   /**
-   * Registers the new directories from a shellenv `PATH` value (the brew bin/sbin directories it
-   * prepends to the inherited `PATH`) as session path prepends, preserving their order. brew
-   * hardcodes `:` as the separator, and this flow only runs on Linux.
+   * Registers the new directories from a shellenv `PATH` value as session path prepends, preserving
+   * order. The brew value is split on the `:` brew hardcodes; this flow only runs on Linux.
    */
   private static registerPathAdditions(pathValue: string): void {
-    const currentSegments: Set<string> = new Set<string>(SubprocessEnvironment.currentPath().split(':'));
+    const currentSegments: Set<string> = new Set<string>(SubprocessEnvironment.currentPath().split(PathEx.delimiter));
     const newSegments: string[] = pathValue
       .split(':')
       .filter((segment: string): boolean => Boolean(segment) && !currentSegments.has(segment));
--- a/src/core/package-managers/brew-package-manager.ts
+++ b/src/core/package-managers/brew-package-manager.ts
@@ -4,6 +4,7 @@ import fs from 'node:fs';
 import os from 'node:os';
 import {ShellRunner} from '../shell-runner.js';
 import {SubprocessCommandProfile} from '../subprocess-command-profile.js';
+import {SubprocessEnvironment} from '../subprocess-environment.js';
 import {type PackageManager} from './package-manager.js';
 import {PathEx} from '../../business/utils/path-ex.js';
 import {getEnvironmentVariable} from '../constants.js';
@@ -46,7 +47,7 @@ export class BrewPackageManager extends ShellRunner implements PackageManager {
   public async install(): Promise<boolean> {
     await this.runHomebrewScript(BrewPackageManager.INSTALL_SCRIPT_URL);
     await this.applyShellEnvironment();
-    process.env.PATH = `${process.env.PATH}${PathEx.delimiter}${BrewPackageManager.LINUXBREW_BIN}`;
+    SubprocessEnvironment.appendSessionPath(BrewPackageManager.LINUXBREW_BIN);
     return this.isAvailable();
   }
 
@@ -86,6 +87,8 @@ export class BrewPackageManager extends ShellRunner implements PackageManager {
   /**
    * Applies the environment that `brew shellenv` would export, without a shell `eval`. Parses the
    * `export KEY="VALUE";` lines and expands the shell parameter references each value contains.
+   * The values are registered as session environment state for subsequent spawned commands
+   * instead of mutating the global `process.env`.
    */
   private async applyShellEnvironment(): Promise<void> {
     const output: string[] = await this.run(`${BrewPackageManager.LINUXBREW_BIN}/brew`, ['shellenv'], {
@@ -97,8 +100,29 @@ export class BrewPackageManager extends ShellRunner implements PackageManager {
         continue;
       }
       const [, key, rawValue]: string[] = match;
-      // eslint-disable-next-line no-restricted-syntax
-      process.env[key] = BrewPackageManager.expandShellValue(rawValue);
+      const value: string = BrewPackageManager.expandShellValue(rawValue);
+      if (key === 'PATH') {
+        BrewPackageManager.registerPathAdditions(value);
+      } else {
+        SubprocessEnvironment.setSessionVariable(key, value);
+      }
+    }
+  }
+
+  /**
+   * Registers the new directories from a shellenv `PATH` value (the brew bin/sbin directories it
+   * prepends to the inherited `PATH`) as session path prepends, preserving their order. brew
+   * hardcodes `:` as the separator, and this flow only runs on Linux.
+   */
+  private static registerPathAdditions(pathValue: string): void {
+    const currentSegments: Set<string> = new Set<string>(SubprocessEnvironment.currentPath().split(':'));
+    const newSegments: string[] = pathValue
+      .split(':')
+      .filter((segment: string): boolean => Boolean(segment) && !currentSegments.has(segment));
+    // prependSessionPath puts each directory in front, so iterate in reverse to preserve order.
+    newSegments.reverse();
+    for (const segment of newSegments) {
+      SubprocessEnvironment.prependSessionPath(segment);
     }
   }
 
--- a/src/core/subprocess-environment.ts
+++ b/src/core/subprocess-environment.ts
@@ -3,6 +3,7 @@
 import os from 'node:os';
 import path from 'node:path';
 import {SubprocessCommandProfile} from './subprocess-command-profile.js';
+import {SoloErrors} from './errors/solo-errors.js';
 
 /**
  * Builds a minimal, explicit environment for spawning external commands.
@@ -15,7 +16,7 @@ import {SubprocessCommandProfile} from './subprocess-command-profile.js';
  *
  * Intentionally dependency-free (no dependency-injection, no logging, no heavy imports) so
  * that the standalone `persist-port-forward` script can import it without pulling in the
- * container.
+ * container. The registered error classes are the one exception; they have no runtime dependencies.
  */
 export class SubprocessEnvironment {
   /** Environment variable names inherited by every external command on every platform. */
@@ -156,6 +157,12 @@ export class SubprocessEnvironment {
    * @param value - the value to hand to child processes
    */
   public static setSessionVariable(name: string, value: string): void {
+    if (name.toLowerCase() === 'path') {
+      throw new SoloErrors.validation.illegalArgument(
+        'PATH must be registered with prependSessionPath/appendSessionPath',
+        name,
+      );
+    }
     SubprocessEnvironment.sessionVariables.set(name, value);
   }
 
@@ -205,11 +212,16 @@ export class SubprocessEnvironment {
     return SubprocessEnvironment.withSessionPathAdditions(base);
   }
 
-  /** Clears every session-scoped variable and path addition. Intended for tests. */
-  public static clearSessionState(): void {
+  /** Clears every session variable and path addition. Session state is write-once per invocation; only tests may call this. */
+  public static resetForTesting(): void {
     SubprocessEnvironment.sessionVariables.clear();
     SubprocessEnvironment.sessionPathPrepends.length = 0;
     SubprocessEnvironment.sessionPathAppends.length = 0;
+  }
+
+  /** The environment's `PATH` key, matched case-insensitively (POSIX `PATH`, Windows `Path`); `PATH` when absent. */
+  public static pathKey(environment: Record<string, string>): string {
+    return Object.keys(environment).find((key: string): boolean => key.toLowerCase() === 'path') ?? 'PATH';
   }
 
   private static isSessionPathRegistered(directory: string): boolean {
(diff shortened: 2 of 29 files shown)


src/core/subprocess-environment.ts
```diff
@@ -162,7 +259,12 @@ export class SubprocessEnvironment {
     );
 
     const environment: Record<string, string> = {};
-    for (const [name, value] of Object.entries(process.env)) {
+    // Session variables overlay the parent environment before filtering, so the allowlists apply to both.
+    const sourceEntries: Array<[string, string | undefined]> = [
```
JeffreyDallas: **This is one of two merge conflicts with #5897** (`fix-5895-subprocess-env-denylist`), which rewrites this same method to collect and log the names it filtered out. I test-merged the branches — the unit test file auto-merges, this file does not, and the semantic result of a naive resolution is wrong in two ways:

1. **Session variables get reported as "withheld".** #5897 pushes every non-allowlisted name into `withheldNames` and logs it so a user can answer "was my variable dropped?" from `solo.log`. With session variables appended to `sourceEntries`, `CONTAINERS_CONF` — set as session state once `PodmanDependencyManager.setupConfig()` runs — would be logged as withheld *from the parent environment* for the `HELM`, `NPM` and `BREW` profiles, where it never was. The test you added in this PR asserts `forCommand(HELM)` has no `CONTAINERS_CONF`, so this fires on a normal podman run.

2. **Duplicates.** A name in both `process.env` and `sessionVariables` that is not allowlisted for the profile gets pushed to `withheldNames` twice — e.g. a user who exports `KIND_EXPERIMENTAL_PROVIDER` themselves, which this PR also sets as session state.

Whichever PR lands second, the resolution needs session variables either excluded from `withheldNames` or reported under a distinct label. Worth agreeing with @JeffreyDallas on ordering first — #5897 is at `CHANGES_REQUESTED` and ~6x the size, so landing this one first and rebasing that one is probably cheaper.

src/core/subprocess-environment.ts
```diff
@@ -125,11 +126,107 @@ export class SubprocessEnvironment {
...
+      SubprocessEnvironment.sessionPathAppends.push(directory);
+    }
+  }
+
+  /**
+   * The `PATH` child processes resolve binaries against: the parent `PATH` wrapped with the
+   * session path additions. For callers that construct explicit environment prefixes
+   * (`sudo env PATH=...`) or `PATH` overrides instead of relying on {@link forCommand}.
+   */
+  public static currentPath(): string {
```
JeffreyDallas, line 210: `currentPath()` covers the callers that build an explicit `PATH` for a spawned command, but two **in-process** readers were left on the raw `process.env.PATH`:

- `src/core/dependency-managers/base-dependency-manager.ts:125` — `getGlobalExecutableWithPath()` splits `process.env.PATH` to locate a global binary.
- `src/core/package-managers/os-package-manager.ts:132` — `isCommandAvailable()` does the same to probe for a package-manager binary.

That matters for the flow this PR is changing: `cluster-task-manager.ts:155` used to extend `process.env.PATH` with the brew podman directory so that `PodmanDependencyManager.getExecutable()` → `getGlobalExecutableWithPath()` would find `/home/linuxbrew/.linuxbrew/bin/podman`. With the directory now only in session state, that lookup misses and `isInstalledGloballyAndMeetsRequirements()` reports no compatible global podman, so solo downloads its own bundle instead of using the one it just installed.

Same class, lower impact — these build an explicit `PATH` for a child process and would also miss session additions: `metrics-server-impl.ts:251`, `default-helm-client.ts:194`, `k8-client-container.ts:86`, `k8-client-pod.ts:297`, `persist-port-forward.ts:194`, `commands/node/tasks.ts:3132`.

Note that `persist-port-forward.ts:194` is the sharpest case, and #5897 already hit and solved this exact class: it added `getOperatorAllowlist()` so the port-forward worker — which runs in its own process and therefore starts with empty static state — can be handed the parent's values and re-seed them. Session state needs the same treatment or it is silently empty in that child.

If leaving the two scanners on the real `process.env.PATH` is deliberate, say so in the class doc — otherwise the next "podman not found" bug gets fixed by putting the mutation back.

src/integration/container-engine/podman-client.ts
```diff
@@ -33,6 +34,18 @@ export class PodmanClient {
       .containerConfigEnvironment();
   }
 
+  /**
+   * The kind provider in effect for this solo invocation: the session value set when solo
+   * configures kind to use podman, falling back to a user-provided `KIND_EXPERIMENTAL_PROVIDER`
+   * environment variable.
+   */
+  public static kindProvider(): string | undefined {
```
JeffreyDallas: This method has no podman behavior in it — it is a read of session state with an env fallback — and putting it here means `DockerClient` now asks `PodmanClient` which engine backs kind (`docker-client.ts:187`). Consider moving it next to the state it reads, so neither engine client owns the other's question:

```typescript
// src/core/subprocess-environment.ts
public static kindProvider(): string | undefined {
  return (
    SubprocessEnvironment.sessionVariable('KIND_EXPERIMENTAL_PROVIDER') ??
    getEnvironmentVariable('KIND_EXPERIMENTAL_PROVIDER')
  );
}
```

If the `constants` import is the blocker for keeping `SubprocessEnvironment` dependency-free, a small dedicated resolver class works too — the point is that neither container-engine client should be the authority.

src/integration/kind/execution/kind-execution-builder.ts
```diff
@@ -152,13 +153,17 @@ export class KindExecutionBuilder extends ExecutionBuilder {
...
+   * @param flag the flag to be added; must start with `-` (e.g. `--retain`), because kind
+   *   silently treats a bare word as a positional argument instead of an option
    * @returns this builder
    */
   public flag(flag: string): KindExecutionBuilder {
     if (!flag) {
       throw new Error('flag must not be null');
     }
+    if (!flag.startsWith('-')) {
+      throw new SoloErrors.validation.illegalArgument('flag must start with "-"', flag);
```
JeffreyDallas: Good guard, but it only covers kind. `HelmExecutionBuilder.flag()` (`src/integration/helm/execution/helm-execution-builder.ts:177`) is the identical method with the identical null check and no dash check — same bug class, one `builder.flag('atomic')` away from the same silent no-op.

Both builders extend `ExecutionBuilder`, so the validation (and arguably `flag()`/`_flags` itself) belongs in the base class rather than being duplicated and then fixed in only one copy. Style guide §DRY / checklist: "this seems like a duplicate — enhance the existing one."

src/integration/kind/execution/kind-execution-builder.ts
```diff
@@ -152,13 +153,17 @@ export class KindExecutionBuilder extends ExecutionBuilder {
...
   /**
    * Adds a flag to the kind execution.
-   * @param flag the flag to be added
+   * @param flag the flag to be added; must start with `-` (e.g. `--retain`), because kind
+   *   silently treats a bare word as a positional argument instead of an option
    * @returns this builder
    */
   public flag(flag: string): KindExecutionBuilder {
     if (!flag) {
       throw new Error('flag must not be null');
```
JeffreyDallas:
```suggestion
      throw new SoloErrors.validation.illegalArgument('flag must not be null', flag);
```
While you are adding a `SoloErrors` throw three lines below, this raw `new Error` is the exact pattern CLAUDE.md "Error Handling — Always Use a Registered `SoloErrors` Subclass" bans (no `code`, no `troubleshootingSteps`). `IllegalArgumentError` sets `message: reason`, so the existing `expect(error.message).to.equal('flag must not be null')` assertion in `kind-execution-builder.test.ts` still passes.

src/core/subprocess-environment.ts
```diff
@@ -125,11 +126,107 @@ export class SubprocessEnvironment {
...
+  /**
+   * Registers an environment variable for every subsequently spawned command in this solo
+   * invocation. The variable is still subject to the per-profile allowlists, exactly as if it
+   * were present in the parent environment. `PATH` must not be set this way — use
+   * {@link prependSessionPath} / {@link appendSessionPath} instead.
+   *
+   * @param name - the environment variable name
+   * @param value - the value to hand to child processes
+   */
+  public static setSessionVariable(name: string, value: string): void {
```
JeffreyDallas, line 159: Two things here.

**The documented restriction is not enforced.** The doc says "`PATH` must not be set this way" but nothing stops it — and if someone does, the value silently lands in `environment` and then gets re-wrapped by the path-additions block below, which is confusing to debug. This PR's own argument for the `flag()` guard applies verbatim: make the class of bug fail fast.

```typescript
if (name.toLowerCase() === 'path') {
  throw new SoloErrors.validation.illegalArgument(
    'PATH must be registered with prependSessionPath/appendSessionPath',
    name,
  );
}
```

**Second, related to #5897:** that PR vets operator-supplied variable names against `NEVER_PASSTHROUGH_NAMES` / `NEVER_PASSTHROUGH_PREFIXES` (`LD_*`, `DYLD_*`, `AWS_ENDPOINT_URL*`, …) before they can reach a child. `setSessionVariable()` becomes a second way to put a name into a child environment with no such check. It is not exploitable today — the per-profile allowlist still gates admission and only Solo's own code calls this — but once both land there are two doors with different locks. Worth an explicit decision on whether this should share the refusal check.

(If importing `SoloErrors` would break the "intentionally dependency-free" contract that lets `persist-port-forward` import this file, that is a fair reason to skip the throw — worth a note either way.)

src/core/package-managers/brew-package-manager.ts
```diff
@@ -97,8 +100,29 @@ export class BrewPackageManager extends ShellRunner implements PackageManager {
...
+    }
+  }
+
+  /**
+   * Registers the new directories from a shellenv `PATH` value (the brew bin/sbin directories it
+   * prepends to the inherited `PATH`) as session path prepends, preserving their order. brew
+   * hardcodes `:` as the separator, and this flow only runs on Linux.
+   */
+  private static registerPathAdditions(pathValue: string): void {
+    const currentSegments: Set<string> = new Set<string>(SubprocessEnvironment.currentPath().split(':'));
```
JeffreyDallas: `SubprocessEnvironment.currentPath()` joins with `path.delimiter`, so splitting its output on a hardcoded `':'` is inconsistent with how the string was built. The comment justifies `':'` for *brew's* output, which is fair, but it does not cover the `currentPath()` side — on Windows this turns `C:\Users\...` into bogus segments and the dedupe check silently stops working.

```suggestion
    const currentSegments: Set<string> = new Set<string>(SubprocessEnvironment.currentPath().split(path.delimiter));
```

The `pathValue.split(':')` on the next line can stay as-is with its comment, since that value really does come from brew.

src/core/subprocess-environment.ts
```diff
@@ -125,11 +126,107 @@ export class SubprocessEnvironment {
...
+   * session path additions. For callers that construct explicit environment prefixes
+   * (`sudo env PATH=...`) or `PATH` overrides instead of relying on {@link forCommand}.
+   */
+  public static currentPath(): string {
+    const base: string = process.env.PATH ?? process.env.Path ?? '';
+    return SubprocessEnvironment.withSessionPathAdditions(base);
+  }
+
+  /** Clears every session-scoped variable and path addition. Intended for tests. */
+  public static clearSessionState(): void {
```
JeffreyDallas: A `public` production API whose doc says "Intended for tests" is a smell — nothing stops a command from calling it mid-run and quietly dropping the brew PATH and `CONTAINERS_CONF` that later tasks depend on. Either name it so the intent survives (`resetForTesting()`) or, better, tighten the doc to state the invariant: session state is write-once-per-invocation and this is the only sanctioned reset.

For precedent, #5897 solves the same "static state needs resetting between tests" problem with `resetWithheldReporting()` — worth matching whichever naming survives.

src/core/subprocess-environment.ts
```diff
@@ -176,6 +278,12 @@ export class SubprocessEnvironment {
       }
     }
 
+    if (SubprocessEnvironment.sessionPathPrepends.length > 0 || SubprocessEnvironment.sessionPathAppends.length > 0) {
```
JeffreyDallas, line 295: This case-insensitive PATH-key lookup is the same logic as `ExecutionBuilder.prefixPath()` (`src/integration/execution-builder.ts:8`), down to the `key.toLowerCase() === 'path'` comparison. Second occurrence, so just flagging it rather than insisting — but if `SubprocessEnvironment` now owns PATH assembly for every spawned command, `ExecutionBuilder.prefixPath()` may be the thing that should be collapsing into it.

Also: when the parent environment has no PATH at all, this branch creates a `PATH` containing *only* the session directories where previously the child got none. Probably fine, but worth a conscious decision rather than a fallthrough.

This block is the **second of the two merge conflicts with #5897**, which replaces these same lines with its withheld-name reporting.

jan-milenkov (dismissed):

src/core/subprocess-environment.ts
Fantomasa: Fixed: session variables are now filtered in a separate pass from the parent environment, so a withheld-name loop merges onto the parent pass only, with no duplicates and nothing solo-set reported as withheld.

src/core/subprocess-environment.ts
Fantomasa: Fixed: all eight readers now use SubprocessEnvironment.currentPath(), including both in-process scanners. The port-forward worker needs no re-seeding since the parent bakes the session PATH and allowlisted variables into its environment at spawn.

src/integration/container-engine/podman-client.ts
Fantomasa: Fixed: moved to a new KindProviderResolver, since constants.getEnvironmentVariable is too heavy an import for the dependency-free SubprocessEnvironment.

src/integration/kind/execution/kind-execution-builder.ts
Fantomasa: Fixed: _flags and flag() moved to ExecutionBuilder with both validations, so helm gets the same guard. All existing call sites already pass dashed flags.

src/integration/kind/execution/kind-execution-builder.ts
Fantomasa: Fixed: the null check now throws SoloErrors.validation.illegalArgument, in the shared base-class flag().

src/core/subprocess-environment.ts
Fantomasa: Fixed: setSessionVariable now throws illegalArgument on a case-insensitive PATH match. The error classes have no runtime dependencies, so the import keeps the file safe for persist-port-forward; wiring the NEVER_PASSTHROUGH check belongs to whichever PR lands second.

src/core/package-managers/brew-package-manager.ts
Fantomasa: Fixed: the currentPath() side now splits on the platform delimiter; the brew value keeps its hardcoded colon split.

src/core/subprocess-environment.ts
Fantomasa: Fixed: renamed to resetForTesting() and the doc states the write-once invariant.

src/core/subprocess-environment.ts
Fantomasa: Fixed: the lookup is now SubprocessEnvironment.pathKey(), shared by forCommand() and prefixPath(), and the no-PATH case is an explicit decision in a comment. prefixPath also no longer emits a trailing delimiter for an empty PATH.

jan-milenkov (approved):

Issue #5940, opened 2026-09-01
Add a flag to load a component image from a docker save archive


JeffreyDallas:
## Overview

GitHub Actions jobs do not share a Docker daemon. A realistic CI pipeline built on the actions from
#5021 will build a component in one job and deploy the Solo network in another, handing the image
over as a `docker save` tarball via `upload-artifact` / `download-artifact`. Today Solo has no way to
consume that tarball — `--component-image` only resolves images that are already present in the local
Docker daemon (`BaseCommand.isLocalImageAvailableInDocker`, `src/commands/base.ts:296-302`).

The underlying capability already exists and is unused by the CLI surface: `KindClient.loadImageArchive`
(`src/integration/kind/kind-client.ts:114`, implemented at
`src/integration/kind/impl/default-kind-client.ts:153-161`) wraps `kind load image-archive`, and is
currently only reached from the image cache handler
(`src/integration/cache/impl/image-cache-handler.ts:128`). No CLI flag exposes it.

## Sub-Tasks

- [ ] Add a flag (e.g. `--component-image-archive`) accepting a path to a `docker save` tarball
- [ ] Load the archive into the target cluster(s) via the existing `KindClient.loadImageArchive`
      rather than adding a parallel mechanism
- [ ] Resolve the image reference from the archive (or require it to be given alongside via
      `--component-image`) so the chart values and `pullPolicy: Never` can still be set correctly
- [ ] Support the flag on all four image-consuming components — block node, mirror node, relay,
      explorer — reusing the shared helper on `BaseCommand` rather than duplicating per command
- [ ] Register the flag per the CLAUDE.md "Adding CLI Flags" checklist (`Flags.allFlags` plus the
      per-command flag lists) and run `npx mocha 'test/unit/commands/flags.test.ts'`
- [ ] Document the build-job / deploy-job handoff pattern in the actions documentation

## Prerequisites

- [ ] requires #5939 — the multi-cluster load path should be settled first so the archive
      load follows the same targeting rules

## References

- Parent initiative: #5021
- New actions repo: #5856
- [Hiero TSC proposal](https://github.com/hiero-ledger/tsc/issues/298#issuecomment-5318815798)

Issue #5939, opened 2026-09-01
Make locally built component images available beyond a single kind cluster


JeffreyDallas:
## Overview

Solo can only make a locally built component image available to the cluster via
`kind load docker-image` into a **single** kind cluster —
`BaseCommand.kindLoadComponentImage` (`src/commands/base.ts:304-313`) derives one cluster name from
one context and loads there. Two consequences block the build actions in #5021:

1. **Multi-cluster deployments.** In a dual-cluster topology the image lands in only one kind
   cluster; pods scheduled in the other cluster fail to start.
2. **Non-kind clusters.** For any cluster Solo did not create with kind (remote, GKE, k3s, a shared
   CI cluster), there is no load path at all — the only option is a registry, but
   `BaseCommand.isLocalImageReference` (`src/commands/base.ts:275-284`) classifies anything whose
   first segment contains `.` or `:` or equals `localhost` as remote, so a kind-attached local
   registry reference such as `localhost:5001/block-node-server:0.38.0-abc1234` is never treated as
   locally available and never gets `pullPolicy: Never`.

The build actions need a documented, working answer for "the image exists locally, get it into every
cluster that will schedule this component".

## Sub-Tasks

- [ ] Load the component image into **every** cluster/context that will run the component, not just
      the one derived from a single context
- [ ] Support a kind-attached local registry: allow `localhost:<port>/…` references to be recognised
      as local-registry-backed and set `pullPolicy` accordingly instead of falling through to the
      remote-registry branch
- [ ] Apply the fix consistently across all four image-consuming components — block node
      (`src/commands/block-node.ts:375-396`), mirror node (`src/commands/mirror-node.ts:539-570`),
      relay (`src/commands/relay.ts:283-297`), explorer (`src/commands/explorer.ts:418-437`)
- [ ] Emit a clear, actionable error when the image is not reachable from the target cluster, rather
      than letting the pod fail later with `ErrImageNeverPull`
- [ ] Document the supported paths (kind load vs. local registry vs. published registry) in the
      `--component-image` flag description and the relevant docs page
- [ ] Tests covering the multi-cluster load and the `localhost:<port>` reference form

## Prerequisites

- [ ] Coordinate with #5938 so the multi-image Mirror Node case is loaded correctly once both
      land

## References

- Parent initiative: #5021
- New actions repo: #5856
- [Hiero TSC proposal](https://github.com/hiero-ledger/tsc/issues/298#issuecomment-5318815798)

Issue #5938, opened 2026-09-01
Fix --component-image for Mirror Node: six module images cannot share one repository


JeffreyDallas:
## Overview

`--component-image` cannot express a locally built Mirror Node. Mirror Node's build produces **six
distinct images** — `hedera-mirror-importer`, `-grpc`, `-rest`, `-rest-java`, `-web3`, `-monitor`
(see [`docker-conventions.gradle.kts`](https://github.com/hiero-ledger/hiero-mirror-node/blob/main/buildSrc/src/main/kotlin/docker-conventions.gradle.kts), image name is `${imageRegistry}/hedera-mirror-${project.name}`).
Solo parses the single `--component-image` reference and sets the **same** `image.repository` on all
six module chart keys (`importer`, `grpc`, `rest`, `restjava`, `web3`, `monitor`) —
`src/commands/mirror-node.ts:541-559`.

The result is that five of the six modules are pointed at an image reference that does not exist, so
the deployment fails with `ErrImageNeverPull` (local, `pullPolicy: Never`) or `ImagePullBackOff`.

This blocks the Mirror Node build action from #5021, which must be able to hand Solo a locally built
Mirror Node with a uniform action contract. Without a fix, Mirror Node would be the one component
forced to emit a Helm values file instead of an image reference, breaking the DRY goal of the action
set.

## Sub-Tasks

- [ ] Decide the reference form. Preferred: `--component-image` accepts a **repository prefix** (e.g.
      `hedera-mirror:0.156.0-abc1234` or `myprefix/hedera-mirror:<tag>`) and Solo appends the module
      suffix per chart key, preserving the `rest-java` → `restjava` chart-key mapping
- [ ] Implement the prefix expansion in `MirrorNodeCommand`, replacing the six identical
      `setLiteral('<module>.image.repository', …)` calls at `src/commands/mirror-node.ts:541-559`
- [ ] Extend `kindLoadComponentImage` (`src/commands/base.ts:304`) so all six derived images are
      loaded into the cluster, not just one
- [ ] Keep `pullPolicy: Never` behaviour, applied only to the images actually found in the local
      Docker daemon
- [ ] Preserve backwards compatibility: an explicit single registry reference must keep working as it
      does today
- [ ] Verify the interaction with `addMirrorNodeMemoryOverrides`
      (`src/commands/mirror-node.ts:1998-2020`), which already branches on `hasCustomComponentImage`
[...]

PR #5949, opened 2026-09-02
fix: load local component images into all Kind clusters

JeffreyDallas:
## Description

* Loads locally available component images into every targeted Kind cluster before deployment.
* Supports images served by a Kind-attached local registry (`localhost:<port>/...`) with `pullPolicy: Never`.
* Ensures Explorer uses the shared image-loading path and fails early with actionable guidance when a target context is not a Kind cluster.
* Documents the supported component-image availability paths in CLI help.

### Related Issues

* Related to #5939 (docs.hedera.com page update tracked separately; not closed by this PR)
* Related to #5940
* Related to #5938

* [x] When this PR merges the commits will be squashed and the title will be used as the commit message, the commit message guidelines below have been followed

### Testing

* Ran the focused unit test suite for flags, BaseCommand, Block Node, Mirror Node, Relay, and Explorer.
* Ran `task format`, `npx tsc --noEmit --pretty false`, and targeted ESLint checks.

The following was not tested:

* Integration/end-to-end testing and deployment to a real multi-cluster Kind environment were not run; the behavior is covered by focused unit tests.

<details>
<summary>
Commit message guidelines
</summary>

We use Conventional Commits to ensure that our commit messages are easy to read, follow a consistent format, and for automated release note generation. Please follow the guidelines below when writing your commit messages:

1. BREAKING CHANGE: a commit that has a footer BREAKING CHANGE:, or appends a ! after the type/scope, introduces a breaking API change (correlating with MAJOR in Semantic Versioning). A BREAKING CHANGE can be part of commits of any type. NOTE: currently breaking changes will only bump the MAJOR version.

| Prefix | Description | Semantic Version Update | Captured in Release Notes |
| --- | --- | --- | --- |
| feat: | a new feature | MINOR | Yes |
| fix: | a bug fix | PATCH | Yes |
| perf: | performance | PATCH | Yes |
| refactor: | code change that is not feature or fix | none | No |
| test: | adding missing tests | none | No |
| docs: | changes to documentation | none | Yes |
| build: | changes to build process | none | No |
| ci: | changes to CI configuration | none | No |
| style: | formatting, missing semi-colons, etc | none | No |
| chore: | updating grunt tasks etc; no production code change | none | No |

</details>

### DIFF
--- a/src/commands/base.ts
+++ b/src/commands/base.ts
@@ -311,7 +311,9 @@ export abstract class BaseCommand extends ShellRunner {
 
   protected async kindLoadComponentImage(componentImage: string, clusterContext: string): Promise<void> {
     const targetContexts: Context[] = [...new Set<Context>([...this.remoteConfig.getContexts(), clusterContext])];
-    const nonKindContexts: Context[] = targetContexts.filter((context: Context): boolean => !context.startsWith('kind-'));
+    const nonKindContexts: Context[] = targetContexts.filter(
+      (context: Context): boolean => !context.startsWith('kind-'),
+    );
 
     if (nonKindContexts.length > 0) {
       throw new SoloErrors.validation.illegalArgument(
--- a/src/commands/base.ts
+++ b/src/commands/base.ts
@@ -48,6 +48,7 @@ import {LoadDockerImageOptionsBuilder} from '../integration/kind/model/load-dock
 import {checkDockerImageExists} from '../core/helpers.js';
 import {PathEx} from '../business/utils/path-ex.js';
 import {OperatingSystem} from '../business/utils/operating-system.js';
+import {ImageReference, type ParsedImageReference} from '../business/utils/image-reference.js';
 import {getEnvironmentVariable} from '../core/constants.js';
 
 interface DockerDesktopContainerdCheckResult {
@@ -292,6 +293,19 @@ export abstract class BaseCommand extends ShellRunner {
   }
 
   protected splitImageNameTag(imageReference: string): {name: string; tag: string} {
+    const firstSegment: string = imageReference.split('/', 1)[0];
+    const explicitRegistryReference: boolean =
+      (firstSegment.includes('.') || firstSegment.includes(':') || firstSegment === 'localhost') &&
+      imageReference.includes('/');
+
+    if (explicitRegistryReference || this.isLocalRegistryImageReference(imageReference)) {
+      const parsedReference: ParsedImageReference = ImageReference.parseImageReference(imageReference);
+      return {
+        name: `${parsedReference.registry}/${parsedReference.repository}`,
+        tag: parsedReference.tag,
+      };
+    }
+
     const colonIndex: number = imageReference.lastIndexOf(':');
     if (colonIndex === -1) {
       throw new SoloErrors.validation.illegalArgument(
@@ -310,7 +324,12 @@ export abstract class BaseCommand extends ShellRunner {
   }
 
   protected async kindLoadComponentImage(componentImage: string, clusterContext: string): Promise<void> {
-    const targetContexts: Context[] = [...new Set<Context>([...this.remoteConfig.getContexts(), clusterContext])];
+    const additionalKindContexts: Context[] = this.remoteConfig
+      .getContexts()
+      .filter(
+        (context: Context): boolean => context.startsWith('kind-') && context !== clusterContext,
+      );
+    const targetContexts: Context[] = [...new Set<Context>([clusterContext, ...additionalKindContexts])];
     const nonKindContexts: Context[] = targetContexts.filter(
       (context: Context): boolean => !context.startsWith('kind-'),
     );
--- a/src/commands/base.ts
+++ b/src/commands/base.ts
@@ -276,11 +276,19 @@ export abstract class BaseCommand extends ShellRunner {
   }
 
   protected isLocalImageReference(imageReference: string): boolean {
+    if (this.isLocalRegistryImageReference(imageReference)) {
+      return true;
+    }
+
     const withoutTag: string = imageReference.includes(':')
       ? imageReference.slice(0, imageReference.lastIndexOf(':'))
       : imageReference;
     const firstSegment: string = withoutTag.split('/', 1)[0];
     return !firstSegment.includes('.') && !firstSegment.includes(':') && firstSegment !== 'localhost';
+  }
+
+  protected isLocalRegistryImageReference(imageReference: string): boolean {
+    return /^localhost:\d+\//.test(imageReference);
   }
 
   protected splitImageNameTag(imageReference: string): {name: string; tag: string} {
@@ -302,14 +310,29 @@ export abstract class BaseCommand extends ShellRunner {
   }
 
   protected async kindLoadComponentImage(componentImage: string, clusterContext: string): Promise<void> {
-    const kindClusterName: string = this.kindClusterNameFromContext(clusterContext);
-    this.logger.debug(`Loading '${componentImage}' into Kind cluster '${kindClusterName}'`);
+    const targetContexts: Context[] = [...new Set<Context>([...this.remoteConfig.getContexts(), clusterContext])];
+    const nonKindContexts: Context[] = targetContexts.filter((context: Context): boolean => !context.startsWith('kind-'));
+
+    if (nonKindContexts.length > 0) {
+      throw new SoloErrors.validation.illegalArgument(
+        `Component image '${componentImage}' requires Kind image loading, but target cluster context(s) ` +
+          `'${nonKindContexts.join("', '")}' are not Kind clusters. Push the image to a registry reachable ` +
+          'from every target cluster and pass that registry image reference to --component-image.',
+        componentImage,
+      );
+    }
+
     const kindExecutable: string = await this.depManager.getExecutable(constants.KIND);
     const kindClient: KindClient = await this.kindBuilder.executable(kindExecutable).build();
-    await kindClient.loadDockerImage(
-      componentImage,
-      LoadDockerImageOptionsBuilder.builder().name(kindClusterName).build(),
-    );
+
+    for (const targetContext of targetContexts) {
+      const kindClusterName: string = this.kindClusterNameFromContext(targetContext);
+      this.logger.debug(`Loading '${componentImage}' into Kind cluster '${kindClusterName}'`);
+      await kindClient.loadDockerImage(
+        componentImage,
+        LoadDockerImageOptionsBuilder.builder().name(kindClusterName).build(),
+      );
+    }
   }
 
   protected async throwIfNamespaceIsMissing(context: Context, namespace: NamespaceName): Promise<void> {
--- a/src/commands/base.ts
+++ b/src/commands/base.ts
@@ -293,12 +293,7 @@ export abstract class BaseCommand extends ShellRunner {
   }
 
   protected splitImageNameTag(imageReference: string): {name: string; tag: string} {
-    const firstSegment: string = imageReference.split('/', 1)[0];
-    const explicitRegistryReference: boolean =
-      (firstSegment.includes('.') || firstSegment.includes(':') || firstSegment === 'localhost') &&
-      imageReference.includes('/');
-
-    if (explicitRegistryReference || this.isLocalRegistryImageReference(imageReference)) {
+    if (this.isLocalRegistryImageReference(imageReference)) {
       const parsedReference: ParsedImageReference = ImageReference.parseImageReference(imageReference);
       return {
         name: `${parsedReference.registry}/${parsedReference.repository}`,
@@ -326,9 +321,7 @@ export abstract class BaseCommand extends ShellRunner {
   protected async kindLoadComponentImage(componentImage: string, clusterContext: string): Promise<void> {
     const additionalKindContexts: Context[] = this.remoteConfig
       .getContexts()
-      .filter(
-        (context: Context): boolean => context.startsWith('kind-') && context !== clusterContext,
-      );
+      .filter((context: Context): boolean => context.startsWith('kind-') && context !== clusterContext);
     const targetContexts: Context[] = [...new Set<Context>([clusterContext, ...additionalKindContexts])];
     const nonKindContexts: Context[] = targetContexts.filter(
       (context: Context): boolean => !context.startsWith('kind-'),
--- a/src/commands/block-node.ts
+++ b/src/commands/block-node.ts
@@ -374,26 +374,35 @@ export class BlockNodeCommand extends BaseCommand {
 
     if ('componentImage' in config && config.componentImage) {
       if (this.isLocalImageReference(config.componentImage)) {
-        const {name: localImageName, tag: rawTag} = this.splitImageNameTag(config.componentImage);
-        const localImageTag: string = SemanticVersion.getValidSemanticVersion(rawTag, false, 'Block node image tag');
-        if (this.isLocalImageAvailableInDocker(`${localImageName}:${localImageTag}`)) {
-          // Image found locally — kind-load task will load it; set pullPolicy: Never.
-          if (this.isLocalRegistryImageReference(config.componentImage)) {
-            const parsedReference: ParsedImageReference = ImageReference.parseImageReference(config.componentImage);
+        if (this.isLocalRegistryImageReference(config.componentImage)) {
+          const parsedReference: ParsedImageReference = ImageReference.parseImageReference(config.componentImage);
+          if (this.isLocalImageAvailableInDocker(config.componentImage)) {
+            // Image found locally — kind-load task will load it; set pullPolicy: Never.
             chartValues
               .setLiteral('image.registry', parsedReference.registry)
               .set('image.repository', parsedReference.repository)
-              .set('image.tag', localImageTag)
+              .set('image.tag', parsedReference.tag)
               .set('image.pullPolicy', 'Never');
           } else {
+            // Preserve the explicit registry/repository when a local registry image is pulled remotely.
+            chartValues
+              .setLiteral('image.registry', parsedReference.registry)
+              .set('image.repository', parsedReference.repository)
+              .set('image.tag', parsedReference.tag);
+          }
+        } else {
+          const {name: localImageName, tag: rawTag} = this.splitImageNameTag(config.componentImage);
+          const localImageTag: string = SemanticVersion.getValidSemanticVersion(rawTag, false, 'Block node image tag');
+          if (this.isLocalImageAvailableInDocker(`${localImageName}:${localImageTag}`)) {
+            // Image found locally — kind-load task will load it; set pullPolicy: Never.
             chartValues
               .set('image.repository', localImageName)
               .set('image.tag', localImageTag)
               .set('image.pullPolicy', 'Never');
+          } else {
+            // Not in local Docker — plain tag override so K8s can pull from a registry.
+            chartValues.set('image.tag', localImageTag);
           }
-        } else {
-          // Not in local Docker — plain tag override so K8s can pull from a registry.
-          chartValues.set('image.tag', localImageTag);
         }
       } else {
         const parsedReference: ParsedImageReference = ImageReference.parseImageReference(config.componentImage);
@@ -1713,7 +1722,7 @@ export class BlockNodeCommand extends BaseCommand {
   private getLivenessCheckPortNumber(chartVersion: string, componentImage?: string): number {
     let blockNodeVersion: SemanticVersion<string> = new SemanticVersion<string>(chartVersion);
 
-    if (componentImage && this.isLocalImageReference(componentImage)) {
+    if (componentImage && this.isLocalImageReference(componentImage) && !this.isLocalRegistryImageReference(componentImage)) {
       const tag: string = this.splitImageNameTag(componentImage).tag;
       const imageVersion: SemanticVersion<string> = new SemanticVersion<string>(tag);
       if (blockNodeVersion.lessThan(imageVersion)) {
@@ -1746,7 +1755,11 @@ export class BlockNodeCommand extends BaseCommand {
     }
 
     const deployConfig: BlockNodeDeployConfigClass = config as BlockNodeDeployConfigClass;
-    if (deployConfig.componentImage && this.isLocalImageReference(deployConfig.componentImage)) {
+    if (
+      deployConfig.componentImage &&
+      this.isLocalImageReference(deployConfig.componentImage) &&
+      !this.isLocalRegistryImageReference(deployConfig.componentImage)
+    ) {
       const tag: string = this.splitImageNameTag(deployConfig.componentImage).tag;
       componentImageVersion = new SemanticVersion<string>(tag);
     }
--- a/src/commands/block-node.ts
+++ b/src/commands/block-node.ts
@@ -1722,11 +1722,15 @@ export class BlockNodeCommand extends BaseCommand {
   private getLivenessCheckPortNumber(chartVersion: string, componentImage?: string): number {
     let blockNodeVersion: SemanticVersion<string> = new SemanticVersion<string>(chartVersion);
 
-    if (componentImage && this.isLocalImageReference(componentImage) && !this.isLocalRegistryImageReference(componentImage)) {
+    if (componentImage && this.isLocalImageReference(componentImage)) {
       const tag: string = this.splitImageNameTag(componentImage).tag;
-      const imageVersion: SemanticVersion<string> = new SemanticVersion<string>(tag);
-      if (blockNodeVersion.lessThan(imageVersion)) {
-        blockNodeVersion = imageVersion;
+      try {
+        const imageVersion: SemanticVersion<string> = new SemanticVersion<string>(tag);
+        if (blockNodeVersion.lessThan(imageVersion)) {
+          blockNodeVersion = imageVersion;
+        }
+      } catch {
+        // non-semver tags (e.g. implicit latest on tagless local registry refs) cannot drive the port choice
       }
     }
 
@@ -1755,13 +1759,13 @@ export class BlockNodeCommand extends BaseCommand {
     }
 
     const deployConfig: BlockNodeDeployConfigClass = config as BlockNodeDeployConfigClass;
-    if (
-      deployConfig.componentImage &&
-      this.isLocalImageReference(deployConfig.componentImage) &&
-      !this.isLocalRegistryImageReference(deployConfig.componentImage)
-    ) {
+    if (deployConfig.componentImage && this.isLocalImageReference(deployConfig.componentImage)) {
       const tag: string = this.splitImageNameTag(deployConfig.componentImage).tag;
-      componentImageVersion = new SemanticVersion<string>(tag);
+      try {
+        componentImageVersion = new SemanticVersion<string>(tag);
+      } catch {
+        // non-semver tags (e.g. implicit latest on tagless local registry refs) do not bump the component version
+      }
     }
 
     const finalVersion: SemanticVersion<string> =
--- a/src/commands/block-node.ts
+++ b/src/commands/block-node.ts
@@ -378,10 +378,19 @@ export class BlockNodeCommand extends BaseCommand {
         const localImageTag: string = SemanticVersion.getValidSemanticVersion(rawTag, false, 'Block node image tag');
         if (this.isLocalImageAvailableInDocker(`${localImageName}:${localImageTag}`)) {
           // Image found locally — kind-load task will load it; set pullPolicy: Never.
-          chartValues
-            .set('image.repository', localImageName)
-            .set('image.tag', localImageTag)
-            .set('image.pullPolicy', 'Never');
+          if (this.isLocalRegistryImageReference(config.componentImage)) {
+            const parsedReference: ParsedImageReference = ImageReference.parseImageReference(config.componentImage);
+            chartValues
+              .setLiteral('image.registry', parsedReference.registry)
+              .set('image.repository', parsedReference.repository)
+              .set('image.tag', localImageTag)
+              .set('image.pullPolicy', 'Never');
+          } else {
+            chartValues
+              .set('image.repository', localImageName)
+              .set('image.tag', localImageTag)
+              .set('image.pullPolicy', 'Never');
+          }
         } else {
           // Not in local Docker — plain tag override so K8s can pull from a registry.
           chartValues.set('image.tag', localImageTag);
--- a/src/commands/flags.ts
+++ b/src/commands/flags.ts
@@ -538,9 +538,10 @@ export class Flags {
     name: 'component-image',
     definition: {
       describe:
-        'Docker image override. Accepts a registry reference (e.g. ghcr.io/hiero-ledger/block-node-server:0.36.0) ' +
-        'or a local reference (e.g. block-node-server:0.36.0-SNAPSHOT). ' +
-        'Local images found in Docker are automatically loaded into the Kind cluster.',
+        'Docker image override. Supports a published registry reference (e.g. ghcr.io/hiero-ledger/component:1.2.3), ' +
+        'a locally built image (e.g. component:1.2.3), or a Kind-attached local registry ' +
+        '(e.g. localhost:5001/component:1.2.3). Locally available images are loaded into every target Kind ' +
+        'cluster and use pullPolicy: Never. For non-Kind targets, publish the image to a registry reachable by the cluster.',
       defaultValue: '',
       type: 'string',
       alias: 'relay-image',
--- a/src/commands/explorer.ts
+++ b/src/commands/explorer.ts
@@ -436,6 +436,10 @@ export class ExplorerCommand extends BaseCommand {
           }
         }
 
+        if (config.componentImage && this.isLocalImageAvailableInDocker(config.componentImage)) {
+          await this.kindLoadComponentImage(config.componentImage, config.clusterContext);
+        }
+
         await this.chartManager.upgrade(
           config.namespace,
           config.releaseName,
--- a/src/commands/explorer.ts
+++ b/src/commands/explorer.ts
@@ -425,8 +425,16 @@ export class ExplorerCommand extends BaseCommand {
               .set('image.tag', parsedReference.tag)
               .setLiteral('image.pullPolicy', 'Never');
           } else if (this.isLocalImageReference(config.componentImage)) {
-            // Local-looking ref but not in Docker — plain tag override, K8s will pull from registry.
-            explorerChartValues.set('image.tag', parsedReference.tag);
+            // Explicit local registry refs keep their registry/repository metadata even when Docker is missing.
+            if (this.isLocalRegistryImageReference(config.componentImage)) {
+              explorerChartValues
+                .setLiteral('image.registry', parsedReference.registry)
+                .setLiteral('image.repository', parsedReference.repository)
+                .set('image.tag', parsedReference.tag);
+            } else {
+              // Local-looking ref but not in Docker — plain tag override, K8s will pull from registry.
+              explorerChartValues.set('image.tag', parsedReference.tag);
+            }
           } else {
             // Explicit registry reference.
             explorerChartValues
--- a/test/unit/commands/base.test.ts
+++ b/test/unit/commands/base.test.ts
@@ -286,6 +286,7 @@ describe('BaseCommand', (): void => {
       'hiero-explorer:my-build',
       'hiero-json-rpc-relay:local',
       'myimage:latest',
+      'localhost:5001/block-node-server',
       'localhost:5001/block-node-server:0.38.0',
     ];
 
@@ -342,14 +343,41 @@ describe('BaseCommand', (): void => {
       );
     });
 
+    it('should ignore non-Kind deployment contexts when loading a local image onto a selected Kind context', async (): Promise<void> => {
+      const baseCommandInternal: BaseCommandInternal = baseCmd as unknown as BaseCommandInternal;
+      const loadDockerImageStub: SinonStub = sinon.stub().resolves();
+      const kindClient: KindClient = {loadDockerImage: loadDockerImageStub} as unknown as KindClient;
+
+      baseCommandInternal.remoteConfig = {
+        getContexts: (): Context[] => ['remote-cluster', 'kind-first'],
+      };
+      baseCommandInternal.depManager = {
+        getExecutable: async (dependency: string): Promise<string> => {
+          expect(dependency).to.equal('kind');
+          return 'kind';
+        },
+      };
+      baseCommandInternal.kindBuilder = {
+        executable: (executable: string): {build: () => Promise<KindClient>} => {
+          expect(executable).to.equal('kind');
+          return {build: async (): Promise<KindClient> => kindClient};
+        },
+      };
+
+      await baseCommandInternal.kindLoadComponentImage('block-node-server:0.38.0', 'kind-first');
+
+      expect(loadDockerImageStub).to.have.been.calledOnce;
+      expect(loadDockerImageStub).to.have.been.calledWith('block-node-server:0.38.0', sinon.match.has('name', 'first'));
+    });
+
     it('should explain how to make an image available to a non-Kind target context', async (): Promise<void> => {
       const baseCommandInternal: BaseCommandInternal = baseCmd as unknown as BaseCommandInternal;
       baseCommandInternal.remoteConfig = {
-        getContexts: (): Context[] => ['remote-cluster'],
+        getContexts: (): Context[] => ['kind-first'],
       };
 
       await expect(
-        baseCommandInternal.kindLoadComponentImage('block-node-server:0.38.0', 'kind-first'),
+        baseCommandInternal.kindLoadComponentImage('block-node-server:0.38.0', 'remote-cluster'),
       ).to.be.rejectedWith(
         "Component image 'block-node-server:0.38.0' requires Kind image loading, but target cluster context(s) " +
           "'remote-cluster' are not Kind clusters. Push the image to a registry reachable from every target " +
--- a/test/unit/commands/base.test.ts
+++ b/test/unit/commands/base.test.ts
@@ -27,6 +27,19 @@ import {type CommandFlag} from '../../../src/types/flag-types.js';
 import {type K8Factory} from '../../../src/integration/kube/k8-factory.js';
 import {OperatingSystem} from '../../../src/business/utils/operating-system.js';
 import {PathEx} from '../../../src/business/utils/path-ex.js';
+import {type KindClient} from '../../../src/integration/kind/kind-client.js';
+
+interface BaseCommandInternal {
+  isLocalImageReference: (imageReference: string) => boolean;
+  kindLoadComponentImage: (componentImage: string, clusterContext: string) => Promise<void>;
+  remoteConfig: {getContexts: () => Context[]};
+  depManager: {getExecutable: (dependency: string) => Promise<string>};
+  kindBuilder: {
+    executable: (executable: string) => {
+      build: () => Promise<KindClient>;
+    };
+  };
+}
 
 class TestBaseCommand extends BaseCommand {
   public async close(): Promise<void> {}
@@ -273,12 +286,12 @@ describe('BaseCommand', (): void => {
       'hiero-explorer:my-build',
       'hiero-json-rpc-relay:local',
       'myimage:latest',
+      'localhost:5001/block-node-server:0.38.0',
     ];
 
     const registryCases: string[] = [
       'ghcr.io/hiero-ledger/block-node-server:0.36.0',
       'docker.io/library/redis:7',
-      'localhost:5000/myimage:tag',
       'registry.example.com/org/image:v1',
     ];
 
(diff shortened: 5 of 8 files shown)


JeffreyDallas:
# Review Log — PR #5949 (hiero-ledger/solo)

Note: `gh` and other network-backed commands were denied by the sandbox for this session
("no approval surface"), so this round is based entirely on the local worktree at HEAD
(`46818c4ce`, branch diffed against `main`) — `git diff main...HEAD`, full file reads, and
`grep`. No PR description/comments could be fetched.

## Round 1 — CLAUDE

### Intent (as inferred from the diff)

`src/commands/base.ts`'s `kindLoadComponentImage` used to load a locally-built component image
into only the single Kind cluster targeted by the current command. This PR (3 commits:
`ddeecb3f2` fix, `0fd5e0e97` style, `46818c4ce` test-fix) extends it to also load into every
context returned by `remoteConfig.getContexts()`, adds `isLocalRegistryImageReference()` to
recognize `localhost:<port>/...` Kind-attached-registry references, teaches `block-node.ts` to
set `image.registry`/`image.repository` from the parsed reference for that case, wires the same
image-load step into `explorer.ts`'s install task, and updates the `--component-image` flag
description accordingly.

### Major

- **`kindLoadComponentImage` requires every consensus-node cluster in the deployment to be Kind,
  even for single-cluster components** — `src/commands/base.ts:312-336`. `targetContexts` is
  built from `this.remoteConfig.getContexts()` (read in full at
  `src/business/runtime-state/config/remote/remote-config-runtime-state.ts:713`: `return
  [...new Set(this.getConsensusNodes().map(node => node.context))]` — i.e. **consensus-node**
  contexts only) unioned with the single `clusterContext` the component is actually being
  deployed to. If any of those contexts isn't Kind, the whole operation throws (lines 316-323),
  even though block-node/mirror-node/relay/explorer (the four call sites, confirmed via `grep -n
  kindLoadComponentImage src/commands/*.ts`) each deploy to exactly one `context`/`clusterContext`
  and never to the consensus-node contexts.
  - Failure scenario: a deployment with consensus nodes on a non-Kind cluster
    (`gke-prod` say) and a block node (or mirror-node/relay/explorer) being added with a locally
    built image to a separate Kind cluster attached to the same deployment. Today that block node
    would only need `gke-prod` to not matter at all; after this change,
    `kindLoadComponentImage` throws `"...target cluster context(s) 'gke-prod' are not Kind
    clusters..."` and blocks the deploy outright, even though `gke-prod` runs no part of this
    component.
  - I confirmed there's no pre-existing precedent for this "load into every cluster" behavior to
    generalize from — `grep -rn loadDockerImage src/commands src/core` shows the single call site
    is this same method; consensus-node image loading (network/node commands) doesn't call it.
  - Counter-evidence I did find: `src/commands/flags.ts:543` ("Locally available images are
    loaded into every target Kind cluster") documents exactly this behavior, so it does look
    intentional rather than accidental — but the flag text doesn't mention that "every target Kind
    cluster" can include clusters that have nothing to do with the component being deployed, nor
    that a single non-Kind cluster anywhere in the deployment blocks components that don't run on
    it. Please confirm this is the intended trade-off (defensive pre-load in case the component
    is later moved/rescheduled onto another of the deployment's Kind clusters) — if so, at least
    the error message should clarify that the offending context need not be where this component
    runs, since as written it will read as a false blocker to an operator debugging a hybrid
    Kind/non-Kind deployment.
  - Evidence tag: **read-in-full** (`remote-config-runtime-state.ts` `getContexts()`, `base.ts`
    `kindLoadComponentImage`) + **executed** (`grep` for other call sites/precedent).

### Minor

- **Duplicated branch bodies** — `src/commands/block-node.ts:381-393`. The `isLocalRegistryImageReference`
  true/false branches both end with `.set('image.tag', localImageTag).set('image.pullPolicy',
  'Never')`; only the repository/registry source differs. Per CLAUDE.md "Enhance Before Creating"
  (DRY), consider factoring out just the repository/registry selection, e.g.:
  ```ts
  const {repository, registry} = this.isLocalRegistryImageReference(config.componentImage)
    ? ImageReference.parseImageReference(config.componentImage)
    : {repository: localImageName, registry: undefined};
  chartValues.set('image.repository', repository).set('image.tag', localImageTag).set('image.pullPolicy', 'Never');
  if (registry) {
    chartValues.setLiteral('image.registry', registry);
  }
  ```
  Not blocking — both branches are short and easy to read as-is.

### Test coverage

- The new cross-cluster behavior is unit-tested in isolation only
  (`test/unit/commands/base.test.ts`, fabricated `remoteConfig.getContexts()`). None of
  `block-node.test.ts`, `explorer.test.ts`, `relay.test.ts`, or the untouched
  `mirror-node.test.ts` exercise their real deploy/upgrade task flow with a `remoteConfig` whose
  `getContexts()` returns a context different from the component's own `context`/`clusterContext`
  — i.e. nothing in the suite would catch a regression in how each of the four call sites plumbs
  `clusterContext` through to the now-multi-cluster `kindLoadComponentImage`, or the new "Major"
  failure mode above. `explorer.test.ts`'s new test (`loads an available local Explorer image
  before chart deployment`) stubs `kindLoadComponentImage` out entirely, so it only proves the
  call happens with the right two arguments, not what happens once multiple contexts are involved.
  Given the Major finding above turns on exactly this interaction, a test with a `remoteConfig`
  mock returning ≥2 contexts (one Kind, one not) for one of the four commands would meaningfully
  raise confidence here.

### Comment and description consistency

- `src/commands/flags.ts:540-544` describe text was checked against the implementation: "Supports
  a published registry reference / a locally built image / a Kind-attached local registry" matches
  `isLocalImageReference`/`isLocalRegistryImageReference` (`src/commands/base.ts:278-291`).
  "Locally available images are loaded into every target Kind cluster" matches the new
  `kindLoadComponentImage` behavior. "For non-Kind targets, publish the image to a registry
  reachable by the cluster" matches the thrown error's remediation text
  (`src/commands/base.ts:318-322`). No stale wording found — see the Major finding above for the
  one place the docs and behavior agree but the combination is worth a second look.
- Comment at `src/commands/block-node.ts:380` ("Image found locally — kind-load task will load
  it; set pullPolicy: Never.") still accurately describes the branch after the new nested
  if/else was added.

### Not flagged (checked and ruled out)

- `--component-image` describe string (`src/commands/flags.ts:544`) runs to ~125 columns,
  over the 120 `printWidth` in `.prettierrc.json`, but there's no `max-len` ESLint rule and the
  file already has many longer pre-existing describe-string lines (verified via `cut -c121-
  src/commands/flags.ts`, e.g. lines 215, 284, 335, 507 predate this PR), so this isn't an
  enforced or new violation — dropping it.
- Considered whether `this.remoteConfig.getContexts()` could be called before remote config is
  loaded for `explorer.ts`'s new call site: confirmed `add`/`upgrade` both run `Initialize` →
  `remoteConfig.loadAndValidate(argv)` before `installExplorerTask` (which contains the new
  `kindLoadComponentImage` call) runs, so this isn't reachable with an unloaded remote config in
  the normal flow.
- `relay.ts`/`mirror-node.ts` unconditionally call `ImageReference.parseImageReference` and set
  `image.registry` even for plain local (non-`localhost:port`) image references, which for a bare
  local tag would set `registry: 'docker.io'` — this looks questionable but is pre-existing code
  untouched by this PR's diff, so it's out of scope here.

No blocking issues found; the Major item above is a request for confirmation/clarification rather
than a confirmed bug, since the flag text suggests the cross-cluster behavior is deliberate.

## Round 1 — CODEX

### P1 — A non-Kind consensus context now blocks a component that is being deployed only to Kind

**Accept (and elevate Claude's Major observation to a blocking behavior regression)** —
`src/commands/base.ts:312-325` builds `targetContexts` from every
`remoteConfig.getContexts()` result plus the selected component context, then rejects the entire
operation if any entry is not `kind-*`.  `getContexts()` is explicitly the distinct contexts of
*consensus nodes* (`src/business/runtime-state/config/remote/remote-config-runtime-state.ts:709-715`),
whereas each affected command obtains one component deployment target from its selected
`clusterRef` (`block-node.ts:871-874`, `relay.ts:675-677`, `mirror-node.ts:1481-1484`, and
`explorer.ts:740-742`) and invokes `kindLoadComponentImage` only for that target
(`block-node.ts:692-700`, `relay.ts:445-452`, `mirror-node.ts:790-805`, and
`explorer.ts:398-441`).

That makes a hybrid deployment fail before a block node, relay, mirror node, or explorer can be
installed on its selected Kind cluster whenever any unrelated consensus-node cluster is managed by
another provider.  This did not happen before this PR, which loaded only the selected context.  Keep
pre-loading every *Kind* consensus context if that is needed by the feature, but filter those extra
contexts to `kind-*` and reject only when the component's own `clusterContext` is non-Kind; add a
command-flow test with a Kind component target and a non-Kind value from `getContexts()`.

Evidence: **read-in-full** (`base.ts`, `remote-config-runtime-state.ts`, and all four callers);
**executed** (the four changed focused unit-test files pass: 51 tests).

### P1 — Tagless `localhost:<port>/…` references no longer follow the supported registry-reference path

`src/commands/base.ts:278-301` now classifies `localhost:5001/component` as local solely from
the `localhost:<port>/` prefix.  Its existing `splitImageNameTag()` then uses the port colon as the
tag delimiter, producing `name: 'localhost'` and `tag: '5001/component'`.  The prior implementation
classified the same string as a registry reference, and `ImageReference.parseImageReference()`
continues to support it by deriving `registry: 'localhost:5001'`, `repository: 'component'`, and
`tag: 'latest'` (`src/business/utils/image-reference.ts:45-76`).  The existing relay test confirms
that an omitted image tag is a supported registry-reference form (`test/unit/commands/relay.test.ts:107-125`).

For block nodes, the malformed `5001/component` value is passed to semantic-version validation
and throws (`src/commands/block-node.ts:375-379`, `src/business/utils/semantic-version.ts:415-428`).
For explorer, the Docker lookup compares the malformed `localhost:5001/component` value with the
`repository:tag` output (`src/core/helpers.ts:750-762`), cannot find the local `:latest` image, and
then takes the local-looking branch that sets only `image.tag` rather than the required registry and
repository (`src/commands/explorer.ts:418-440`).  Thus a previously valid tagless local-registry
override either fails or deploys the chart with the wrong image values.

Restore the previous registry path for a tagless local-registry reference (for example, recognize a
Kind-attached registry as local only when an image tag follows the final slash), or normalize it to
the parser's explicit `:latest` result before Docker lookup and chart-value construction.  Add a
test for `localhost:5001/component` for block-node and explorer as well as the tagged case.

Evidence: **read-in-full** (`base.ts`, `block-node.ts`, `explorer.ts`, `image-reference.ts`,
`helpers.ts`, and `semantic-version.ts`); **executed** (the focused suite passes, but contains no
tagless-local-registry coverage).

### Responses to the remaining Claude observations

- **Reject — duplicated-branch Minor.** `src/commands/block-node.ts:381-393` has only the three
  common Helm setters; factoring it would either introduce a temporary tuple/object solely to avoid
  a short, readable branch or obscure the two distinct image-reference contracts.  This does not
  reach the style guide's §10.1 extraction threshold.
- **Accept — coverage concern.** The new explorer test stubs `kindLoadComponentImage`, and the base
  test has no real command-flow hybrid-context case.  The tests requested in the two P1 findings
  cover this gap directly, so I have not made a separate coverage-only finding.

## Round 2 — CLAUDE

Same sandbox constraint as Round 1: `gh`/network tools were denied ("no approval surface"), so
this round is again based entirely on the local worktree at HEAD (`46818c4ce`) — full file reads
and `grep`, no PR description/comments fetched.

### Response to Codex Round 1 — P1 "non-Kind consensus context blocks a Kind-only component"

**Accept — no objection.** This is the same issue I raised as a Major/confirmation-request in
Round 1; Codex's evidence (all four call sites read in full, each passing only its own
`clusterContext`/`context`) confirms it's a genuine behavior regression, not just a documentation
gap. I have nothing further to add here — converging with Codex's framing (fix: filter
`getContexts()` results to `kind-*` before validating, reject only on the component's own
non-Kind `clusterContext`) rather than my weaker "confirm intent" framing.

### Response to Codex Round 1 — P1 "tagless `localhost:<port>/…` breaks the registry-reference path"

**Accept, with additional confirmation and two more affected call sites.** I independently traced
the same defect and it holds up:

- `isLocalRegistryImageReference()` (`src/commands/base.ts:290-292`) matches `localhost:5001/component`
  (no tag) just as readily as the tagged form — the regex only anchors on `localhost:<port>/`.
- `splitImageNameTag()` (`src/commands/base.ts:294-302`) splits on the *last* colon. For the tagless
  form, that's the port colon, producing `{name: 'localhost', tag: '5001/component'}` — confirmed by
  reading the method directly, not just inferring.
- I confirmed the pre-PR behavior differed: with `isLocalRegistryImageReference` absent,
  `isLocalImageReference('localhost:5001/component')` computed `withoutTag = 'localhost'`
  (only one colon, split there) and `firstSegment = 'localhost'`, which the `firstSegment !==
  'localhost'` check correctly rejected as local — so it fell through to the registry-reference
  path via `ImageReference.parseImageReference()`, which correctly defaults an omitted tag to
  `'latest'` (`src/business/utils/image-reference.ts:47-49`). This PR's new early-return in
  `isLocalImageReference` (`base.ts:279-281`) short-circuits before that check runs, so the regression
  is real and newly introduced by this diff, not pre-existing.
- Confirmed the "supported tagless form" premise directly (not just via Codex's citation):
  `test/unit/commands/relay.test.ts:107-123` ("should accept full relay image reference...") asserts
  `docker.io/library/v400.0` (no tag) resolves to `tag=latest` — so an omitted tag on an
  explicit-registry reference is an existing, tested contract that `localhost:<port>/...` (also an
  explicit-registry form per `ImageReference.parseImageReference`'s own `explicitRegistry` check,
  `image-reference.ts:56-57`) should share but no longer does for the tagless case.

**Additional scope beyond block-node/explorer:** the same root cause also silently degrades
`relay.ts` and `mirror-node.ts`, neither of which is in this PR's diff but both of which call the
now-changed `isLocalImageReference`/`isLocalImageAvailableInDocker` via inheritance:

- `relay.ts:283-296` and `mirror-node.ts:540-570` both call `ImageReference.parseImageReference()`
  unconditionally for chart values (so `registry`/`repository`/`tag=latest` are set correctly even
  for the tagless case — no crash), but both gate `pullPolicy: Never` *and* the `kindLoadComponentImage`
  call (`relay.ts:294`/`450-451`, `mirror-node.ts:561`/`803-804`) behind
  `isLocalImageAvailableInDocker(componentImage)`. That method calls `splitImageNameTag` on the raw
  tagless componentImage, gets the same mis-split (`name: 'localhost', tag: '5001/component'`), and
  `checkDockerImageExists('localhost', '5001/component')` returns `false`. Net effect: a genuinely
  local, Docker-available image at a tagless Kind-attached-registry reference silently deploys with
  the default pull policy (image never gets Kind-loaded, and the cluster may pull-fail or pull a
  stale remote image) instead of failing loudly or working as documented.
- This wasn't reachable before the PR (old `isLocalImageReference` correctly rejected these as
  registry references), so it's an in-scope regression of this diff even though the two files
  aren't in the changed-file list — the shared `BaseCommand` method is what changed.
- Recommend the fix (whatever form it takes — recognizing the reference as local only when a tag
  follows the final slash, or normalizing to `ImageReference.parseImageReference()`'s explicit
  `:latest` before the local/Docker check) be verified against all four call sites, not just
  block-node/explorer, and that a regression test be added for at least one of `relay.ts`/`mirror-node.ts`
  in addition to the two Codex already requested.

Evidence tag: **read-in-full** (`base.ts` isLocalImageReference/isLocalRegistryImageReference/
splitImageNameTag, `image-reference.ts`, `relay.ts` prepareRelayValueArguments region,
`mirror-node.ts` chart-values region) + **executed** (`grep -n kindLoadComponentImage\|
isLocalRegistryImageReference\|isLocalImageAvailableInDocker src/commands/mirror-node.ts
src/commands/relay.ts`).

### Response to Codex's Reject of my Round 1 Minor (duplicated branch bodies)

**Accept the rejection — no objection.** Agreed this doesn't clear the extraction bar; withdrawing
that Minor.

### Response to Codex's Accept of my Round 1 coverage concern

**Agreed, no separate finding needed.** The tests requested by both P1 findings (a hybrid
Kind/non-Kind `getContexts()` case, and a tagless `localhost:<port>/...` case per call site) fully
subsume the coverage gap I flagged in Round 1.

### No new findings

I have no further unresolved objections or new findings beyond the two P1s both reviewers now
agree on and the scope extension above.

STATUS: CLAUDE: RESOLVED: ROUND 2

## Round 2 — CODEX

### Response to Claude/Codex P1 — a non-Kind consensus context blocks a Kind-only component

**Accept.** I rechecked the implementation and the premise holds: `getContexts()` returns only
consensus-node contexts, while `kindLoadComponentImage()` unconditionally validates their union
with the selected component context (`src/commands/base.ts:312-325`). The selected component
context must be validated as Kind, but unrelated consensus contexts must not make its local-image
deployment fail. Filter the extra preload targets to Kind contexts, then validate/load the selected
component context independently.

### Response to Claude/Codex P1 — tagless local-registry references

**Accept, including Claude's relay/mirror-node scope extension.** `localhost:5001/component`
matches the new local-registry predicate but `splitImageNameTag()` treats the port colon as its tag
delimiter (`src/commands/base.ts:279-309`), whereas `ImageReference.parseImageReference()`
correctly gives this explicit-registry form `tag: 'latest'`. The remediation and regression tests
need to cover block-node, explorer, relay, and mirror-node, not only the two changed command files.

### P1 — A tagged local-registry image missing from Docker loses its registry/repository override

`src/commands/block-node.ts:394-396` and `src/commands/explorer.ts:427-429` now treat
`localhost:5001/component:1.2.3` as a local image. If the attached Kind registry contains that
image but `docker images` does not (for example, its local tag was pruned after pushing),
`isLocalImageAvailableInDocker()` is false and both commands set only `image.tag`. They therefore
discard the user-supplied `localhost:5001` registry and `component` repository; Helm uses its
chart defaults instead of the registry image. Before this PR, the same reference was an explicit
registry reference and both commands passed the parser's registry, repository, and tag through.

Keep `pullPolicy: Never` and `kind load docker-image` conditional on Docker availability, but set
the parsed registry/repository/tag for every explicit local-registry reference regardless of that
availability. Add block-node and explorer regressions that make the Docker check return false and
assert the registry/repository/tag are still forwarded (and that explorer does not invoke the load
task). This preserves the existing registry-reference behavior; see the review skill's backwards
compatibility guidance, §10.4.

Evidence: **executed** — the four changed focused suites pass (51 tests), but their new
`localhost:5001/...:tag` tests force Docker availability to true; direct `git show main` comparison
confirms the old explicit-registry path set all three parsed image fields.

## Round 3 — CLAUDE

Same sandbox constraint as prior rounds: `gh`/network tools denied ("no approval surface"), so
this round is based entirely on the local worktree at HEAD (`46818c4ce`) — full file reads,
`git show main:<file>` for pre-PR comparison, and `grep`. No PR description/comments fetched.

### Response to Codex Round 2 — P1 "tagged local-registry image missing from Docker loses its registry/repository override"

**Accept — independently verified, no objection.**

- Confirmed the current code at `src/commands/block-node.ts:375-396`: when
  `isLocalRegistryImageReference(config.componentImage)` is true but
  `isLocalImageAvailableInDocker(...)` is false, execution falls to the outer `else` at line
  391-393 (`chartValues.set('image.tag', localImageTag);` only) — registry and repository are
  never set.
- Confirmed the same shape in `src/commands/explorer.ts:417-434`: the middle `else if
  (this.isLocalImageReference(config.componentImage))` branch (Docker-unavailable, local-looking)
  sets only `image.tag` (line 426), not registry/repository.
- Confirmed the regression via direct pre-PR comparison (`git show main:src/commands/block-node.ts`
  and `git show main:src/commands/explorer.ts`):
  - Pre-PR, `isLocalImageReference('localhost:5001/component:1.2.3')` returned **false**, because
    `isLocalRegistryImageReference` didn't exist yet and the old logic's `firstSegment` for this
    input is `'localhost:5001'`, which contains `:` and fails the `!firstSegment.includes(':')`
    check. So this reference always took the unconditional "explicit registry reference" branch in
    both files (`block-node.ts`'s outer `else`, `explorer.ts`'s final `else`), which sets
    registry+repository+tag regardless of Docker availability.
  - Post-PR, the same reference is classified local by the new `isLocalRegistryImageReference`,
    so it now depends on `isLocalImageAvailableInDocker`. When that's false (image pruned from
    local Docker after being pushed to the Kind-attached registry, or the chart is applied from a
    machine other than the one that built the image — the exact scenario a Kind-attached local
    registry exists to support), the registry/repository override silently disappears and Helm
    falls back to chart defaults instead of the user-specified `localhost:5001/component` image.
  - This is a genuine, newly introduced regression, not pre-existing behavior.
- Confirmed the two other call sites of `isLocalImageAvailableInDocker`
  (`src/commands/relay.ts:283-296`, `src/commands/mirror-node.ts:538-565`) do **not** share this
  defect: both unconditionally call `ImageReference.parseImageReference()` and set
  registry/repository/tag first, then only gate `pullPolicy: Never` on Docker availability. So
  Codex's finding is correctly scoped to block-node and explorer only — no further scope
  extension needed here (unlike the Round 1/2 tagless-reference P1, which did extend to
  relay/mirror-node).
- Confirmed the test-coverage gap: `test/unit/commands/block-node.test.ts:277` and
  `test/unit/commands/explorer.test.ts:541` both stub `isLocalImageAvailableInDocker` to return
  `true` for their new `localhost:5001/...` cases, so neither exercises the Docker-unavailable
  path this finding describes.

Evidence tag: **read-in-full** (`block-node.ts` and `explorer.ts` componentImage branches, both
current and via `git show main:...`) + **executed** (`grep` confirming relay/mirror-node's
unconditional registry/repository assignment, and the test files' stub values).

### No new findings

I have no further unresolved objections or new findings. All three P1s (cross-cluster Kind
validation, tagless local-registry references, and Docker-unavailable local-registry override
loss) are now agreed upon by both reviewers with consistent, independently-verified evidence and
consistent remediation/test guidance.

STATUS: CLAUDE: RESOLVED: ROUND 3

## Round 3 — CODEX

### Responses to the agreed P1 findings

- **Accept — unrelated non-Kind consensus contexts must not block a component's selected Kind target.**
  `getContexts()` returns consensus-node contexts (`remote-config-runtime-state.ts:713-715`), while each command
  supplies its own component context to `kindLoadComponentImage()`. The selected context needs Kind validation;
  additional consensus contexts should be preloaded only when they are Kind contexts.

- **Accept — tagless local-registry references remain broken, including relay and mirror-node.**
  `localhost:5001/component` satisfies the new predicate (`base.ts:290-292`) but `splitImageNameTag()` then takes
  the port colon as the delimiter (`base.ts:294-301`). This is a newly changed path; the image parser's existing
  explicit-registry behavior correctly defaults a missing tag to `latest` (`image-reference.ts:45-62`).

- **Accept — block-node and explorer discard the registry/repository for a Docker-unavailable local-registry image.**
  The branches at `block-node.ts:394-396` and `explorer.ts:427-429` retain only a tag despite already having a
  parsed `localhost:<port>/repository:tag` reference. This is correctly limited to those two commands; relay and
  mirror-node set parsed registry/repository before their Docker-availability checks.

### P1 — block-node newly rejects or rewrites valid local-registry Docker tags

`src/commands/block-node.ts:376-387` routes every `localhost:<port>/...` reference through the bare-local-image
branch because of the new early return in `isLocalImageReference()` (`src/commands/base.ts:278-281`). It therefore
passes the registry tag to `SemanticVersion.getValidSemanticVersion()` at line 378 rather than preserving the parsed
Docker tag. A valid registry reference such as `localhost:5001/block-node-server:latest` now fails before chart
installation: the image-reference validator explicitly accepts `latest` (`image-reference.ts:8-15` and
`test/unit/business/utils/image-reference.test.ts:8-11`), but the semantic-version predicate permits numeric
semantic versions only (`semantic-version.ts:390-428`). Before this PR the same reference did not satisfy
`isLocalImageReference()` and took the explicit-registry branch at `block-node.ts:398-403`, which preserved its
registry, repository, and tag without semantic-version validation.

The regression also rewrites an available `localhost:5001/block-node-server:v1.2.3` image to `:1.2.3`: the semantic
version helper deliberately removes the `v` prefix (`semantic-version.ts:405-442`; covered at
`test/unit/business/utils/semantic-version.test.ts:75-83`), so line 379 probes a Docker tag different from the
user-supplied image and line 386 sends the different tag to Helm. Keep local-registry references on the generic
`ImageReference` path (and preserve their parsed tag) while conditionally setting `pullPolicy: Never` and invoking
Kind load when the *original* registry reference is locally available. Restrict block-node's semantic-version
normalization to bare locally-built image names, which preserves the pre-PR registry-reference contract. Add
regressions for `localhost:5001/block-node-server:latest` and `localhost:5001/block-node-server:v1.2.3`, with the
Docker lookup both available and unavailable as applicable.

Evidence: **executed** — `npx mocha test/unit/commands/base.test.ts test/unit/commands/block-node.test.ts
test/unit/commands/explorer.test.ts test/unit/commands/relay.test.ts` (51 passing),
`npx mocha test/unit/business/utils/image-reference.test.ts` (7 passing), and
`npx mocha test/unit/business/utils/semantic-version.test.ts` (48 passing); the current added command tests cover
only `localhost:5001/block-node-server:0.38.0`, so none exercises this non-semver/preserved-tag path. Direct
`git show main:src/commands/block-node.ts` comparison confirms the old explicit-registry branch.

## Round 4 — CLAUDE

Same sandbox constraint as prior rounds: `gh`/network tools denied ("no approval surface"), so
this round is based entirely on the local worktree at HEAD (`46818c4ce`) — full file reads,
`git show main:<file>` for pre-PR comparison, and targeted `grep`. No PR description/comments
fetched. Confirmed via `git status`/`git log` that no code has changed since Round 3 — this round
is purely evaluating Codex's new Round 3 finding against the same diff.

### Response to Codex Round 3 — P1 "block-node newly rejects or rewrites valid local-registry Docker tags"

**Accept — independently verified in full, no objection.**

- Confirmed current `block-node.ts:375-397`: for any `componentImage` where
  `isLocalImageReference()` is true (which now includes every `localhost:<port>/...` reference,
  tagged or not, per the Round 1-3 P1s already agreed), the code unconditionally calls
  `SemanticVersion.getValidSemanticVersion(rawTag, false, 'Block node image tag')` on the raw tag
  *before* checking Docker availability or registry-ness.
- Confirmed `SemanticVersion.isSemanticVersion('latest')` (`semantic-version.ts:390-401`) is
  false — `'latest'` is non-numeric and fails the semver regex — so
  `getValidSemanticVersion('latest', ...)` throws `IllegalArgumentError` at
  `semantic-version.ts:427` before block-node ever reaches the Docker-availability or
  local-registry branching. A `localhost:5001/block-node-server:latest` reference — a completely
  ordinary Kind-attached-registry image tag — now hard-fails deployment.
- Confirmed the `v`-prefix rewrite independently: `SemanticVersion.isSemanticVersion` strips a
  leading `v` only for validation (`semantic-version.ts:381-382`), but `toString()`
  (`semantic-version.ts:347-352`) always reconstructs from `major.minor.patch[...]` with no `v`,
  and `getValidSemanticVersion(rawTag, false, ...)` passes `isNeedPrefix = false`
  (`semantic-version.ts:442`: `isNeedPrefix ? value.toPrefixedString() : value.toString()`). So
  `rawTag = 'v1.2.3'` becomes `localImageTag = '1.2.3'` — a different string than what the user
  supplied and (more importantly) different from whatever tag actually exists in the local Docker
  daemon/Kind-attached registry, so the Docker lookup at line 379 (`isLocalImageAvailableInDocker(
  \`${localImageName}:${localImageTag}\`)`) probes the wrong tag and the value sent to Helm
  (line 386/391) is wrong too.
- Confirmed via `git show main:src/commands/block-node.ts` (lines 375-393) that pre-PR,
  `isLocalImageReference('localhost:5001/block-node-server:latest')` returned **false** (old
  logic: `firstSegment = 'localhost:5001'`, which `.includes(':')`, so the check
  `!firstSegment.includes(':')` fails) — so this reference always took the explicit-registry
  branch (`ImageReference.parseImageReference` + `ImageReference.validateImageTag`'s general
  `IMAGE_TAG_REGEX = /^[\w][\w.-]{0,127}$/`, which accepts `latest` and `v1.2.3` verbatim, no
  semver coercion). This is a genuine, newly introduced regression, confirmed by direct
  before/after comparison rather than inference.
- Confirmed the test-coverage gap directly: `test/unit/commands/block-node.test.ts:281` is the
  only `localhost:...` componentImage case in the file and uses `0.38.0` — already semver-shaped
  and un-prefixed — so it cannot detect either the `latest` throw or the `v1.2.3` → `1.2.3`
  silent rewrite. `grep -n "componentImage\|localhost:5001" test/unit/commands/block-node.test.ts`
  confirms no other case exists.
- Confirmed explorer.ts does **not** share this defect: `grep -n "SemanticVersion"
  src/commands/explorer.ts` shows `SemanticVersion` is used for `soloChartVersion` and
  `explorerVersion`, never for `config.componentImage` — explorer's componentImage branches
  (`explorer.ts:418-434`) operate on the raw parsed tag with no semver coercion. So this finding
  is correctly scoped to block-node only, consistent with the pattern established in Round 3 for
  the Docker-unavailable-override-loss P1 (also block-node/explorer-only, not relay/mirror-node).

This is now the fourth agreed P1 in this PR's `componentImage` handling, all independently
re-verified against pre-PR behavior via `git show main:...`:

1. Cross-cluster Kind validation over-broadens `kindLoadComponentImage`'s failure surface
   (Round 1, both reviewers).
2. Tagless `localhost:<port>/...` references break across block-node/explorer/relay/mirror-node
   (Round 1-2, both reviewers).
3. Docker-unavailable local-registry references silently lose registry/repository in
   block-node/explorer (Round 2-3, both reviewers).
4. Block-node's semver coercion rejects or silently rewrites non-semver/`v`-prefixed local-registry
   tags (Round 3 Codex, accepted here).

All four share the same root cause area (`isLocalImageReference`/`isLocalRegistryImageReference`
in `base.ts` plus block-node's per-command handling) and the same recommended remediation
direction: preserve the pre-PR explicit-registry-reference contract (parse via
`ImageReference.parseImageReference`, validate tags with `ImageReference.validateImageTag`, no
semver coercion) for anything matching `isLocalRegistryImageReference`, and reserve
`SemanticVersion` handling for bare locally-built image names that never had registry semantics.
A single focused fix pass addressing all four together (rather than four separate patches) seems
like the right scope for the PR author, but that's a design suggestion, not a new finding.

### No new findings

I have no further unresolved objections or new findings. All four P1s are now agreed upon by both
reviewers with consistent, independently-verified evidence (including direct pre-PR/post-PR
diffing for each) and consistent remediation/test guidance.

STATUS: CLAUDE: RESOLVED: ROUND 4

## Round 4 — CODEX

### Responses to Claude's Round 4 review

- **Accept — P1 cross-cluster Kind validation.** `getContexts()` is limited to consensus-node contexts, so its non-Kind entries must not reject a component targeted at a separate Kind context. The selected component context still requires Kind validation; only additional Kind contexts should be preloaded.

- **Accept — P1 tagless local-registry references.** `localhost:5001/component` is newly classified local, but the existing splitter reads the port as the image-tag delimiter. Preserve the parser's implicit `latest` behavior for this explicit-registry form across block-node, explorer, relay, and mirror-node.

- **Accept — P1 Docker-unavailable local-registry references.** Block-node and explorer retain only the tag when Docker availability is false, losing the requested registry and repository. Forward the parsed image fields independently of local-Docker availability, and limit `Never`/Kind loading to images that are locally available.

- **Accept — P1 block-node tag coercion.** The new classification sends local-registry tags through `SemanticVersion`; this rejects valid Docker tags such as `latest` and rewrites `v1.2.3`. Keep those references on the `ImageReference` path and apply semantic-version normalization only to bare local image names.

Evidence: **executed** — the focused command and image/version suites pass (106 tests), and ESLint reports zero errors. Direct current-versus-parent inspection confirms each behavior was introduced by this PR. No new findings or objections.

STATUS: CODEX: RESOLVED: ROUND 4

JeffreyDallas (commented):
@copilot check if review comments are correct and reasonable, respond or resolve accordingly

src/commands/base.ts
```diff
@@ -302,14 +324,36 @@ export abstract class BaseCommand extends ShellRunner {
   }
 
   protected async kindLoadComponentImage(componentImage: string, clusterContext: string): Promise<void> {
-    const kindClusterName: string = this.kindClusterNameFromContext(clusterContext);
-    this.logger.debug(`Loading '${componentImage}' into Kind cluster '${kindClusterName}'`);
+    const additionalKindContexts: Context[] = this.remoteConfig
+      .getContexts()
+      .filter(
+        (context: Context): boolean => context.startsWith('kind-') && context !== clusterContext,
+      );
```
Fantomasa: Prettier fails Code Style / Standard here.

```suggestion
      .filter((context: Context): boolean => context.startsWith('kind-') && context !== clusterContext);
```

src/commands/base.ts
```diff
@@ -302,14 +324,36 @@ export abstract class BaseCommand extends ShellRunner {
...
+    const targetContexts: Context[] = [...new Set<Context>([clusterContext, ...additionalKindContexts])];
+    const nonKindContexts: Context[] = targetContexts.filter(
+      (context: Context): boolean => !context.startsWith('kind-'),
+    );
+
+    if (nonKindContexts.length > 0) {
```
Fantomasa: A kind cluster attached under a renamed context (no `kind-` prefix) worked before via the passthrough in `kindClusterNameFromContext`, which this makes unreachable. Intentional breaking change?

src/commands/block-node.ts
```diff
@@ -1704,7 +1722,7 @@ export class BlockNodeCommand extends BaseCommand {
   private getLivenessCheckPortNumber(chartVersion: string, componentImage?: string): number {
     let blockNodeVersion: SemanticVersion<string> = new SemanticVersion<string>(chartVersion);
 
-    if (componentImage && this.isLocalImageReference(componentImage)) {
+    if (componentImage && this.isLocalImageReference(componentImage) && !this.isLocalRegistryImageReference(componentImage)) {
```
Fantomasa: This drops valid semver tags on local registry refs (e.g. `localhost:5001/block-node-server:0.40.0`) from the port decision, so an image >= 0.39.0 under an older chart version is probed on the wrong port. Only non-semver tags (implicit `latest`) need excluding, e.g.:

```ts
    if (componentImage && this.isLocalImageReference(componentImage)) {
      const tag: string = this.splitImageNameTag(componentImage).tag;
      try {
        const imageVersion: SemanticVersion<string> = new SemanticVersion<string>(tag);
        if (blockNodeVersion.lessThan(imageVersion)) {
          blockNodeVersion = imageVersion;
        }
      } catch {
        // non-semver tags (e.g. implicit latest on tagless local registry refs) cannot drive the port choice
      }
    }
```

Also fixes the prettier error failing Code Style. Related: the config map update in `blockNodeDeployTask` (`this.splitImageNameTag(componentImage).tag`) writes `VERSION: latest` for a tagless local registry ref.

src/commands/block-node.ts
```diff
@@ -1737,7 +1755,11 @@ export class BlockNodeCommand extends BaseCommand {
...
     const deployConfig: BlockNodeDeployConfigClass = config as BlockNodeDeployConfigClass;
-    if (deployConfig.componentImage && this.isLocalImageReference(deployConfig.componentImage)) {
+    if (
+      deployConfig.componentImage &&
+      this.isLocalImageReference(deployConfig.componentImage) &&
+      !this.isLocalRegistryImageReference(deployConfig.componentImage)
+    ) {
       const tag: string = this.splitImageNameTag(deployConfig.componentImage).tag;
       componentImageVersion = new SemanticVersion<string>(tag);
     }
```
Fantomasa: Same issue: a valid semver tag on a local registry ref should still bump the component version.

```suggestion
    if (deployConfig.componentImage && this.isLocalImageReference(deployConfig.componentImage)) {
      const tag: string = this.splitImageNameTag(deployConfig.componentImage).tag;
      try {
        componentImageVersion = new SemanticVersion<string>(tag);
      } catch {
        // non-semver tags (e.g. implicit latest on tagless local registry refs) do not bump the component version
      }
    }
```

src/commands/flags.ts
```diff
@@ -538,9 +538,10 @@ export class Flags {
     name: 'component-image',
     definition: {
       describe:
-        'Docker image override. Accepts a registry reference (e.g. ghcr.io/hiero-ledger/block-node-server:0.36.0) ' +
-        'or a local reference (e.g. block-node-server:0.36.0-SNAPSHOT). ' +
-        'Local images found in Docker are automatically loaded into the Kind cluster.',
+        'Docker image override. Supports a published registry reference (e.g. ghcr.io/hiero-ledger/component:1.2.3), ' +
+        'a locally built image (e.g. component:1.2.3), or a Kind-attached local registry ' +
+        '(e.g. localhost:5001/component:1.2.3). Locally available images are loaded into every target Kind ' +
+        'cluster and use pullPolicy: Never. For non-Kind targets, publish the image to a registry reachable by the cluster.',
```
Fantomasa: #5939 also asks for the docs page update; it is not in this PR, so the PR should not fully close that issue yet.

src/commands/base.ts
Fantomasa: nit: prettier wants this on one line, it is what trips the Code Style check. A quick `task format` run should clear it.

```suggestion
      .filter((context: Context): boolean => context.startsWith('kind-') && context !== clusterContext);
```

src/commands/base.ts
Fantomasa, line 330: Question: could this affect kind clusters attached under a renamed context (without the `kind-` prefix)? `kindClusterNameFromContext` used to pass those through, so they worked before. Totally fine if dropping that is intentional, just wanted to double check.

src/commands/base.ts
```diff
@@ -302,14 +324,36 @@ export abstract class BaseCommand extends ShellRunner {
...
 
   protected async kindLoadComponentImage(componentImage: string, clusterContext: string): Promise<void> {
-    const kindClusterName: string = this.kindClusterNameFromContext(clusterContext);
-    this.logger.debug(`Loading '${componentImage}' into Kind cluster '${kindClusterName}'`);
+    const additionalKindContexts: Context[] = this.remoteConfig
+      .getContexts()
+      .filter(
+        (context: Context): boolean => context.startsWith('kind-') && context !== clusterContext,
+      );
+    const targetContexts: Context[] = [...new Set<Context>([clusterContext, ...additionalKindContexts])];
```
Fantomasa, line 325: Two small edge cases I was wondering about: a stale or unrelated `kind-*` context in the remote config would now fail the whole deploy, and extra non-kind contexts are skipped quietly, so their pods could still hit `ErrImageNeverPull` later. Would it make sense to target only the contexts the component actually deploys to?

src/commands/base.ts
```diff
@@ -276,14 +277,35 @@ export abstract class BaseCommand extends ShellRunner {
...
+    return /^localhost:\d+\//.test(imageReference);
+  }
+
   protected splitImageNameTag(imageReference: string): {name: string; tag: string} {
+    const firstSegment: string = imageReference.split('/', 1)[0];
+    const explicitRegistryReference: boolean =
+      (firstSegment.includes('.') || firstSegment.includes(':') || firstSegment === 'localhost') &&
+      imageReference.includes('/');
+
+    if (explicitRegistryReference || this.isLocalRegistryImageReference(imageReference)) {
```
Fantomasa: Small simplification idea, feel free to ignore: since all current call sites are already guarded to local refs, `isLocalRegistryImageReference` alone might be enough here.

src/commands/block-node.ts
Fantomasa: I might be missing context here, but this looks like it could drop valid semver tags on local registry refs (e.g. `localhost:5001/block-node-server:0.40.0`) from the port decision, so an image >= 0.39.0 under an older chart version might get probed on the wrong port. Maybe only the non-semver tags (implicit `latest`) need excluding, something like:

```ts
    if (componentImage && this.isLocalImageReference(componentImage)) {
      const tag: string = this.splitImageNameTag(componentImage).tag;
      try {
        const imageVersion: SemanticVersion<string> = new SemanticVersion<string>(tag);
        if (blockNodeVersion.lessThan(imageVersion)) {
          blockNodeVersion = imageVersion;
        }
      } catch {
        // non-semver tags (e.g. implicit latest on tagless local registry refs) cannot drive the port choice
      }
    }
```

This would also take care of the prettier error on this line. One more tiny thing I noticed nearby: the config map update in `blockNodeDeployTask` would write `VERSION: latest` for a tagless local registry ref, not sure if that is intended.

src/commands/block-node.ts
Fantomasa: Same small concern as above, a valid semver tag on a local registry ref could probably still bump the component version, e.g.:

```suggestion
    if (deployConfig.componentImage && this.isLocalImageReference(deployConfig.componentImage)) {
      const tag: string = this.splitImageNameTag(deployConfig.componentImage).tag;
      try {
        componentImageVersion = new SemanticVersion<string>(tag);
      } catch {
        // non-semver tags (e.g. implicit latest on tagless local registry refs) do not bump the component version
      }
    }
```

src/commands/flags.ts
Fantomasa, line 544: Nice improvement to the help text! One small note: #5939 also mentions a docs page update, so it might be worth keeping that issue open until the page lands, or adding it here.

src/commands/base.ts
JeffreyDallas: Fixed in 6fdf624 — ran `task format` on the file.

src/commands/base.ts
JeffreyDallas: Good catch, applied in 6fdf624 — every current call site already guards `splitImageNameTag` behind `isLocalImageReference`, and the only "explicit registry" shape that survives that guard is the `localhost:<port>/...` one, so `isLocalRegistryImageReference` alone covers it. Dropped the redundant `explicitRegistryReference` branch.

src/commands/block-node.ts
JeffreyDallas: You're right, applied your suggested fix in 6fdf624 — a tagged local-registry ref like `localhost:5001/block-node-server:0.40.0` now still participates in the port decision; only tags that fail `SemanticVersion` parsing (e.g. the implicit `latest` on a tagless ref) are skipped.

src/commands/block-node.ts
JeffreyDallas: Same fix applied in 6fdf624 — a valid semver tag on a local-registry ref can still bump the recorded component version; only non-semver tags are ignored now.

src/commands/base.ts
JeffreyDallas: Good question. This intentionally matches the `kind-` prefix convention already used elsewhere in the codebase (`kindClusterNameFromContext` right above it, `backup-restore.ts`, `cache.ts`) — a context renamed away from that prefix would already behave inconsistently at those other call sites today, so this isn't introducing a new fragile heuristic, just reusing the existing one. Doing real detection (e.g. querying `kind get clusters` and matching by name regardless of context naming) would be a bigger change I'd rather scope as a follow-up if renamed contexts turn out to matter in practice. Leaving this thread open in case you'd like an issue filed for that.

src/commands/base.ts
JeffreyDallas: Agreed the precise fix is to target only the contexts the component itself deploys to, but today `remoteConfig` doesn't expose that per-component (only `getContexts()` over consensus nodes), so doing it right means threading each command's actual deployment targets through `kindLoadComponentImage` — bigger than this PR. In the meantime the contract is: the caller-supplied `clusterContext` is the only one required to succeed (that's what throws on non-Kind), and the extra `remoteConfig` Kind contexts are best-effort preloading only — if the component isn't actually deployed to one of those extra contexts, there's no pod there to hit `ErrImageNeverPull`. A stale `kind-*` context that's been deleted would still surface as a real (if less friendly) error from the Kind CLI rather than silently succeeding. Open to filing a follow-up issue to scope the per-component-target version if you think it's worth doing before this ships.

src/commands/flags.ts
JeffreyDallas: Good call — the docs.hedera.com page update is real follow-up work, so I'll keep #5939 open (or split a dedicated docs sub-issue) rather than relying on this PR to close it. Will adjust the PR description accordingly.

src/commands/base.ts
Fantomasa: Filed #5963 to track Kind cluster detection that does not depend on the kind- context prefix. Resolving.

src/commands/base.ts
Fantomasa: Filed #5964 to track loading only into the component's actual deployment contexts and making the extra-context preloading truly best-effort. Resolving.

src/commands/flags.ts
Fantomasa: The PR description now marks #5939 as tracked separately, works for me. Resolving.

Fantomasa (approved):
Review feedback has been addressed, thanks for the quick turnaround. The two remaining review discussions are tracked as follow-up issues:

- #5963 support Kind clusters attached under renamed contexts in component image loading
- #5964 load component images only into the component's actual target clusters and make the extra-context preloading best-effort

jan-milenkov (approved):

PR #5950, opened 2026-09-02
feat: load component images from archives

### LINKED ISSUE #5940: Add a flag to load a component image from a docker save archive

JeffreyDallas:
## Description

* Need to merge #5949 First
* Adds `--component-image-archive` for loading a `docker save` archive into every target Kind cluster.
* Requires `--component-image` with the archive to set the chart image reference and `pullPolicy: Never`.
* Supports Block Node, Mirror Node, JSON-RPC Relay, and Explorer deployments using shared archive-loading behavior.
* Validates missing archives and non-Kind target contexts before deployment.

### Related Issues

* Closes #5940
* Related to #5939
* Related to #5938

* [x] When this PR merges the commits will be squashed and the title will be used as the commit message, the commit message guidelines below have been followed

### Testing

* Ran the focused unit tests for flags, BaseCommand, Block Node, Mirror Node, JSON-RPC Relay, and Explorer.
* Ran TypeScript compilation, targeted ESLint checks, circular-dependency analysis, and `task format`.

The following was not tested:

* A real multi-cluster Kind deployment using a `docker save` archive and integration/end-to-end tests were not run.

<details>
<summary>
Commit message guidelines
</summary>

We use Conventional Commits to ensure that our commit messages are easy to read, follow a consistent format, and for automated release note generation. Please follow the guidelines below when writing your commit messages:

1. BREAKING CHANGE: a commit that has a footer BREAKING CHANGE:, or appends a ! after the type/scope, introduces a breaking API change (correlating with MAJOR in Semantic Versioning). A BREAKING CHANGE can be part of commits of any type. NOTE: currently breaking changes will only bump the MAJOR version.

| Prefix | Description | Semantic Version Update | Captured in Release Notes |
| --- | --- | --- | --- |
| feat: | a new feature | MINOR | Yes |
| fix: | a bug fix | PATCH | Yes |
| perf: | performance | PATCH | Yes |
| refactor: | code change that is not feature or fix | none | No |
| test: | adding missing tests | none | No |
| docs: | changes to documentation | none | Yes |
| build: | changes to build process | none | No |
| ci: | changes to CI configuration | none | No |
| style: | formatting, missing semi-colons, etc | none | No |
| chore: | updating grunt tasks etc; no production code change | none | No |

</details>


src/commands/explorer.ts
```diff
@@ -443,9 +451,7 @@ export class ExplorerCommand extends BaseCommand {
           }
         }
 
-        if (config.componentImage && this.isLocalImageAvailableInDocker(config.componentImage)) {
-          await this.kindLoadComponentImage(config.componentImage, config.clusterContext);
-        }
+await this.loadComponentImage(config.componentImage, config.componentImageArchive, config.clusterContext);
```
JeffreyDallas: This line sits at column 1 inside an 8-space-indented method body — an unresolved merge/format conflict. `prettier/prettier` is an error-level rule in this repo's ESLint config, so this fails `task build`/CI's `build:lint` step as-is. Run `task format` and commit the result (also affects `test/unit/commands/base.test.ts` and `test/unit/commands/relay.test.ts`, which have the same pattern).

src/commands/block-node.ts
```diff
@@ -1719,10 +1735,14 @@ export class BlockNodeCommand extends BaseCommand {
   /// web server on `BLOCK_NODE_HEALTH_PORT`; earlier versions served them from the gRPC port
   /// (`BLOCK_NODE_PORT`). The effective version is the higher of the requested chart version and
   /// a local image tag (when set), mirroring `updateBlockNodeVersionInRemoteConfig`.
-  private getLivenessCheckPortNumber(chartVersion: string, componentImage?: string): number {
+  private getLivenessCheckPortNumber(
+    chartVersion: string,
+    componentImage?: string,
+    componentImageArchive?: string,
```
JeffreyDallas: `componentImageArchive` is never read in this method's body (confirmed by ESLint's `unused-imports/no-unused-vars` error at this line). Separately, the guard at line 1745 was widened from the pre-existing `componentImage && this.isLocalImageReference(componentImage)` to a bare `if (componentImage)`, so any `--component-image` value — including a genuine remote-registry reference with no archive and nothing in local Docker — can now influence the liveness-port/version selection, unlike every other call site touched in this PR (`generateChartValues`, `isComponentImageAvailableForKind`, `updateBlockNodeVersionInRemoteConfig`), which all gate this behind `hasComponentImageArchiveValue(...) || isLocalImageAvailableInDocker(...)`. Please either wire the parameter into the same gating pattern, or remove it and update the stale doc comment above (lines 1736-1737, "a local image tag (when set)") plus add a test for a plain remote-registry image with no archive.

src/commands/base.ts
```diff
@@ -349,6 +393,47 @@ export abstract class BaseCommand extends ShellRunner {
...
+      );
+    }
+  }
+
+  private getKindTargetContexts(
+    componentImageSource: string,
+    sourceFlagName: string,
+    clusterContext: Context,
+  ): Context[] {
+    const targetContexts: Context[] = [...new Set<Context>([...this.remoteConfig.getContexts(), clusterContext])];
```
JeffreyDallas: This drops the `context.startsWith('kind-')` pre-filter that `main`'s `kindLoadComponentImage` (from already-merged #5949) applied to `remoteConfig.getContexts()` before computing `nonKindContexts`. Previously a non-Kind sibling context (e.g., a remote consensus-node cluster alongside a Kind-hosted block node) was excluded from the Kind-loading target set and never tripped this check; now any non-Kind entry anywhere in `remoteConfig.getContexts()` makes it throw, even when the selected `clusterContext` is itself a valid Kind cluster. The test asserting the old behavior (`'should ignore non-Kind deployment contexts when loading a local image onto a selected Kind context'` in `test/unit/commands/base.test.ts`) was deleted with no replacement for the mixed remote+Kind case. Please restore the `kind-*` pre-filter on `remoteConfig.getContexts()` before unioning with `clusterContext`, and reinstate a test for the mixed-cluster case.

src/commands/base.ts
```diff
@@ -318,24 +319,67 @@ export abstract class BaseCommand extends ShellRunner {
...
 
-  protected async kindLoadComponentImage(componentImage: string, clusterContext: string): Promise<void> {
-    const additionalKindContexts: Context[] = this.remoteConfig
-      .getContexts()
-      .filter((context: Context): boolean => context.startsWith('kind-') && context !== clusterContext);
-    const targetContexts: Context[] = [...new Set<Context>([clusterContext, ...additionalKindContexts])];
-    const nonKindContexts: Context[] = targetContexts.filter(
-      (context: Context): boolean => !context.startsWith('kind-'),
-    );
+  protected hasComponentImageArchiveValue(
```
JeffreyDallas: `componentImage` is never read in this method's body (`Boolean(componentImageArchive?.trim())` only uses `componentImageArchive`). Either drop the unused parameter or use it in the check.

src/commands/block-node.ts
```diff
@@ -393,7 +399,7 @@ export class BlockNodeCommand extends BaseCommand {
         } else {
           const {name: localImageName, tag: rawTag} = this.splitImageNameTag(config.componentImage);
           const localImageTag: string = SemanticVersion.getValidSemanticVersion(rawTag, false, 'Block node image tag');
-          if (this.isLocalImageAvailableInDocker(`${localImageName}:${localImageTag}`)) {
+          if (this.isLocalImageAvailableInDocker(`${localImageName}:${localImageTag}`) || hasComponentImageArchive) {
```
JeffreyDallas: This condition also covers `hasComponentImageArchive`, but the following comment still says "Image found locally." Please use the archive-aware wording already present in the sibling branch.

src/commands/block-node.ts
```diff
@@ -1759,7 +1780,10 @@ export class BlockNodeCommand extends BaseCommand {
     }
 
     const deployConfig: BlockNodeDeployConfigClass = config as BlockNodeDeployConfigClass;
-    if (deployConfig.componentImage && this.isLocalImageReference(deployConfig.componentImage)) {
+    if (
+      deployConfig.componentImage &&
+      this.isComponentImageAvailableForKind(deployConfig.componentImage, deployConfig.componentImageArchive)
```
JeffreyDallas: This replacement makes remote-config version tracking narrower than before: without an archive, `isComponentImageAvailableForKind()` also requires the local-style reference to exist in Docker, while this branch previously accepted every `isLocalImageReference()` value. A valid local-style reference that Kubernetes pulls from its registry will therefore leave the recorded Block Node version at the chart version, which can select the wrong migration start point later. Preserve the prior local-reference predicate and add archive presence as a separate alternative, as the liveness-version branch does above.

src/commands/block-node.ts
```diff
@@ -1718,11 +1731,19 @@ export class BlockNodeCommand extends BaseCommand {
...
+    chartVersion: string,
+    componentImage?: string,
+    componentImageArchive?: string,
+  ): number {
     let blockNodeVersion: SemanticVersion<string> = new SemanticVersion<string>(chartVersion);
 
-    if (componentImage && this.isLocalImageReference(componentImage)) {
+    if (
+      componentImage &&
+      (this.isLocalImageReference(componentImage) || this.hasComponentImageArchiveValue(componentImageArchive))
```
JeffreyDallas: The archive disjunct makes a tagless registry reference such as `ghcr.io/hiero-ledger/block-node-server` reach `splitImageNameTag()` below, which throws before the following `try`. `ImageReference.parseImageReference()`—already used in the chart-value branch—accepts the same reference and defaults its tag to `latest`. Please use that tag-aware parser here and at the equivalent tag extraction sites on lines 986 and 1788, then cover the tagless registry/archive case with a regression test.

src/commands/base.ts
```diff
@@ -369,6 +417,58 @@ export abstract class BaseCommand extends ShellRunner {
...
+  }
+
+  private getKindTargetContexts(
+    componentImageSource: string,
+    sourceFlagName: string,
+    clusterContext: Context,
+  ): Context[] {
+    const additionalKindContexts: Context[] = this.remoteConfig
+      .getContexts()
+      .filter((context: Context): boolean => context.startsWith('kind-') && context !== clusterContext);
```
Fantomasa: This reintroduces the remote-config fan-out that #5970 removed: `getContexts()` only exposes consensus-node contexts, not the clusters this component deploys to, and since the archive loop has no per-context error handling, a stale extra Kind context will fail the whole deploy again (#5964). Can we scope the archive load to the caller's context plus an optional best-effort `additionalContexts`, mirroring `kindLoadComponentImage` just above?

src/commands/base.ts
```diff
@@ -369,6 +417,58 @@ export abstract class BaseCommand extends ShellRunner {
...
+    const targetContexts: Context[] = [...new Set<Context>([clusterContext, ...additionalKindContexts])];
+    const nonKindContexts: Context[] = targetContexts.filter(
+      (context: Context): boolean => !context.startsWith('kind-'),
```
Fantomasa: The `kind-` prefix check regresses #5979: a renamed Kind context (whose kubeconfig cluster entry is still `kind-<name>`) is detected by `kindClusterNameFromContext`, so `--component-image` alone works there, but the same deploy with `--component-image-archive` would throw here. Suggest using `kindClusterNameFromContext` for the detection so both paths agree.

Fantomasa: **AI Suggestion**: The `kind-` prefix check regresses #5979: a renamed Kind context works with `--component-image` via `kindClusterNameFromContext`, but throws here with `--component-image-archive`. Suggest using `kindClusterNameFromContext` so both paths agree.

src/commands/base.ts
```diff
@@ -369,6 +417,63 @@ export abstract class BaseCommand extends ShellRunner {
...
+    for (const context of targetContexts) {
+      const kindClusterName: string | undefined = this.kindClusterNameFromContext(context);
+      if (kindClusterName === undefined) {
+        nonKindContexts.push(context);
+        continue;
+      }
+      kindClusterNames.add(kindClusterName);
+    }
+
+    if (nonKindContexts.length > 0) {
```
ddeskov-limechain, line 457: This is stricter than the `--component-image` path. There, only the selected context must be Kind. Other contexts get a warn-and-skip, and a failed extra load is best-effort (see `kindLoadComponentImage` above).

Here, any non-Kind context in the deployment fails the whole command, even though the component is only deployed to `clusterContext`. A deployment with one Kind cluster and one remote cluster can not use an archive for a block node that targets the Kind cluster.

Suggestion: keep the throw only for `clusterContext`, and treat the other contexts the same way as `kindLoadComponentImage` does (warn on non-Kind, try/catch the extra loads). That loop could also be shared by both paths, since the only difference is the kind client call.

Related: `loadComponentImage` never passes the extra contexts to `kindLoadComponentImage`, so the docker path still loads only the primary cluster. If both paths used the same loop, they would agree.

src/commands/base.ts
```diff
@@ -319,6 +320,58 @@ export abstract class BaseCommand extends ShellRunner {
...
+    }
+
+    if (!componentImage?.trim()) {
+      throw new SoloErrors.validation.illegalArgument(
+        `--${flags.componentImageArchive.name} requires --${flags.componentImage.name} to identify the image in the archive.`,
+        componentImageArchive,
+      );
+    }
+
+    if (!fs.existsSync(componentImageArchive)) {
```
ddeskov-limechain, line 342: Nothing checks that the archive really contains the tag given in `--component-image`. If they differ, the archive loads fine, but the pod ends in `ErrImageNeverPull` much later and the reason is not obvious.

A cheap check: read `manifest.json` from the tar (`tar -xOf <archive> manifest.json`) and make sure `RepoTags` contains the image reference. Failing early here with a clear message would save a confusing debug session. Optional, but I think worth it.

ddeskov-limechain (commented):
Read through the archive path end to end. The flag wiring, chart values (`pullPolicy: Never`) and validation look right, and the tests cover the main cases. Two mid-level points inline about the multi-cluster behavior and a wrong-tag archive.

src/commands/base.ts
JeffreyDallas: Fixed in 561e56a7b: extracted a shared `preloadIntoKindClusters` helper used by both `kindLoadComponentImage` and `kindLoadComponentImageArchive`. Now only the caller-supplied `clusterContext` is a hard requirement; additional contexts get warn-and-skip (non-Kind) or try/catch best-effort (stale/unreachable), matching the docker-image path exactly. Also fixed the related gap: `loadComponentImage` now passes `remoteConfig.getContexts()` as `additionalContexts` to both paths, so a plain `--component-image` deploy preloads every known Kind cluster too, not just the primary one.

src/commands/base.ts
JeffreyDallas: Fixed in 561e56a7b: `validateComponentImageArchive` now extracts `manifest.json` from the archive with the `tar` package (already a dependency, no shell-out) and confirms `RepoTags` includes the `--component-image` reference, throwing a new registered `ComponentImageArchiveTagMismatchSoloError` (SOLO-4082) if not. If the manifest is missing or unparsable (e.g. a non-Docker tar), the check is skipped rather than blocking an otherwise-valid deploy. Covered by new tests in `base.test.ts` ("validateComponentImageArchive manifest tag check").

Fantomasa (approved):
LGTM

jan-milenkov (approved):

Issue #5931, opened 2026-08-31
Remediate pre-existing zizmor security findings across CI/CD workflows



JeffreyDallas:
## Overview
PR #5930 adds a `zizmor-scan` job (`.github/workflows/zxf-prs-automation.yaml`) that runs [zizmor](https://docs.zizmor.sh/) static security analysis on any PR touching `.github/workflows/**` or `.github/actions/**`. Running zizmor v1.30.0 against the full repository today found 270 pre-existing findings across 28 workflow files plus `.github/actions/solo-image-cache/action.yaml`. To let the new check start green (and only flag genuinely new issues) these were baselined — suppressed via `.github/zizmor.yml` (`rules.<id>.ignore` entries by `file:line`) and inline `# zizmor: ignore[...]` comments in `solo-image-cache/action.yaml` (composite actions can't use the config-file mechanism) — **not fixed**. This issue tracks the actual remediation.

Findings by rule:
- `template-injection` (code injection via unescaped `${{ }}` expansion into `run:` blocks): 139 — high severity/high confidence
- `self-repository`: 43
- `artipacked` (checkout not setting `persist-credentials: false`): 27
- `excessive-permissions`: 23
- `adhoc-packages`: 13
- `cache-poisoning`: 6
- `secrets-inherit`: 3
- `dangerous-triggers`: 2
- `unsound-condition`: 2
- `unpinned-images`: 1

**Caution:** zizmor's own `--fix=safe` auto-fix is broken as of v1.30.0 — it corrupted local reusable-workflow references (rewrote `uses: ./.github/workflows/x.yaml` to `uses: $/.github/workflows/x.yaml`). Do not trust automated fixes; each finding needs manual review and testing per workflow.

## Sub-Tasks
- [ ] Remediate `flow-deploy-release-artifact.yaml` (highest finding count)
- [ ] Remediate `flow-ledger-reset-warm-start-timing.yaml`
- [ ] Remediate `flow-one-shot-timing.yaml`
- [ ] Remediate `zxc-e2e-test.yaml`
- [ ] Remediate remaining flagged workflows and `solo-image-cache/action.yaml`
- [ ] Remove the corresponding `.github/zizmor.yml` ignore entries (and inline ignore comments) as each file is fixed and verified
- [ ] Once the baseline is empty, delete `.github/zizmor.yml`

## References
- [PR #5930](https://github.com/hiero-ledger/solo/pull/5930) — adds the zizmor-scan job and the baseline this issue tracks

PR #5953, opened 2026-09-02
fix(ci): stop persisting the job token in checkout steps

JeffreyDallas:
## Description

Tiers 3 and 4 of the stacked zizmor remediation series, combined on this branch.

**Why combined:** this PR was originally scoped to tier 3 only (`artipacked`). Tier 4
(`template-injection`, originally #5956) was opened with base `zizmor-cleanup-tier3` and merged
there rather than into `main`. Separately, once tier 2 (#5948) merged and its branch was deleted,
GitHub auto-retargeted this PR's base from `zizmor-cleanup-tier2` to `main`. The net effect is that
`zizmor-cleanup-tier3` — this PR's head branch — now carries both tiers' commits, so the diff below
is genuinely **144 additions / 55 deletions across 26 files**, not the tier-3-only 50/0/23 originally
described. Tier 4 was reviewed and approved on its own as #5956 before landing here; this PR now
carries both, and merging it brings both tiers into `main` together.

zizmor is run **offline** (v1.30.0, `--no-config`); this series adds no scanning job and no
ignore/baseline configuration.

### Tier 3 — clears `artipacked`: 27 → 0

Every `actions/checkout` step flagged by the audit now opts out of credential persistence:

```diff
       - name: Checkout Code
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
```

27 `persist-credentials: false` keys across 23 workflows, plus 23 `with:` blocks for the steps that
did not already have one (4 steps had an existing `with:` and only needed the key).

By default `actions/checkout` writes the job token into the runner's git config as an
`http.extraheader`. Anything that later archives the workspace — an uploaded artifact, a cached
directory — can carry `.git/config` along with it and leak a usable credential outside the job.
`persist-credentials: false` stops the token from being written at all.

`persist-credentials: false` is only safe if nothing after the checkout relies on the stored
credential, so all 23 files were audited rather than assuming. Every git operation that runs after a
checkout falls into one of three safe cases:

* **Unauthenticated clones of a public third-party repo** — `git clone
  https://github.com/hiero-ledger/hiero-consensus-node.git` in `flow-examples-test.yaml`,
  `flow-ledger-reset-warm-start-timing.yaml`, `flow-one-shot-timing.yaml` and `zxc-e2e-test.yaml`.
  Different repository, no credential involved.
* **An anonymous fetch of this repository** — `zxc-code-style.yaml` runs `git fetch origin
  "$target_branch":"$target_branch"` to diff lint warnings against the PR base. This repository is
  public, so the fetch succeeds without a credential.
* **`gh release upload`** in `flow-deploy-release-artifact.yaml`, which authenticates with
  `GH_TOKEN`/`GH_ACCESS_TOKEN` rather than through git.

No workflow pushes, and none uses `submodules:` or a custom `extraheader`. `zxc-code-style.yaml` is
the one call site whose behaviour depends on this repository staying public — worth recording,
though it is not a change in risk introduced here.

### Tier 4 — clears High-severity `template-injection`: 42 → 0

GitHub substitutes `${{ … }}` into a `run:` block **as text**, before the shell ever sees it. A value
containing quotes or shell metacharacters can terminate the surrounding quoting and execute arbitrary
commands with the job's privileges. Routing the value through the step environment instead means the
shell receives it as data.

Where the expression is a `github.*` context the runner already publishes, the existing variable is
used rather than inventing a new binding — `GITHUB_REF`, `GITHUB_REF_NAME`, `GITHUB_EVENT_NAME`,
`GITHUB_ACTOR`. Workflow-level `env:` values (`CG_EXEC`, `SOLO_CLUSTER_DUALITY`) are likewise already
shell variables. Only `inputs.*` and event-payload fields needed new `env:` entries. Comments record
that split where a step mixes both, so the asymmetry does not read as an oversight.

Blocks were hoisted whole rather than line-by-line, which also cleared the Informational and Low
findings sharing those blocks — `template-injection` drops by 42 rather than by the 20 High findings
alone.

Files touched:

* **`.github/actions/solo-image-cache/action.yaml`** (6, all High) — a composite action used by ~10
  workflows. Four inputs were interpolated into two `run:` blocks, including a `case` statement and
  the cache-key assignment. `${RUNNER_OS}`/`${RUNNER_ARCH}` were already shell references and are
  untouched, so **the computed cache key is byte-identical and existing caches still hit**.
* **`.github/workflows/flow-hugo-publish.yaml`** (8, all High) — all in the `Set Inputs` step.
* **`.github/workflows/zxc-e2e-test.yaml`** (11 High) — spread across four `run:` blocks, including
  the task-invocation line and the kind-cluster cleanup loop.
* **`.github/workflows/flow-one-shot-timing.yaml`** (5 High) — the five version-override inputs in
  `Resolve Versions`.
* **`.github/workflows/flow-deploy-release-artifact.yaml`** (4 High) — the npm dist-tag branch and the
  dry-run `Set Inputs` step.
* **`.github/workflows/flow-performance-test.yaml`** (1 High), **`flow-examples-test.yaml`** (2
  High), **`flow-gcs-test.yaml`** (1 High) — the last stragglers, included so the High class reaches
  zero rather than one.

**Latent bug fixed along the way:** `flow-deploy-release-artifact.yaml` had this in its dry-run
branch:

```bash
echo "Dry run enabled, setting ref to `github.ref` since tag won't exist."
```

The backticks are inside a **double-quoted** string, so the shell treated them as command
substitution and tried to execute `github.ref` as a command. The line is now single-quoted.
Unrelated to the audit, but it was inside a line this change was already rewriting.

**Equivalence:** the rewrite is behaviour-preserving by construction — `GITHUB_REF`,
`GITHUB_REF_NAME`, `GITHUB_EVENT_NAME` and `GITHUB_ACTOR` hold exactly the values of the contexts
they replace, and every other substitution moves a value from one expansion point to another without
changing it. Two previously unquoted expansions (`export CONSENSUS_NODE_VERSION=…`, the `.env` writes
in `zxc-e2e-test.yaml`) are now quoted, which is strictly safer for values containing spaces.

### Remaining after this PR

126 findings: `template-injection` 104 (**0 High** — 97 Informational, 7 Low), `adhoc-packages` 13,
`cache-poisoning` 6, `dangerous-triggers` 2, `unpinned-images` 1.

Four rules are now fully cleared across the series: `unsound-condition` and `secrets-inherit`
(#5947), `self-repository` (#5948), and `artipacked` (this PR, tier 3). Tier 4 (this PR) takes
High-severity `template-injection` to zero; the remaining 104 `template-injection` findings are
Informational/Low noise from `${{ inputs.* }}` / `${{ env.* }}` on trusted sources.

What's left after this PR is either low-risk noise or three rules that need an owner decision rather
than a cleanup commit:

* `dangerous-triggers` — `pull_request_target` and `workflow_run`; remediation means redesigning how
  those workflows obtain privileges, or consciously accepting the risk.
* `cache-poisoning` — Low confidence; needs per-workflow judgement about whether caching belongs in a
  publishing path.
* `unpinned-images` — `archlinux:latest`, where an inline comment already documents why a rolling
  distro is not pinned. Repo policy in `CLAUDE.md` forbids floating tags, so this needs an explicit
  ruling.

### Related Issues

* Related to #5931
* Absorbs #5956 (tier 4, "clear all high-severity template injection findings") — that PR was
  reviewed and merged on its own, into this PR's head branch rather than into `main`; its commit now
  merges to `main` as part of this PR.

Two boxes are left unchecked. **No tests were added:** these are CI workflow definitions, which the
repository's unit and E2E suites do not exercise; verification was static and is described below.
**No documentation update:** these changes add a security option and relocate expression expansion
within existing steps, with no user-facing surface.

### Testing

`task build` and `task format` were both run before pushing: **exit 0**, with `0 errors` from ESLint.
The pre-existing `no-explicit-any` warnings in `test/` are untouched by this YAML-only change.

With zizmor 1.30.0:

* Scanned before and after with `zizmor --no-config .github/workflows .github/actions` and diffed by
  rule: `artipacked` **27 → 0**, `template-injection` **146 → 104** (High **42 → 0**), total
  **196 → 126**. `cache-poisoning`, `adhoc-packages`, `unpinned-images` and `dangerous-triggers` are
  all unchanged, so the edits fixed exactly the intended classes and introduced nothing new.
* Parsed all 34 workflow and composite-action files with `js-yaml`. All parse, and walking the parsed
  documents finds exactly **27** `actions/checkout` steps with `with.persist-credentials === false`
  and **0** findings with `ident == template-injection && severity == High` — verified against the
  loaded YAML rather than by grepping text.
* Re-grepped the composite action for `${{` and confirmed every remaining occurrence sits in a safe
  position (`env:`, `if:`, `with:`, output `value:`), none inside a `run:` block.
* Audited all 23 checkout files for post-checkout git operations, `submodules:` and `extraheader`
  usage before applying the persist-credentials change, as described above.

The following was not tested:

* No workflow was executed. The persist-credentials change only removes a credential from the
  runner's git config, and the audit establishes that no step reads it — but `zxc-code-style.yaml`'s
  anonymous `git fetch origin` is the one call site whose behaviour depends on it, and is worth
  watching on this PR's first run. `zxc-e2e-test.yaml` and `flow-performance-test.yaml` run on pull
  requests and so should get real coverage from this PR's own CI; `flow-examples-test.yaml` and
  `flow-gcs-test.yaml` partly so.
* `flow-hugo-publish.yaml`, `flow-deploy-release-artifact.yaml`, `flow-one-shot-timing.yaml` and
  `flow-gcs-bucket-reaper.yaml` are release, publish and scheduled-maintenance paths this PR does not
  trigger; their checkouts and rewritten `Set Inputs` steps will first run outside PR CI, on merge to
  `main` or on a real release.


jan-milenkov (dismissed):

.github/actions/solo-image-cache/action.yaml
```diff
@@ -47,19 +47,23 @@ runs:
     - name: Build Cache Metadata
       id: cache-metadata
       shell: bash
+      env:
```
JeffreyDallas: This `env:` hoisting (and the rest of this file's changes) has nothing to do with `persist-credentials` / `artipacked` — it's the tier-4 template-injection fix from #5956, which already merged (into `zizmor-cleanup-tier3`, this PR's own head branch, rather than into `main`). It's showing up here because GitHub auto-retargeted this PR's base from the deleted `zizmor-cleanup-tier2` branch to `main`. See the Critical finding in the review summary — same thing shows up in `flow-deploy-release-artifact.yaml`, `flow-examples-test.yaml`, `flow-gcs-test.yaml`, `flow-hugo-publish.yaml`, `flow-one-shot-timing.yaml`, `flow-performance-test.yaml`, and `zxc-e2e-test.yaml`.

jan-milenkov (approved):

PR #5954, opened 2026-09-02
ci: auto-retry CI failures only when a transient error is verified

JeffreyDallas:
## Description

* TLDR: Make your life easier to `auto rerun failed test` due to infrastructure issue or network glitch.
* Adds `.github/workflows/zxf-transient-failure-retry.yaml`, a `workflow_run`-triggered workflow that reruns the failed jobs of a pull request workflow when, and only when, **every failed job** carries a known transient infrastructure signature, and leaves every other failure red.
* **A run mixing transient and genuine failures is not retried at all.** `rerun-failed-jobs` is a run-level endpoint: it reruns every failed job and its dependents and cannot target a subset, so retrying a mixed run would rerun the real regression too. Measured against the historical checkout-timeout runs, 7 of 8 are pure and still qualify; the one mixed run (`33410899316`, a checkout timeout alongside two genuine `Unit Tests` failures) is now correctly refused.
* **A cancelled or interrupted run is never retried.** An interrupted run concludes as `cancelled` at run level even when one of its jobs had already failed — run `33650236015` is the evidence, with a genuinely failed `Code Style` job. The gate requires a completed, `failure`-concluded, pull-request-triggered run, the check is restated inside the script so it survives a trigger change, and a run still holding a cancelled job is left alone because it has almost certainly been superseded by a newer commit.
* Watches the ten pull-request-triggered workflows: `PR Checks`, `Test Examples`, `Test Migration Process`, `Test GCS as Bucket Storage`, `Standard Runner One-Shot Deploy and Destroy`, `Windows One-Shot Deploy and Destroy`, `Install Validation`, `Install Validation (Podman Runtime)`, `Performance Test Solo Deployment`, and `CodeQL`.
* A job is retried only when a failed **step name** matches a rule *and* the **job log** confirms the signature, with a `forbidden` pattern to veto it and a cap of 3 total attempts per run:

| Rule | Step gate | Log must contain | Log must not contain |
|---|---|---|---|
| Checkout cannot reach the remote | `Checkout*` | `The action 'Checkout…' has timed out after N minutes` | `fatal:`, `RPC failed`, `index-pack failed`, `Repository not found`, `could not read Username` |
| WSL distribution download over budget | `Setup WSL2`, `Retry Setup WSL2`, `Install Ubuntu WSL distribution fallback` | `The action '…' has timed out after N minutes` | `error TS####` |
| kind cannot bind a leaked host port | any failed step | `port is already allocated`, `driver failed programming external connectivity` | `error TS####` |
| Distribution mirror rejected metadata | `Bootstrap Container` | `Failed to retrieve new repository metadata`, `Access to requested URL is forbidden`, `Curl error (22)` | `error TS####`, `Failed to run task "build:compile"` |

The rules come from reading 88 failed job logs across 149 failed runs from 2026-08-28 to 2026-09-02. Two findings shaped the design:

1. **Roughly 99% of failures in that window were genuine bugs, not flakes** (87 of 88 sampled jobs). A blanket retry policy would recover about 1% of failures while masking reproducible product bugs and burning roughly 13 self-hosted E2E runner slots per attempt. Retries are therefore opt-in per signature rather than opt-out.
2. **Log confirmation is required, not optional.** Solo's own unit tests assert on infrastructure error text, so *passing* test names contain `ImagePullBackOff`, `ENOSPC`, `connection reset by peer`, and `dial tcp … connect: connection refused`. A retry policy keyed on generic network wording would rerun **every** unit test failure indefinitely. Every rule's `required` pattern names the failing step so it cannot match a test fixture.

`Install grpcurl` is deliberately excluded despite looking like an infrastructure step: it failed deterministically because `go install …@latest` resolved grpcurl v1.9.4, which needs Go >= 1.25, against a pinned Go 1.22.3 with `GOTOOLCHAIN=local`. It recurred on attempts 1, 2 **and** 3 of run `33480747973` before #5934 pinned the version, so retrying it only wasted three full E2E matrices.

Two operational notes for reviewers:

* `workflow_run` triggers only take effect once the file is on the default branch, so this cannot be exercised from the PR branch and will start acting on the first failure after merge.
* `workflow_run` is on zizmor's dangerous-triggers list. This workflow never checks out pull request code and passes no pull-request-controlled value into a shell — it only reads job logs and calls the rerun API — but it may need an allow entry alongside the current zizmor remediation work.

### Related Issues

* Related to #5934
* Technical debt tracked separately: #5960 (checkout step timeout), #5961 (WSL2 setup step timeout)
* No issue filed for this automation; it came out of an audit of five days of CI failures

Notes on the unchecked boxes:

* **Tests** — the repository has no harness for executing workflow YAML, so the logic was verified by simulating the embedded script against real GitHub API payloads and real job logs (documented below) rather than by a committed test.
* **Documentation** — no user-facing behaviour changes; the rules and their rationale are documented in comments inside the workflow itself.
* **Technical debt** — two root-cause fixes are deliberately **not** included here because they change shared CI paths for every pull request, and are raised as their own issues instead: #5960 tracks the checkout step's 5-minute `STEP_TIMEOUT_MINUTES` budget in `zxc-e2e-test.yaml`, and #5961 tracks the WSL step's 3-minute budget being well below the roughly 9 minutes the ~1 GB rootfs download actually takes, so that step can essentially never pass on a cache miss and retrying it only delays the failure.

### Testing

* **Rule regexes against every retained log.** The four shipped rules were extracted directly from the workflow YAML and run against all 148 real job and run logs collected during the audit. The kind and mirror rules each match exactly 1 of 148 logs — the intended one — with no false positives. The checkout rule matches only checkout-timeout logs, and the WSL rule only the six WSL runs.
* **End-to-end simulation of the embedded script** with real `listJobsForWorkflowRun` payloads and real job logs, asserting the rerun decision in twelve scenarios. All twelve passed:

| Scenario | Run | Expected | Result |
|---|---|---|---|
| Checkout timeout in `PR Checks` | `33544214584` | rerun | rerun (3 jobs matched) |
| kind host-port collision | `33528665104` | rerun | rerun |
| grpcurl Go toolchain pin (deterministic) | `33480747973` | no rerun | no rerun |
| Genuine E2E product failures | `33376447381` | no rerun | no rerun |
| Unit test infra fixtures + Docs Automation | `33623537541` | no rerun | no rerun |
| Checkout timeout already at attempt 3 of 3 | `33544214584` | no rerun | no rerun |
| Transient failure mixed with a genuine regression | `33410899316` | no rerun | no rerun (1 of 3 failed jobs transient) |
| Cancelled run holding a genuinely failed job | `33650236015` | no rerun | no rerun (blocked at the gate) |
| Same cancelled run, gate bypassed to exercise the in-script guard | `33650236015` | no rerun | no rerun |
| Interrupted run reported as `failure` (synthetic cancelled job) | `33544214584` | no rerun | no rerun |
| Run still `in_progress` | `33544214584` | no rerun | no rerun |
| Non-pull-request (`push`) run | `33544214584` | no rerun | no rerun |

* **Static validation** — the YAML parses, the embedded `github-script` body parses as JavaScript, and the file is Prettier-clean. `task build` and `task format` both report clean with no tracked file changes.

The following was not tested:

* Live behaviour after merge — `workflow_run` does not fire for a workflow file that is not yet on the default branch, so the first real exercise will be the first matching failure after merge.
* An earlier iteration also retried the whole `Docs Automation` job on any failure. That was **removed**: all 7 sampled Docs Automation failures were genuine — `Command failed: solo --version (exit code 1)` on the very branch fixing that bug, the `TS1117` compile break, a lint gate failure, and a real relay regression — so retrying them would have masked bugs rather than recovering flakes.

<details>
<summary>
Commit message guidelines
</summary>

</details>

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.github/workflows/zxf-transient-failure-retry.yaml
```diff
@@ -0,0 +1,208 @@
...
+              {
+                description: 'checkout could not reach the remote',
+                stepPrefixes: ['checkout'],
+                required: /The action 'Checkout[^']*' has timed out after \d+ minutes?/,
+                forbidden: /fatal:|RPC failed|index-pack failed|Repository not found|could not read Username/,
+              },
+              {
+                description: 'WSL distribution download exceeded its budget',
+                stepPrefixes: ['setup wsl2', 'retry setup wsl2', 'install ubuntu wsl distribution fallback'],
+                required: /The action '(Setup WSL2|Retry Setup WSL2|Install Ubuntu WSL distribution fallback)' has timed out after \d+ minutes?/,
```
JeffreyDallas, line 102: `Retry Setup WSL2` and `Install Ubuntu WSL distribution fallback` do not exist. `grep -rn "Retry Setup WSL2\|Install Ubuntu WSL distribution fallback" .github/` at this SHA returns nothing, and `git diff --stat origin/main...aea7235 -- .github/` confirms this PR adds no workflow that would define them. The only WSL step in the repo is `Setup WSL2` at `flow-standard-runner-one-shot-windows.yaml:126`.

So two of the three prefixes and two of the three regex alternatives are dead, which makes the rule read like it covers a retry-and-fallback chain that was never built.

```suggestion
                stepPrefixes: ['setup wsl2'],
                required: /The action 'Setup WSL2' has timed out after \d+ minutes?/,
```

Separately — is this rule still worth shipping? #5934 (`ec11d629e`) gave that step `timeout-minutes: 10` and `use-cache: true` on 2026-09-01, so the 3-minute budget the PR description cites is no longer what is in `main`.

.github/workflows/zxf-transient-failure-retry.yaml
```diff
@@ -0,0 +1,208 @@
...
+
+            if (run.run_attempt >= maxRunAttempts) {
+              core.warning(
+                `Transient failures found (${reasons.join('; ')}), but run ${run.id} is already at ` +
+                `attempt ${run.run_attempt} of ${maxRunAttempts}; not rerunning. ${run.html_url}`
+              );
+              return;
+            }
+
+            await github.rest.actions.reRunWorkflowFailedJobs({ owner, repo, run_id: run.id });
```
JeffreyDallas, line 220: `reRunWorkflowFailedJobs` reruns **every** failed job in the run, not just the ones that matched a rule. The loop above already knows which jobs matched — it just throws the ids away at line 182.

On a mixed run (one checkout timeout plus a real E2E regression) this pulls the genuinely-failed jobs back onto the runners too, and does it again on attempt 3. That is the exact cost the `install grpcurl` exclusion was written to avoid, and run `33480747973` shows what the bill looks like: 12 E2E jobs, three attempts.

Roughly:

```js
// in the loop, alongside reasons.push(...)
matchedJobs.push(job);

// at the bottom
for (const job of matchedJobs) {
  await github.rest.actions.reRunJobForWorkflowRun({ owner, repo, job_id: job.id });
}
```

I confirmed `reRunJobForWorkflowRun` exists on the octokit that github-script v9 ships. What I have **not** confirmed is whether GitHub permits an individual rerun of a job that came from a called workflow — every `E2E Tests (X) / X` job is produced by `zxc-e2e-test.yaml` via `workflow_call`. If it does not, keeping `reRunWorkflowFailedJobs` is fine, but the comment block at lines 83-87 should say that the blast radius is the whole failed set rather than the matched job.

.github/workflows/zxf-transient-failure-retry.yaml
```diff
@@ -0,0 +1,208 @@
...
+        github.event.workflow_run.conclusion == 'failure'
+      }}
+    runs-on: hiero-solo-linux-medium
+    timeout-minutes: 15
+    permissions:
+      contents: read
+      actions: write # reads the observed run's job logs and reruns its failed jobs
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@05e31511f85b41b11d1cf0ef85d0992719546e2c # v2.21.0
```
JeffreyDallas: The other 39 `harden-runner` usages in this repo are all on v2.21.1 (`e14015d583714f6e62063499dc959a02595150a1`). Both SHAs resolve correctly against the upstream tags, so this is only a consistency and dependabot-noise thing.

```suggestion
        uses: step-security/harden-runner@e14015d583714f6e62063499dc959a02595150a1 # v2.21.1
```

.github/workflows/zxf-transient-failure-retry.yaml
```diff
@@ -0,0 +1,208 @@
...
+            // almost always means a newer commit superseded it. Retrying then spends runners on
+            // code that is already stale, so the whole run is left alone.
+            const cancelledJobs = jobs.filter(job => job.conclusion === 'cancelled');
+            if (cancelledJobs.length > 0) {
+              core.info(
+                `${run.name} run ${run.id} was interrupted (cancelled jobs: ` +
+                `${cancelledJobs.map(job => job.name).join(', ')}). Leaving the run alone.`
+              );
+              return;
+            }
```
JeffreyDallas: This guard never fires. I checked the 40 most recent failed `pull_request` runs in the repo — **zero** of them contain a job with conclusion `cancelled`. That is the expected outcome: eight of the ten watched workflows set `cancel-in-progress: true`, so a superseded run concludes `cancelled` and the `if:` at line 57 already drops it, and every matrix I looked at (`flow-pull-request-checks.yaml:116`, `zxc-unit-test.yaml:52`, `flow-examples-test.yaml:181`, `flow-install-validation.yaml:42`) sets `fail-fast: false`, so siblings are never cancelled either.

No objection to keeping the belt-and-braces check, but the comment claims a cause ("in this repo almost always means a newer commit superseded it") that the data does not show, and a future reader will trust it. Consider trimming it to what the code actually guarantees.

The two exceptions are `flow-install-validation.yaml` and `flow-install-validation-runtime.yaml`, which have no `concurrency` block at all — those two *can* reach the retry path on a stale commit. They run on `ubuntu-latest` so the cost is small, but it is the one case where a `run.head_sha` check would buy something.

.github/workflows/zxf-transient-failure-retry.yaml
```diff
@@ -0,0 +1,208 @@
...
+    # Only a run that completed and concluded in failure is eligible. A run that was cancelled -
+    # by `cancel-in-progress` when a newer commit arrives, or by hand - concludes as 'cancelled'
+    # even when one of its jobs had already failed, so it never reaches the retry logic.
+    if: >-
+      ${{
+        github.event.workflow_run.status == 'completed' &&
+        startsWith(github.event.workflow_run.event, 'pull_request') &&
+        github.event.workflow_run.conclusion == 'failure'
+      }}
+    runs-on: hiero-solo-linux-medium
```
JeffreyDallas: This job only makes API calls — list jobs, download logs, POST a rerun. Putting it on the self-hosted pool spends a `hiero-solo-linux-medium` slot on every failed PR run across ten workflows, which cuts against the runner-conservation argument the PR is built on. `flow-install-validation.yaml`, `flow-install-validation-runtime.yaml`, and `flow-deploy-release-artifact.yaml` already use the hosted runner.

```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/zxf-transient-failure-retry.yaml
```diff
@@ -0,0 +1,208 @@
...
+              },
+              {
+                description: 'WSL distribution download exceeded its budget',
+                stepPrefixes: ['setup wsl2', 'retry setup wsl2', 'install ubuntu wsl distribution fallback'],
+                required: /The action '(Setup WSL2|Retry Setup WSL2|Install Ubuntu WSL distribution fallback)' has timed out after \d+ minutes?/,
+                forbidden: /error TS\d{4}/,
+              },
+              {
+                description: 'kind could not bind a host port held by a leaked cluster',
+                stepPrefixes: [],
```
JeffreyDallas, line 107: This is the only rule with an empty `stepPrefixes`, which means every failed job in all ten workflows becomes a candidate and has its log downloaded. On run `33480747973` that is 12 E2E log downloads against a `timeout-minutes: 15` budget.

The cost is survivable, but the classification risk is the part I would think about: if Solo ever ships a genuine port-allocation regression that surfaces this string, this rule retries it to the cap on ~13 self-hosted jobs. The real log line has more context to key on:

```
docker: Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint solo-cluster-control-plane (...) 127.0.0.1:35211 failed: port is already allocated
```

Would requiring the daemon context — something like `/Error response from daemon:[\s\S]{0,200}(port is already allocated|driver failed programming external connectivity)/` — narrow it enough without losing the case you sampled?

.github/workflows/zxf-transient-failure-retry.yaml
```diff
@@ -0,0 +1,208 @@
...
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@05e31511f85b41b11d1cf0ef85d0992719546e2c # v2.21.0
+        with:
+          egress-policy: audit
+
+      - name: Rerun Failed Jobs With A Transient Signature
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          github-token: ${{ secrets.GH_ACCESS_TOKEN != '' && secrets.GH_ACCESS_TOKEN || github.token }}
```
JeffreyDallas: Question rather than a finding: does `workflow_run` re-fire after a rerun issued with `GITHUB_TOKEN`? GitHub suppresses recursive triggering for events created by the default token, and if that covers API-issued reruns then `MAX_RUN_ATTEMPTS: '3'` never binds and the effective cap is a single retry — not the "cap of 3 total attempts per run" the description promises.

Is the `GH_ACCESS_TOKEN` preference here specifically to get the re-fire? `actions: write` on `github.token` is sufficient for `reRunWorkflowFailedJobs` on its own, so if the PAT is not buying recursion I would rather not widen the blast radius of this workflow. Either way it is worth a comment on this line saying which it is, since it is the difference between one retry and two.

JeffreyDallas (commented):
@copilot resolve PR comment if reasonable , or respond with your argument

.github/workflows/zxf-transient-failure-retry.yaml
```diff
@@ -0,0 +1,224 @@
...
+jobs:
+  retry-transient-failure:
+    name: 'Retry Transient Failure'
+    # Only a completed run that concluded in failure is eligible. A cancelled or interrupted run
+    # does not reach the retry path, and a stale run is not retried once the job-level gate has
+    # already rejected it.
+    if: >-
+      ${{
+        github.event.workflow_run.status == 'completed' &&
+        startsWith(github.event.workflow_run.event, 'pull_request') &&
```
JeffreyDallas: Minor: `startsWith(github.event.workflow_run.event, 'pull_request')` is a prefix match, but every one of the 10 watched workflows triggers this bot only via a plain `pull_request` event (none use `pull_request_target`). An exact `===` comparison would be equivalent today and wouldn't quietly start matching a future `pull_request_target`-triggered workflow without that being a deliberate decision. Optional tightening, not a defect.

jan-milenkov (approved):

PR #5962, opened 2026-09-03
ci: cache verified Ubuntu WSL installer

### LINKED ISSUE #5936: Windows Deploy and Destroy > The action 'Setup WSL2' has timed out after 10 minutes.
## Failure Summary

The **Windows Deploy and Destroy** job failed on branch `fix-pin-grpcurl-version`. Error: The action 'Setup WSL2' has timed out after 10 minutes.

The step was killed by its timeout (`exit status 124`) rather than failing with a
thrown error, so the pod/container state below is the root-cause evidence — any
`ERROR`-level lines logged before the kill are downstream symptoms.

## Error Details

```
2026-09-01T13:27:54.8418936Z ##[error]The action 'Setup WSL2' has timed out after 10 minutes.
```

## Workflow Information

- **Run**: [#3757](https://github.com/hiero-ledger/solo/actions/runs/33512322943) (attempt 1)
- **Job**: [Windows Deploy and Destroy](https://github.com/hiero-ledger/solo/actions/runs/33512322943/job/99870776596)
- **Branch**: `fix-pin-grpcurl-version`

## Log Archives

- [CI Logs (secret gist)](https://gist.github.com/JeffreyDallas/79851248f1bfd3336403faaf6320818e) — 3 files (job log + complete `standard-runner-one-shot-windows-logs` artifact)

<details>
<summary>All files</summary>

- [job-99870776596.log](https://gist.githubusercontent.com/JeffreyDallas/79851248f1bfd3336403faaf6320818e/raw/job-99870776596.log)
- [solo.log](https://gist.githubusercontent.com/JeffreyDallas/79851248f1bfd3336403faaf6320818e/raw/solo.log)
- [solo.ndjson](https://gist.githubusercontent.com/JeffreyDallas/79851248f1bfd3336403faaf6320818e/raw/solo.ndjson)

</details>

JeffreyDallas:
## Description

This PR makes the Windows WSL bootstrap resilient to intermittent Ubuntu-distribution acquisition failures without allowing cache availability to turn the Windows compatibility check green.

### Root-cause clarification

An unclear or unstable runner state is **not** the root cause of the Ubuntu-distribution acquisition failure. The failed job diagnostics showed a healthy WSL service, no registered Ubuntu distribution, and no residual WSL process. The timeout occurred in the distribution-acquisition path before the installer was available.

### Current cache behavior

| Condition | Installer handling | Cache write | Test result |
| --- | --- | --- | --- |
| Valid cache, any ref | Hash-validate and add the required `x64.complete` marker | None | Run the full WSL integration test |
| Cache miss or invalid installer on `main` | Direct-download and SHA-256 verify | Save the content-addressed installer | Run the full WSL integration test |
| Cache miss or invalid installer on a pull request | Direct-download and SHA-256 verify | Skip save | Run the full WSL integration test |

### Changes

* Uses a content-addressed installer cache key: Ubuntu version plus expected SHA-256.
* Hash-validates restored installers and creates the sibling `x64.complete` marker required by the setup action. It then installs verified local bytes before reaching the action internal cache or download path.
* Keeps the direct SHA-verified download as self-healing for cache misses and invalid cache contents. A miss cannot skip the test or report a false green result.
* Limits cache publishing to `main`, so pull-request runs never create private 372 MB cache entries.
* Sets `use-cache: false` on the setup action, avoiding a duplicate cache restore.
* Removes post-setup recovery retries. A setup failure captures bounded diagnostics and fails the job promptly.
* Extends the project-build timeout to ten minutes so a slow dependency installation does not consume the compilation budget.

### Related Issues

* Closes #5936

### Pull request checklist

* [x] The commit title follows the project guidelines

### Testing

* [x] `task build`
* [x] `npx prettier --check .github/workflows/flow-standard-runner-one-shot-windows.yaml`
* [x] YAML parse validation
* [x] [Marker validation](https://github.com/hiero-ledger/solo/actions/runs/33830097179/job/100900653361) passed end to end: one restore, hash validation, and no second setup-action restore.
* [x] [Content-addressed cache seed](https://github.com/hiero-ledger/solo/actions/runs/33873624502/job/101025237385) passed end to end: cache miss, direct SHA-verified seed, primary WSL setup, and full deploy/destroy test.
* [x] [Current workflow attempt 1](https://github.com/hiero-ledger/solo/actions/runs/33877470289/job/101037774963), [attempt 2](https://github.com/hiero-ledger/solo/actions/runs/33877470289/job/101041914071), and [attempt 3](https://github.com/hiero-ledger/solo/actions/runs/33877470289/job/101046634081) all passed end to end. Each restored and hash-validated the installer, skipped seed/save on the PR, then completed WSL setup, deployment, connectivity, and destroy.

### DIFF
--- a/.github/workflows/flow-standard-runner-one-shot-windows.yaml
+++ b/.github/workflows/flow-standard-runner-one-shot-windows.yaml
@@ -231,11 +231,12 @@ jobs:
           $available = '${{ steps.validate_ubuntu_wsl_installer.outputs.valid }}' -eq 'true' -or '${{ steps.seed_ubuntu_wsl_installer.outputs.valid }}' -eq 'true'
           Add-Content -Path $env:GITHUB_OUTPUT -Value "available=$($available.ToString().ToLowerInvariant())"
           if (-not $available) {
-            @"
-            ## Windows WSL integration skipped
-
-            The verified Ubuntu installer cache was unavailable or invalid. This PR did not attempt online Ubuntu acquisition. A `main` run will seed a missing cache; an invalid cache is deleted and replaced on `main`.
-            "@ | Add-Content -Path $env:GITHUB_STEP_SUMMARY
+            $summary = @(
+              '## Windows WSL integration skipped'
+              ''
+              'The verified Ubuntu installer cache was unavailable or invalid. This PR did not attempt online Ubuntu acquisition. A main run will seed a missing cache; an invalid cache is deleted and replaced on main.'
+            ) -join [Environment]::NewLine
+            Add-Content -Path $env:GITHUB_STEP_SUMMARY -Value $summary
             Write-Warning 'Skipping WSL-dependent integration steps because the verified Ubuntu installer cache is unavailable.'
           }
 
--- a/.github/workflows/flow-standard-runner-one-shot-windows.yaml
+++ b/.github/workflows/flow-standard-runner-one-shot-windows.yaml
@@ -309,7 +309,6 @@ jobs:
         # Bypass both setup-wsl's downloader and WSL's online downloader when the validated
         # repository cache is available. Retry Setup WSL2 still runs afterwards to configure
         # the distribution's packages and the wsl-bash shell wrapper.
-        continue-on-error: true
         timeout-minutes: 3
         shell: pwsh
         run: |
@@ -341,9 +340,11 @@ jobs:
 
       - name: Retry Setup WSL2
         id: retry_setup_wsl2
-        if: ${{ !cancelled() && steps.setup_wsl2.outcome == 'failure' }}
-        # Run after the cached fallback even if it failed. Without !cancelled(), GitHub's
-        # implicit success() check skips this retry after the fallback fails.
+        if: ${{ !cancelled() && steps.setup_wsl2.outcome == 'failure' && steps.install_cached_ubuntu_wsl.outcome == 'success' }}
+        # Only retry after the cached fallback installed the distribution. This explicit
+        # condition bypasses GitHub's implicit success() check after the first setup failure.
+        # A failed cached fallback stops the job rather than allowing this action to download
+        # Ubuntu online again.
         uses: Vampire/setup-wsl@d1da7f2c0322a5ee4f24975344f67fc0f5baf364 # v7.0.0
         timeout-minutes: 5
         with:
--- a/.github/workflows/flow-standard-runner-one-shot-windows.yaml
+++ b/.github/workflows/flow-standard-runner-one-shot-windows.yaml
@@ -204,7 +204,7 @@ jobs:
           Write-Host "Seeded verified Ubuntu WSL installer cache: $installerPath"
 
       - name: Save seeded Ubuntu WSL installer cache
-        if: ${{ success() && steps.seed_ubuntu_wsl_installer.outputs.valid == 'true' }}
+        if: ${{ success() && github.ref == 'refs/heads/main' && steps.seed_ubuntu_wsl_installer.outputs.valid == 'true' }}
         uses: actions/cache/save@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
         with:
           path: ${{ runner.tool_cache }}\Ubuntu\${{ env.WSL_UBUNTU_VERSION }}\x64
--- a/.github/workflows/flow-standard-runner-one-shot-windows.yaml
+++ b/.github/workflows/flow-standard-runner-one-shot-windows.yaml
@@ -45,6 +45,9 @@ env:
   DOCKER_HOST: tcp://localhost:2375
   ACTIONS_STEP_DEBUG: 'true'
   ACTIONS_RUNNER_DEBUG: 'true'
+  WSL_UBUNTU_VERSION: '24.4.4'
+  WSL_UBUNTU_INSTALLER_FILE: 'ubuntu-24.04.4-wsl-amd64.wsl'
+  WSL_UBUNTU_INSTALLER_SHA256: '9b2f7730dc68227dd04a9f3e5eab86ad85caf556b8606ad94f1f29ff5c4fd3f5'
 
 jobs:
   standard-runner-one-shot-windows:
@@ -101,7 +104,7 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         timeout-minutes: ${{ fromJSON(env.STEP_TIMEOUT_MINUTES) }}
 
-      - name: Setup Node.js 22
+      - name: Setup Node.js 24
         uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         timeout-minutes: ${{ fromJSON(env.STEP_TIMEOUT_MINUTES) }}
         with:
@@ -120,36 +123,63 @@ jobs:
         run: |
           task build:compile
 
-      - name: Validate Ubuntu WSL cache
+      - name: Restore verified Ubuntu WSL installer cache
+        id: restore_ubuntu_wsl_installer
+        uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: ${{ runner.tool_cache }}\Ubuntu\${{ env.WSL_UBUNTU_VERSION }}\x64
+          key: solo-wsl-installer-${{ runner.os }}-${{ runner.arch }}-${{ env.WSL_UBUNTU_VERSION }}-${{ env.WSL_UBUNTU_INSTALLER_SHA256 }}
+
+      - name: Validate restored Ubuntu WSL installer cache
+        id: validate_ubuntu_wsl_installer
         shell: pwsh
         timeout-minutes: 1
         run: |
-          $cacheDirectory = Join-Path $env:RUNNER_TOOL_CACHE 'Ubuntu\24.4.4\x64'
-          $installerPath = Join-Path $cacheDirectory 'ubuntu-24.04.4-wsl-amd64.wsl'
+          $cacheDirectory = Join-Path $env:RUNNER_TOOL_CACHE "Ubuntu\$env:WSL_UBUNTU_VERSION\x64"
+          $installerPath = Join-Path $cacheDirectory $env:WSL_UBUNTU_INSTALLER_FILE
           $completeMarker = Join-Path $cacheDirectory '.complete'
-          $valid = (
-            (Test-Path -LiteralPath $installerPath) -and
-            (Test-Path -LiteralPath $completeMarker) -and
-            ((Get-Item -LiteralPath $installerPath).Length -gt 100MB)
-          )
+          $cacheHit = '${{ steps.restore_ubuntu_wsl_installer.outputs.cache-hit }}'
+          $valid = $false
+          $validationResult = 'cache miss'
 
-          if ($valid) {
-            Write-Host "Ubuntu WSL cache is valid: $installerPath"
+          if ((Test-Path -LiteralPath $installerPath) -and (Test-Path -LiteralPath $completeMarker)) {
+            $actualHash = (Get-FileHash -LiteralPath $installerPath -Algorithm SHA256).Hash.ToLowerInvariant()
+            if ($actualHash -eq $env:WSL_UBUNTU_INSTALLER_SHA256) {
+              $valid = $true
+              $validationResult = 'valid SHA-256'
+              Write-Host "Verified Ubuntu WSL installer cache: $installerPath"
+            } else {
+              $validationResult = "invalid SHA-256: $actualHash"
+            }
           } elseif (Test-Path -LiteralPath $cacheDirectory) {
-            Write-Warning "Removing incomplete Ubuntu WSL cache: $cacheDirectory"
-            Remove-Item -LiteralPath $cacheDirectory -Recurse -Force
-          } else {
-            Write-Host "Ubuntu WSL cache is not present; setup-wsl will download it."
+            $validationResult = 'installer or completion marker is missing'
           }
+
+          if (-not $valid -and (Test-Path -LiteralPath $cacheDirectory)) {
+            Write-Warning "Discarding invalid Ubuntu WSL installer cache: $validationResult"
+            Remove-Item -LiteralPath $cacheDirectory -Recurse -Force
+          }
+
+          if (-not $valid -and $cacheHit -eq 'true') {
+            throw "Restored Ubuntu WSL installer cache failed validation: $validationResult"
+          }
+
+          Add-Content -Path $env:GITHUB_OUTPUT -Value "valid=$($valid.ToString().ToLowerInvariant())"
+          @"
+          ## Ubuntu WSL installer cache
+
+          - GitHub Actions cache hit: $cacheHit
+          - Validation: $validationResult
+          "@ | Add-Content -Path $env:GITHUB_STEP_SUMMARY
 
       - name: Setup WSL2
         id: setup_wsl2
         # Kind requires Linux containers. GitHub's windows-latest runner has Docker in Windows
         # container mode with no Docker Desktop to switch it. WSL2 + Docker inside it replicates
         # what Docker Desktop does for Windows users, exposing a Linux daemon via TCP.
-        # use-cache is enabled so the distribution installer is cached across runs via
-        # actions/cache instead of re-downloaded every time — this workflow was intermittently
-        # timing out on the fresh download from a shared runner's network.
+        # The action cache restores the installer across fresh GitHub-hosted runner VMs.
+        # The verified repository cache above restores the same installer before this action
+        # touches its own tool-cache lookup, avoiding an Ubuntu download on the common path.
         uses: Vampire/setup-wsl@d1da7f2c0322a5ee4f24975344f67fc0f5baf364 # v7.0.0
         continue-on-error: true
         # A healthy setup completes in about one minute. Bound a stalled action so the
@@ -158,8 +188,7 @@ jobs:
         with:
           distribution: Ubuntu-24.04
           wsl-version: 2
-          # The setup-wsl cache has hung on hosted runners while resolving this distribution.
-          use-cache: false
+          use-cache: true
           set-as-default: true
           additional-packages: |
             apt-transport-https
@@ -257,9 +286,51 @@ jobs:
           }
           exit 0
 
-      - name: Install Ubuntu WSL distribution fallback
-        if: ${{ steps.setup_wsl2.outcome == 'failure' }}
-        # Bypass setup-wsl's downloadTool path when the hosted tool cache is incomplete.
+      - name: Install cached Ubuntu WSL distribution fallback
+        id: install_cached_ubuntu_wsl
+        if: ${{ !cancelled() && steps.setup_wsl2.outcome == 'failure' }}
+        # Bypass both setup-wsl's downloader and WSL's online downloader when the validated
+        # repository cache is available. Retry Setup WSL2 still runs afterwards to configure
+        # the distribution's packages and the wsl-bash shell wrapper.
+        continue-on-error: true
+        timeout-minutes: 3
+        shell: pwsh
+        run: |
+          $distributionName = 'Ubuntu-24.04'
+          $cacheDirectory = Join-Path $env:RUNNER_TOOL_CACHE "Ubuntu\$env:WSL_UBUNTU_VERSION\x64"
+          $installerPath = Join-Path $cacheDirectory $env:WSL_UBUNTU_INSTALLER_FILE
+          $completeMarker = Join-Path $cacheDirectory '.complete'
+
+          if (-not (Test-Path -LiteralPath $installerPath) -or -not (Test-Path -LiteralPath $completeMarker)) {
+            Write-Host 'Validated Ubuntu WSL installer is unavailable; continuing to web fallback.'
+            Add-Content -Path $env:GITHUB_OUTPUT -Value 'installed=false'
+            exit 0
+          }
+
+          $actualHash = (Get-FileHash -LiteralPath $installerPath -Algorithm SHA256).Hash.ToLowerInvariant()
+          if ($actualHash -ne $env:WSL_UBUNTU_INSTALLER_SHA256) {
+            throw "Cached Ubuntu WSL installer SHA-256 mismatch: $actualHash"
+          }
+
+          $installedDistributions = @(wsl.exe --list --quiet 2>$null)
+          if ($installedDistributions -match [regex]::Escape($distributionName)) {
+            Write-Host "WSL distribution already installed from the validated cache: $distributionName"
+          } else {
+            Write-Host "Installing $distributionName from the validated repository cache"
+            & "$env:SystemRoot\System32\wsl.exe" --install --from-file $installerPath --no-launch --name $distributionName
+            if ($LASTEXITCODE -ne 0) {
+              throw "wsl.exe failed to install $distributionName from the validated cache with exit code $LASTEXITCODE"
+            }
+          }
+
+          Add-Content -Path $env:GITHUB_OUTPUT -Value 'installed=true'
+
+      - name: Install Ubuntu WSL distribution web fallback
+        id: install_web_ubuntu_wsl
+        if: ${{ !cancelled() && steps.setup_wsl2.outcome == 'failure' && steps.install_cached_ubuntu_wsl.outputs.installed != 'true' }}
+        # A last-resort path for a cache miss. Marking it continue-on-error lets the bounded
+        # retry below decide the job outcome instead of suppressing the retry after this fails.
+        continue-on-error: true
         timeout-minutes: 5
         shell: pwsh
         run: |
@@ -277,8 +348,10 @@ jobs:
           }
 
       - name: Retry Setup WSL2
-        if: ${{ steps.setup_wsl2.outcome == 'failure' }}
-        # Retry once after removing the process and distribution left by the timed-out setup action.
+        id: retry_setup_wsl2
+        if: ${{ !cancelled() && steps.setup_wsl2.outcome == 'failure' }}
+        # Run after either fallback even if it failed. Without !cancelled(), GitHub's implicit
+        # success() check skips this retry after a failed web-download fallback.
         uses: Vampire/setup-wsl@d1da7f2c0322a5ee4f24975344f67fc0f5baf364 # v7.0.0
         timeout-minutes: 5
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -45,6 +45,11 @@ env:
   DOCKER_HOST: tcp://localhost:2375
   ACTIONS_STEP_DEBUG: 'true'
   ACTIONS_RUNNER_DEBUG: 'true'
+  WSL_UBUNTU_VERSION: '24.4.4'
+  WSL_UBUNTU_INSTALLER_FILE: 'ubuntu-24.04.4-wsl-amd64.wsl'
+  WSL_UBUNTU_INSTALLER_SHA256: '9b2f7730dc68227dd04a9f3e5eab86ad85caf556b8606ad94f1f29ff5c4fd3f5'
```
JeffreyDallas: I verified this against `https://releases.ubuntu.com/24.04.4/SHA256SUMS` and it matches. Worth recording where it came from so the next person bumping Ubuntu knows how to re-derive it rather than copying it out of a run log.

```suggestion
  # SHA-256 from https://releases.ubuntu.com/24.04.4/SHA256SUMS
  WSL_UBUNTU_INSTALLER_SHA256: '9b2f7730dc68227dd04a9f3e5eab86ad85caf556b8606ad94f1f29ff5c4fd3f5'
```

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -45,6 +45,11 @@ env:
...
+  # Cache key used by the pinned Vampire/setup-wsl action for Ubuntu 24.04.4.
+  WSL_UBUNTU_CACHE_KEY: '2:distributionDirectory_Ubuntu_24.4.4'
```
JeffreyDallas: This key is an internal implementation detail of the pinned action, and `.github/dependabot.yml:3-13` bumps all actions daily and grouped. When a bump moves the Ubuntu version, the key changes, this restore misses, and the fail-early step below hard-fails every non-main run — including Dependabot's own PR. Pointing at the two upstream sources makes that a two-minute fix instead of an archaeology session.

```suggestion
  # Cache key used by the pinned Vampire/setup-wsl action:
  # "2:distributionDirectory_${distributionName}_${version}" (SetupWsl.kt), where the version is
  # hardcoded per distribution (Distribution.kt: Ubuntu-24.04 -> SemVer("24.4.4")).
  # Bumping Vampire/setup-wsl requires re-deriving all four WSL_UBUNTU_* values in the same PR.
  WSL_UBUNTU_CACHE_KEY: '2:distributionDirectory_Ubuntu_24.4.4'
```

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -120,46 +125,76 @@ jobs:
...
+          $cacheHit = '${{ steps.restore_ubuntu_wsl_installer.outputs.cache-hit }}'
+          $valid = $false
+          $validationResult = 'cache miss'
+
+          if (Test-Path -LiteralPath $installerPath) {
+            $actualHash = (Get-FileHash -LiteralPath $installerPath -Algorithm SHA256).Hash.ToLowerInvariant()
+            if ($actualHash -eq $env:WSL_UBUNTU_INSTALLER_SHA256) {
+              $valid = $true
+              $validationResult = 'valid SHA-256'
+              Write-Host "Verified Ubuntu WSL installer cache: $installerPath"
```
JeffreyDallas: **Major — this restore is currently thrown away and re-downloaded.** `@actions/tool-cache.find()` needs the sibling completion marker `<tool_cache>/Ubuntu/24.4.4/x64.complete`, which is not inside the cached `x64` directory. So `SetupWsl.kt:179` misses, `SetupWsl.kt:185` calls `cacheDir(...)` → `_createToolPath()` → `rmRF()` on the directory you just restored, and `SetupWsl.kt:194` re-restores the same 372 MB entry. Visible in the fail-fast run:

```
01:00:03.4  Cache hit for: 2:distributionDirectory_Ubuntu_24.4.4   <- this step
01:00:12.8  Cache hit for: 2:distributionDirectory_Ubuntu_24.4.4   <- setup-wsl, again
```

Writing the marker after the hash check makes `find()` hit, so the action returns immediately at `SetupWsl.kt:182` — no second restore, and the bytes you verified are the exact bytes `wsl --install --from-file` consumes rather than a discarded copy. `"$cacheDirectory.complete"` expands correctly in PowerShell (property access in a string needs `$(...)`, so the `.complete` is literal).

Keep `use-cache: true` on the action as the miss-path fallback; it just stops being exercised on the happy path.

```suggestion
              Write-Host "Verified Ubuntu WSL installer cache: $installerPath"
              # Publish the @actions/tool-cache completion marker for the verified directory so
              # setup-wsl's find() hits it. Without the marker it deletes this directory and
              # re-downloads the same cache entry through its own restoreCache call.
              New-Item -ItemType File -Path "$cacheDirectory.complete" -Force | Out-Null
```

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -120,46 +125,76 @@ jobs:
...
+          Add-Content -Path $env:GITHUB_OUTPUT -Value "valid=$($valid.ToString().ToLowerInvariant())"
+          @"
+          ## Ubuntu WSL installer cache
+
+          - GitHub Actions cache hit: $cacheHit
+          - Validation: $validationResult
+          "@ | Add-Content -Path $env:GITHUB_STEP_SUMMARY
+
+      - name: Fail early when verified Ubuntu WSL installer cache is unavailable
+        if: ${{ !cancelled() && github.ref != 'refs/heads/main' && steps.validate_ubuntu_wsl_installer.outputs.valid != 'true' }}
```
JeffreyDallas: Two things here.

**Policy (Major).** A plain cache miss now hard-fails every PR at once. The entry is a single 372 MB blob (`ref=refs/heads/main`, so it is visible repo-wide — that part is right), but `gh api repos/hiero-ledger/solo/actions/cache/usage` currently reports `active_caches_size_in_bytes: 10591348250` across 206 entries, i.e. ~10.6 GB against a 10 GB budget, so the repo is continuously evicting. Every PR run refreshes the access time, which protects it under steady traffic — but one quiet stretch and every open PR's Windows check goes red until someone pushes to main. Consider hard-failing only on *hit-but-invalid* (which the previous step already throws on) and letting a plain miss fall through to `setup-wsl`'s normal download, bounded by the existing `timeout-minutes: 3`.

**Condition (Minor).** `!cancelled()` drops the implicit `success()`, so this step also fires after unrelated earlier failures (e.g. `task build:compile`) and double-reports whenever the validation step already threw. The default is what you want:

```suggestion
        if: ${{ github.ref != 'refs/heads/main' && steps.validate_ubuntu_wsl_installer.outputs.valid != 'true' }}
```

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -120,46 +125,76 @@ jobs:
...
+        shell: pwsh
+        run: |
+          throw 'Verified Ubuntu WSL installer cache is unavailable. Online Ubuntu acquisition is disabled on non-main runs; rerun after a main workflow seeds the setup-wsl cache.'
```
JeffreyDallas: "rerun after a main workflow seeds the setup-wsl cache" only works for the eviction case. Cache keys are immutable and `@actions/cache.saveCache` swallows the duplicate-key failure as an `info` log (bundled `index.mjs:87729`, `:87739-87741`), so if the entry exists but fails validation, a main run downloads fresh, goes green, silently fails to save, and every PR keeps failing forever. Same for a `setup-wsl` bump that moves the Ubuntu version. Both need `gh cache delete`, not a rerun.

```suggestion
          throw "Verified Ubuntu WSL installer cache is unavailable ($env:WSL_UBUNTU_CACHE_KEY). Online Ubuntu acquisition is disabled on non-main runs. If the entry was evicted, a push to main reseeds it. If it exists but failed validation, or Vampire/setup-wsl was bumped to a different Ubuntu version, a main run cannot overwrite the immutable key -- delete it first with: gh cache delete '$env:WSL_UBUNTU_CACHE_KEY' -R hiero-ledger/solo"
```

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -120,46 +125,76 @@ jobs:
...
         id: setup_wsl2
         # Kind requires Linux containers. GitHub's windows-latest runner has Docker in Windows
         # container mode with no Docker Desktop to switch it. WSL2 + Docker inside it replicates
         # what Docker Desktop does for Windows users, exposing a Linux daemon via TCP.
-        # use-cache is enabled so the distribution installer is cached across runs via
-        # actions/cache instead of re-downloaded every time — this workflow was intermittently
-        # timing out on the fresh download from a shared runner's network.
+        # The action cache restores the installer across fresh GitHub-hosted runner VMs.
+        # The verified restore above uses the pinned action's cache key so this action finds
+        # the installer in its tool cache before it considers an Ubuntu download.
```
JeffreyDallas: This claim does not hold at the pinned SHA: the action never finds the installer in its tool cache (no `.complete` marker), and `SetupWsl.kt:185` deletes the restored directory before its own `restoreCache` re-fetches the entry. Please rewrite once you have decided on the marker suggestion in the validation step — if you take it, the sentence becomes true; if not, the comment should say that the restore exists purely to gate the run on a SHA-256 check and that the action restores its own copy.

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -120,46 +125,76 @@ jobs:
...
         uses: Vampire/setup-wsl@d1da7f2c0322a5ee4f24975344f67fc0f5baf364 # v7.0.0
-        continue-on-error: true
-        # A healthy setup completes in about one minute. Bound a stalled action so the
-        # recovery path can run instead of consuming the previous ten-minute timeout.
+        # A healthy setup completes in about one minute. Bound a stalled action so a
+        # distribution problem is reported promptly instead of consuming ten minutes.
         timeout-minutes: 3
```
JeffreyDallas: Not changed by this PR, but its meaning changed: with `continue-on-error` and the retry gone, breaching this bound now fails the job outright instead of falling through to recovery. The fail-fast run's healthy path was 84 s (`01:00:11.9` → `01:01:35.8`), so there is ~2x headroom — but `apt-get update` against `archive.ubuntu.com` sits inside that window, and the action itself wraps `wsl --update` in `retry(10)` on hosted `win22` runners, which implies upstream slowness is expected. 5 minutes costs nothing on the failure path and removes a class of false failure. Deliberate at 3?

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -116,49 +123,137 @@ jobs:
...
+        shell: pwsh
+        run: |
+          $available = '${{ steps.validate_ubuntu_wsl_installer.outputs.valid }}' -eq 'true' -or '${{ steps.seed_ubuntu_wsl_installer.outputs.valid }}' -eq 'true'
+          Add-Content -Path $env:GITHUB_OUTPUT -Value "available=$($available.ToString().ToLowerInvariant())"
+          if (-not $available) {
+            @"
+            ## Windows WSL integration skipped
+
+            The verified Ubuntu installer cache was unavailable or invalid. This PR did not attempt online Ubuntu acquisition. A `main` run will seed a missing cache; an invalid cache is deleted and replaced on `main`.
+            "@ | Add-Content -Path $env:GITHUB_STEP_SUMMARY
```
JeffreyDallas: **Critical — this fails to parse, so the step fails on every run.** From run [33829683914](https://github.com/hiero-ledger/solo/actions/runs/33829683914):

```
ParserError: ...ps1:9
   9 |    "@ | Add-Content -Path $env:GITHUB_STEP_SUMMARY
     | White space is not allowed before the string terminator.
```

PowerShell requires `"@` to be the first thing on its line. The here-string at `:179-184` gets away with it because both delimiters sit at the block scalar's base indent; inside this `if` block they cannot. It is a parse-time error, so the branch never has to be taken — the run above had a perfectly valid cache and still died here.

A string array avoids the rule entirely, and single quotes also stop the backtick escape from eating the `` `main` `` code formatting (in a double-quoted here-string, `` ` `` is the escape character, so `` `main` `` renders as bare `main`).

```suggestion
            @(
              '## Windows WSL integration skipped',
              '',
              'The verified Ubuntu installer cache was unavailable or invalid. This PR did not attempt online Ubuntu acquisition. A `main` run will seed a missing cache; an invalid cache is deleted and replaced on `main`.'
            ) | Add-Content -Path $env:GITHUB_STEP_SUMMARY
```

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -116,49 +123,137 @@ jobs:
...
+            Write-Warning 'Skipping WSL-dependent integration steps because the verified Ubuntu installer cache is unavailable.'
           }
 
       - name: Setup WSL2
         id: setup_wsl2
+        if: ${{ success() && steps.ubuntu_wsl_installer.outputs.available == 'true' }}
```
JeffreyDallas: **Major — this is where the gate stops gating.** With `available != 'true'` this step and the eight below it skip, every remaining step succeeds, and the check reports **green** having verified nothing on Windows. `:18-22` still promises "Any PR that breaks Solo on Windows will cause this workflow to fail".

A `Write-Warning` inside a passing job is not a signal anyone acts on, and the trigger is realistic: the repo is at `active_caches_size_in_bytes: 10591348250` across 206 entries against a 10 GB budget, so this single 372 MB entry can be evicted at any time. Last revision that was a loud failure; this one is a silent lapse in coverage.

You already have the fix in this PR. Drop `github.ref == 'refs/heads/main'` from `Seed verified Ubuntu WSL installer cache` at `:198` so a miss self-heals on any branch — it is a direct, hash-verified download from `releases.ubuntu.com`, not the `setup-wsl` path that stalled in #5936, and `timeout-minutes: 10` already bounds it. Then `available` is always true and this condition, the report step at `:226-240`, and the eight other gates all disappear.

Keep the ref check on `Save seeded Ubuntu WSL installer cache` at `:220` so PR branches do not each write a 372 MB branch-scoped copy into an already-full cache budget.

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -45,6 +46,12 @@ env:
   DOCKER_HOST: tcp://localhost:2375
   ACTIONS_STEP_DEBUG: 'true'
   ACTIONS_RUNNER_DEBUG: 'true'
+  WSL_UBUNTU_VERSION: '24.4.4'
+  WSL_UBUNTU_INSTALLER_FILE: 'ubuntu-24.04.4-wsl-amd64.wsl'
+  WSL_UBUNTU_INSTALLER_URL: 'https://releases.ubuntu.com/24.04.4/ubuntu-24.04.4-wsl-amd64.wsl'
+  WSL_UBUNTU_INSTALLER_SHA256: '9b2f7730dc68227dd04a9f3e5eab86ad85caf556b8606ad94f1f29ff5c4fd3f5'
+  # Cache key and tool-cache path used by setup-wsl for Ubuntu 24.04.4.
+  WSL_UBUNTU_CACHE_KEY: '2:distributionDirectory_Ubuntu_24.4.4'
```
JeffreyDallas: **Major — this no longer needs to be setup-wsl's key.** `use-cache: false` at `:257` means the action never calls `restoreCache`/`saveCache` (`SetupWsl.kt:194`, `:233`, both guarded by `useCache`), and with the marker in place it returns at `SetupWsl.kt:182` before either. The workflow is now the only reader and writer of this entry, so the borrowed key buys nothing.

A content-addressed key removes the invalid-cache state by construction — a version or hash change rotates the key instead of poisoning it. No suggestion block because it cannot be expressed here: the `env` context is **not** available inside a workflow-level `env:` map (only `github`, `vars`, `inputs`), so `WSL_UBUNTU_CACHE_KEY` cannot interpolate its siblings — it would silently resolve to `ubuntu-wsl-installer--`. Compose it at the two use sites instead, where `env` *is* available (as `:136` and `:223` already do for the path):

```yaml
# :137 and :224
key: ubuntu-wsl-installer-${{ env.WSL_UBUNTU_VERSION }}-${{ env.WSL_UBUNTU_INSTALLER_SHA256 }}
```

and drop `WSL_UBUNTU_CACHE_KEY` here. With that, `Delete invalid Ubuntu WSL installer cache from main` (`:186-194`, its only other consumer) and `permissions: actions: write` (`:37`) both become unnecessary — worth removing, since that permission currently applies to a job that builds and runs PR code.

One consequence to expect: the switch abandons the existing main-scoped entry, so the first main run after merge re-seeds once (~1 min through the seed path).

The **path** coupling at `:49` is the real one and stays — `WSL_UBUNTU_VERSION` must equal `distribution.version` from `Distribution.kt:183-188`. Worth a note at `:250` that bumping `setup-wsl` requires re-deriving it, since `.github/dependabot.yml:3-13` will open that bump automatically and a mismatch degrades silently to a full uncached download rather than failing.

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -116,49 +123,137 @@ jobs:
...
+          Add-Content -Path $env:GITHUB_OUTPUT -Value "valid=$($valid.ToString().ToLowerInvariant())"
+          @"
+          ## Ubuntu WSL installer cache
+
+          - GitHub Actions cache hit: $cacheHit
+          - Validation: $validationResult
+          "@ | Add-Content -Path $env:GITHUB_STEP_SUMMARY
+
+      - name: Delete invalid Ubuntu WSL installer cache from main
+        if: ${{ success() && github.ref == 'refs/heads/main' && steps.restore_ubuntu_wsl_installer.outputs.cache-hit == 'true' && steps.validate_ubuntu_wsl_installer.outputs.valid != 'true' }}
```
JeffreyDallas: Correct as written — `[uri]::EscapeDataString` on both params, `GH_TOKEN` from `github.token`, and the `cache-hit == 'true'` guard so it only fires on a real poisoned entry. Two notes if you keep the step rather than adopting the content-addressed key above:

- No `continue-on-error`, so a transient failure here (entry evicted between restore and delete, cache service blip) fails the main run before `Seed` gets a chance — which is the one run that can repair the cache.
- It depends on `GITHUB_REF` being `refs/heads/main`, which holds for `push` but also lets a `workflow_dispatch` on main delete the entry. Harmless, just noting it is broader than "push to main".

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -116,49 +123,137 @@ jobs:
...
+        shell: pwsh
+        timeout-minutes: 10
+        run: |
+          $cacheDirectory = Join-Path $env:RUNNER_TOOL_CACHE "Ubuntu\$env:WSL_UBUNTU_VERSION\x64"
+          $installerPath = Join-Path $cacheDirectory $env:WSL_UBUNTU_INSTALLER_FILE
+          $completeMarkerPath = "$cacheDirectory.complete"
+          New-Item -ItemType Directory -Path $cacheDirectory -Force | Out-Null
+          Remove-Item -LiteralPath $installerPath,$completeMarkerPath -Force -ErrorAction SilentlyContinue
+
+          Invoke-WebRequest -Uri $env:WSL_UBUNTU_INSTALLER_URL -OutFile $installerPath
```
JeffreyDallas: This is now the only acquisition path in the workflow, so a transient blip on `releases.ubuntu.com` fails the run that exists to repair the cache. One retry is cheap insurance, and silencing progress keeps 372 MB worth of progress records out of the step log:

```suggestion
          $ProgressPreference = 'SilentlyContinue'
          Invoke-WebRequest -Uri $env:WSL_UBUNTU_INSTALLER_URL -OutFile $installerPath -MaximumRetryCount 2 -RetryIntervalSec 10
```

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -116,49 +123,137 @@ jobs:
...
+      - name: Save seeded Ubuntu WSL installer cache
+        if: ${{ success() && steps.seed_ubuntu_wsl_installer.outputs.valid == 'true' }}
+        uses: actions/cache/save@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: ${{ runner.tool_cache }}\Ubuntu\${{ env.WSL_UBUNTU_VERSION }}\x64
+          key: ${{ env.WSL_UBUNTU_CACHE_KEY }}
+
+      - name: Report Ubuntu WSL installer cache availability
+        id: ubuntu_wsl_installer
+        if: ${{ success() }}
```
JeffreyDallas: `success()` is already implied for any step whose `if` contains no status-check function, so this condition is a no-op. (The same is true of the leading `success() &&` on the other new steps — harmless, just noise.) If Major 1 lands, this step goes away entirely.

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -404,22 +447,23 @@ jobs:
...
+        if: ${{ success() && steps.ubuntu_wsl_installer.outputs.available == 'true' }}
         shell: pwsh
         timeout-minutes: ${{ fromJSON(env.STEP_TIMEOUT_MINUTES) }}
         run: |
           $env:PATH = "$env:USERPROFILE\.solo\bin;$env:PATH"
           npm run solo -- deployment diagnostics connections -d one-shot --check
 
       - name: Collect Diagnostics on Failure
-        if: ${{ failure() }}
+        if: ${{ failure() && steps.ubuntu_wsl_installer.outputs.available == 'true' }}
```
JeffreyDallas: Gating this on `available` is right. Separately, the `; $true` at `:463` (and `:471`) does not do what it looks like: under `shell: pwsh` the runner appends an exit on `$LASTEXITCODE`, so a failing `npm run solo` still fails the step. Attempt 4 of run `33823850099` shows it — the script printed `True` and the step still ended `##[error]Process completed with exit code 1`, for both this step and `One-Shot Single Destroy`.

A diagnostics step should not be able to turn a run red on its own. `; $global:LASTEXITCODE = 0` (or a trailing `exit 0`) is what the `; $true` was reaching for. Pre-existing, so entirely fine to leave for a follow-up — flagging it because this PR is touching these two steps anyway.

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -116,49 +120,110 @@ jobs:
...
+            Remove-Item -LiteralPath $installerPath -Force -ErrorAction SilentlyContinue
+            throw "Seeded Ubuntu WSL installer failed SHA-256 validation: $actualHash"
+          }
+
+          New-Item -ItemType File -Path $completeMarkerPath -Force | Out-Null
+          Add-Content -Path $env:GITHUB_OUTPUT -Value 'valid=true'
+          Write-Host "Seeded verified Ubuntu WSL installer cache: $installerPath"
+
+      - name: Save seeded Ubuntu WSL installer cache
+        if: ${{ success() && steps.seed_ubuntu_wsl_installer.outputs.valid == 'true' }}
```
JeffreyDallas: **Major — this needs the ref check that `Seed` just lost.** With seeding on every branch, `Save` runs on every branch too, and a `refs/pull/N/merge` cache entry is visible only to that one PR. It already happened here:

```
ubuntu-wsl-installer-v1-24.4.4-9b2f7730...
   ref=refs/pull/5962/merge  size=389575574  created=2026-09-04T12:38:30Z
```

372 MB that no other PR and not main can read, in a repo whose cache usage is now `11467414502` bytes across 84 entries — 15% over the 10 GB limit, and down from 206 entries yesterday, so it is evicting aggressively. Each private copy pushes out caches other workflows actually reuse.

The cost of not saving on a PR is 6 seconds (the measured seed time in this run), so this is close to free:

```suggestion
        if: ${{ github.ref == 'refs/heads/main' && steps.seed_ubuntu_wsl_installer.outputs.valid == 'true' }}
```

(`success()` is already implied for an `if` with no status-check function, so dropping it changes nothing.)

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -116,49 +120,110 @@ jobs:
...
+        shell: pwsh
+        timeout-minutes: 10
+        run: |
+          $cacheDirectory = Join-Path $env:RUNNER_TOOL_CACHE "Ubuntu\$env:WSL_UBUNTU_VERSION\x64"
+          $installerPath = Join-Path $cacheDirectory $env:WSL_UBUNTU_INSTALLER_FILE
+          $completeMarkerPath = "$cacheDirectory.complete"
+          New-Item -ItemType Directory -Path $cacheDirectory -Force | Out-Null
+          Remove-Item -LiteralPath $installerPath,$completeMarkerPath -Force -ErrorAction SilentlyContinue
+
+          Invoke-WebRequest -Uri $env:WSL_UBUNTU_INSTALLER_URL -OutFile $installerPath
```
JeffreyDallas: Carried over from the last round — still worth doing, and more so now that this is the *only* acquisition path in the workflow. A transient blip on `releases.ubuntu.com` fails the run outright, with no cache fallback behind it. `-MaximumRetryCount` / `-RetryIntervalSec` are native to `Invoke-WebRequest` in PowerShell 7, and silencing progress keeps 372 MB of progress records out of the step log:

```suggestion
          $ProgressPreference = 'SilentlyContinue'
          Invoke-WebRequest -Uri $env:WSL_UBUNTU_INSTALLER_URL -OutFile $installerPath -MaximumRetryCount 2 -RetryIntervalSec 10
```

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -116,49 +120,110 @@ jobs:
...
       - name: Setup WSL2
         id: setup_wsl2
         # Kind requires Linux containers. GitHub's windows-latest runner has Docker in Windows
         # container mode with no Docker Desktop to switch it. WSL2 + Docker inside it replicates
         # what Docker Desktop does for Windows users, exposing a Linux daemon via TCP.
-        # use-cache is enabled so the distribution installer is cached across runs via
-        # actions/cache instead of re-downloaded every time — this workflow was intermittently
-        # timing out on the fresh download from a shared runner's network.
+        # The verified restore above creates setup-wsl's .complete marker so find()
+        # returns the installer before the action reaches its cache or download path.
```
JeffreyDallas: This claim is now **verified end to end**, which it was not when I flagged it last round. Run [33830097179](https://github.com/hiero-ledger/solo/actions/runs/33830097179) has exactly one 372 MB transfer (the workflow's restore at `03:34:08`), and the `Setup WSL2` group goes from `Install Distribution` at `03:34:15.8` straight to `wsl --install --from-file C:\hostedtoolcache\windows\Ubuntu\24.4.4\x64\ubuntu-24.04.4-wsl-amd64.wsl` at `03:34:16.1` with no cache or download activity in between. `find()` hits and the action returns at `SetupWsl.kt:182`, exactly as the comment says.

One addition worth making while you are here: the marker path binds `WSL_UBUNTU_VERSION` to `Distribution.kt`'s `SemVer` for this distribution, and `.github/dependabot.yml:3-13` will open a `setup-wsl` bump automatically. If a bump moves the Ubuntu version, `find()` misses and — with `use-cache: false` — the action silently falls back to a full download on every run rather than failing, so nothing surfaces the drift.

```suggestion
        # The verified restore above creates setup-wsl's .complete marker so find()
        # returns the installer before the action reaches its cache or download path.
        # WSL_UBUNTU_VERSION must match Distribution.kt's SemVer for this distribution:
        # a setup-wsl bump that moves the Ubuntu version makes find() miss, and with
        # use-cache: false the action silently falls back to a full download per run.
```

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -116,49 +120,110 @@ jobs:
...
         run: |
           task build:compile
 
-      - name: Validate Ubuntu WSL cache
+      - name: Restore verified Ubuntu WSL installer cache
+        id: restore_ubuntu_wsl_installer
+        uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: ${{ runner.tool_cache }}\Ubuntu\${{ env.WSL_UBUNTU_VERSION }}\x64
+          key: ubuntu-wsl-installer-v1-${{ env.WSL_UBUNTU_VERSION }}-${{ env.WSL_UBUNTU_INSTALLER_SHA256 }}
```
JeffreyDallas: Content-addressing the key is exactly the right call, and the `v1` prefix is a better idea than what I suggested — it gives you a deliberate invalidation lever independent of the version and hash.

One consequence worth being ready for: this key has never been written on `main`, so the first main run after merge seeds and saves (~6 s + 8 s), and every PR run before that misses and seeds. That is the design working as intended, but it is also why `Save` needs the main-only guard at `:207` — otherwise that window fills the cache budget with per-PR copies.

.github/workflows/flow-standard-runner-one-shot-windows.yaml
JeffreyDallas: Non-blocking, carried from the last two rounds — this is the only acquisition path in the workflow, so a transient blip on `releases.ubuntu.com` fails the run with nothing behind it. `-MaximumRetryCount` / `-RetryIntervalSec` are native to `Invoke-WebRequest` in PowerShell 7:

```suggestion
          $ProgressPreference = 'SilentlyContinue'
          Invoke-WebRequest -Uri $env:WSL_UBUNTU_INSTALLER_URL -OutFile $installerPath -MaximumRetryCount 2 -RetryIntervalSec 10
```

Happy for this to be a follow-up if you would rather not touch a green workflow again.

.github/workflows/flow-standard-runner-one-shot-windows.yaml
```diff
@@ -116,49 +120,110 @@ jobs:
...
       - name: Setup WSL2
         id: setup_wsl2
         # Kind requires Linux containers. GitHub's windows-latest runner has Docker in Windows
         # container mode with no Docker Desktop to switch it. WSL2 + Docker inside it replicates
         # what Docker Desktop does for Windows users, exposing a Linux daemon via TCP.
-        # use-cache is enabled so the distribution installer is cached across runs via
-        # actions/cache instead of re-downloaded every time — this workflow was intermittently
-        # timing out on the fresh download from a shared runner's network.
+        # The verified restore above creates setup-wsl's .complete marker so find()
+        # returns the installer before the action reaches its cache or download path.
```
JeffreyDallas: Optional. The marker path binds `WSL_UBUNTU_VERSION` to `Distribution.kt`'s `SemVer` for this distribution, and `.github/dependabot.yml:3-13` will open a `setup-wsl` bump automatically. If a bump moves the Ubuntu version, `find()` misses and — with `use-cache: false` — the action quietly falls back to a full download on every run instead of failing, so nothing surfaces the drift.

```suggestion
        # The verified restore above creates setup-wsl's .complete marker so find()
        # returns the installer before the action reaches its cache or download path.
        # WSL_UBUNTU_VERSION must match Distribution.kt's SemVer for this distribution:
        # a setup-wsl bump that moves the Ubuntu version makes find() miss, and with
        # use-cache: false the action silently falls back to a full download per run.
```

Fantomasa (approved):
LGTM

jan-milenkov (approved):

PR #5969, opened 2026-09-05
fix: load component images only into target Kind contexts

### LINKED ISSUE #5964: Load component images only into the component's target Kind clusters
## Overview
`BaseCommand.kindLoadComponentImage` (introduced in #5949) preloads a local component image into every `kind-*` context returned by `remoteConfig.getContexts()`, because remote config only exposes consensus node contexts, not the contexts a component actually deploys to. This has two consequences: a stale `kind-*` consensus context fails the whole component deploy even though it is not required for correctness, and the extra-context preloading is not truly best-effort since the load loop has no per-context error handling.

## Sub-Tasks
- [ ] Thread each command's actual deployment target contexts through `kindLoadComponentImage` so only those are loaded
- [ ] Warn and continue when loading into a non-primary context fails, keeping the caller-supplied cluster context as the only required one
- [ ] Add unit tests for stale or unreachable extra Kind contexts

## References
- [PR #5949 review thread](https://github.com/hiero-ledger/solo/pull/5949#discussion_r3922495544) - original discussion

AdityaOP007:
#5964 

## Summary

`kindLoadComponentImage` previously discovered Kind contexts globally and attempted to load component images into all of them. This could cause a component deployment to fail because of an unrelated or stale Kind cluster.

This change makes image loading target-aware by using the component's primary deployment context directly, while allowing explicitly provided additional target contexts to be handled on a best-effort basis.

## Behavior

* **Primary context:** Image loading remains required. A failure is propagated and fails the deployment.
* **Additional target contexts:** Image loading is best-effort. Failures emit a warning and do not stop deployment or prevent subsequent contexts from being processed.
* **Unrelated contexts:** No longer discovered or loaded automatically.

## Tests

Added regression coverage for:

* Primary-only image loading
* Multi-context image loading
* Target-context isolation
* Primary-context failure propagation
* Best-effort additional-context failures
* Continuing after an additional-context failure
* Stale/unreachable additional Kind contexts
* Duplicate contexts
* Non-Kind contexts

## Validation

* `npx mocha test/unit/commands/base.test.ts` — PASS
* `npx eslint src/commands/base.ts test/unit/commands/base.test.ts` — PASS
* `npx prettier --check src/commands/base.ts test/unit/commands/base.test.ts` — PASS
* `npx tsc --noEmit` — PASS
* `git diff --check` — PASS
* DCO sign-off verified

Two unrelated Windows environment failures were observed in existing tests (`runShell` and `dockerDesktopPreflightTask); they are unrelated to this change and the #5964 regression suite passes successfully.

Closes #5964

### DIFF
--- a/src/commands/base.ts
+++ b/src/commands/base.ts
@@ -292,15 +292,50 @@ export abstract class BaseCommand extends ShellRunner {
     return checkDockerImageExists(name, tag);
   }
 
-  protected async kindLoadComponentImage(componentImage: string, clusterContext: string): Promise<void> {
-    const kindClusterName: string = this.kindClusterNameFromContext(clusterContext);
-    this.logger.debug(`Loading '${componentImage}' into Kind cluster '${kindClusterName}'`);
+  protected async kindLoadComponentImage(
+    componentImage: string,
+    clusterContext: string,
+    additionalTargetContexts: Context[] = [],
+  ): Promise<void> {
     const kindExecutable: string = await this.depManager.getExecutable(constants.KIND);
     const kindClient: KindClient = await this.kindBuilder.executable(kindExecutable).build();
+
+    // The primary context is required for deployment; failure propagates to the caller.
+    const primaryKindClusterName: string = this.kindClusterNameFromContext(clusterContext);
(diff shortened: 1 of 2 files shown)


JeffreyDallas:
The problem already closed

PR #5977, opened 2026-09-09
fix(ci): clear remaining adhoc-packages zizmor findings

JeffreyDallas:
## Description

Tier 6 of the stacked zizmor remediation series. **Stacked on #5958** — base is `zizmor-cleanup-tier5`, so this diff shows only the tier 6 change. Merge #5958 first.

zizmor is run **offline** (v1.30.0, `--no-config`); this series adds no scanning job and no ignore/baseline configuration.

**This PR clears the last mechanically-actionable rule: `adhoc-packages` 13 → 0.** After it, the only findings left in the repository are `template-injection` (104, all Informational/Low, from `${{ inputs.* }}` / `${{ env.* }}` on trusted sources) — the backlog tier 5 already characterized as "cleanup work with no security urgency, and a reasonable place to stop or continue at leisure."

### What the rule flags, and what it actually wants

Per zizmor's own docs, `adhoc-packages` flags `run:` steps that install packages outside a managed, locked manifest — `npm install -g`, `pip install`, etc. — for two reasons: the install can resolve to a floating version, and **even when the top-level package is pinned, its transitive dependencies are still resolved fresh on every run.** The intended fix is a manifest + lockfile, installed with a lockfile-aware command (`npm ci`), not just adding a version pin to the ad-hoc command.

That fix is applied where it was safe to verify locally. The rest are either the same audit-flags-it-regardless-of-pinning case as a judgment call, or don't actually fit the rule.

### Fixed: the Snyk toolchain (1 finding)

`zxc-code-analysis.yaml`'s `Setup Snyk` step installed `snyk`, `snyk-to-html` and `@wcj/html-to-markdown-cli` with no version pins at all — the one case here with no "this needs to float" justification and no unusual blast radius (it only runs when `inputs.enable-snyk-scan` is set, and never gates a release). It now gets a dedicated manifest at `.github/workflows/support/snyk-tools/package.json`, installed with `npm ci --prefix` and added to `GITHUB_PATH` so the existing bare `snyk` / `snyk-to-html` / `html-to-markdown` invocations later in the job keep working unchanged.

Verified end to end, not just parsed: ran `npm ci` against the generated `package-lock.json` in a scratch directory and confirmed all three binaries install and run (`snyk --version` → `1.1307.1`, `html-to-markdown` → `v2.1.1`; `snyk-to-html` has no `--version` flag but is present and invokable).

### Removed as dead code (1 finding)

`flow-gcs-test.yaml`'s `Install Dependencies` step ran `SOLO_NO_CACHE=true npm install -g @hashgraph/solo` — unpinned, and the legacy pre-org-rename package name. Traced every use of a `solo` binary in this job's path: `PATH=~/.solo/bin:${PATH}` is exported, but `gcs_test.sh` (the script this job runs) only ever calls `npm run solo-test`, which runs the **locally compiled** source via `tsx`, never a bare `solo` command. Nothing in this job reads the globally installed CLI. The line is removed rather than pinned.

### Deferred, with the reasoning recorded inline (11 findings)

Each of these is either the "audit flags it regardless of pinning" case from a toolchain too sensitive to migrate on a hunch, or a case the rule doesn't actually fit. Each `# zizmor: ignore[adhoc-packages]` comment carries its own reasoning in place, per the convention tier 5 established:

* **`flow-deploy-release-artifact.yaml`, two "Install Semantic Release" blocks (6 findings)** — `semantic-release` and every plugin/helper are exact-pinned; only their transitive dependencies lack a lockfile. This toolchain runs directly ahead of the steps that consume its output — git tag, GitHub release, npm publish. A lockfile migration is the correct fix in principle, but this is the release pipeline, and the only way to fully verify that migration is against a real release, not a dry run. Deferred rather than risked.
* **`flow-deploy-release-artifact.yaml`, two `npm install -g npm@11.7.0` steps (2 findings)** — upgrading npm itself, pinned, before any lockfile-aware command can run. A textbook chicken-and-egg case: there's no `npm ci` to reach for when the thing being installed is the tool `npm ci` needs.
* **`flow-ledger-reset-warm-start-timing.yaml` and `flow-one-shot-timing.yaml`, `npm i @hiero-ledger/solo@latest` (2 findings)** — both gated on `inputs.use-released-solo == true`, both named "Install Solo CLI (released)". `@latest` is the point: the step exists to time the currently-published release against a from-source build in the same job. Pinning would defeat the step and require a workflow edit on every release.
* **`flow-standard-runner-one-shot.yaml`, `npm install -g ./output/hl/hiero-ledger-solo-*.tgz` (1 finding)** — installs the tarball the immediately-preceding `pack` step built from this checkout. Not a registry fetch; there is no version to pin.

### Remaining after this PR

104 findings, all `template-injection`, all Informational (97) or Low (7) — `${{ inputs.* }}` / `${{ env.* }}` expansions from trusted sources, already routed through `env:` by tier 4. Per #5931, the endpoint is for #5930 to be updated once the backlog is at an acceptable level, so the scan starts green and only flags genuinely new findings. After this PR, that level is effectively reached — what's left is Informational/Low noise, not a queue of unaddressed risk.

### Related Issues

* Related to #5931
* Stacked on #5958

Two boxes are left unchecked. **No tests were added:** these are CI workflow definitions, which the repository's unit and E2E suites do not exercise; verification was static plus one local end-to-end install check, both described below. **No documentation update:** the change adds a CI-only dependency manifest and inline suppression comments, with no user-facing surface.

The new `.github/workflows/support/snyk-tools/package.json` (and its generated `package-lock.json`) is a `package.json` in the repository, so per the checklist item above: it is a CI-only manifest, not part of the published `@hiero-ledger/solo` package graph — pinning `snyk@1.1307.1`, `snyk-to-html@3.7.9` and `@wcj/html-to-markdown-cli@2.1.1`, the current published versions at the time of this PR.

### Testing

`task build` and `task format` were both run before pushing: **exit 0**, `0 errors` from ESLint. The pre-existing `no-explicit-any` warnings in `test/` are untouched by this change.

With zizmor 1.30.0:

* Scanned before and after with `zizmor --no-config .github/workflows .github/actions` and diffed by rule: `adhoc-packages` **13 → 0**, total **117 → 104**. `template-injection` (104, unchanged) confirms nothing else was touched.
* Verified the Snyk manifest is real, not just plausible-looking: ran `npm ci` against `.github/workflows/support/snyk-tools/package-lock.json` in a scratch directory and confirmed `snyk`, `snyk-to-html` and `html-to-markdown` all install into `node_modules/.bin` and run (`snyk --version` → `1.1307.1`; `html-to-markdown` prints `v2.1.1`).
* Traced every `solo`-binary code path in `flow-gcs-test.yaml`'s job before removing the dead install: `gcs_test.sh` calls only `npm run solo-test` (the locally compiled CLI), never a bare `solo` command, so the removed global install had no consumer.
* Parsed all workflow and composite-action files with `js-yaml` after editing — all valid, including the `run: | # zizmor: ignore[...]` lines (the ignore comment sits outside the block-literal content, per zizmor's own documented placement rule, so it is a real YAML comment and not shell text).

The following was not tested:

* No workflow was executed. `zxc-code-analysis.yaml`'s Snyk step only runs with `inputs.enable-snyk-scan` set, so it won't get coverage from this PR's own PR-triggered checks; the local `npm ci` verification above is what stands in for it.
* `flow-deploy-release-artifact.yaml`, `flow-ledger-reset-warm-start-timing.yaml`, `flow-one-shot-timing.yaml` and `flow-standard-runner-one-shot.yaml` are release, nightly-timing and matrix paths this PR does not trigger on its own; none of their content changed here beyond a comment and a suppression annotation, so the risk is limited to that annotation being misplaced, which the `js-yaml` parse above rules out.


.github/workflows/zxc-code-analysis.yaml
```diff
@@ -148,7 +148,9 @@ jobs:
         env:
           SNYK_TOKEN: ${{ secrets.snyk-token }}
         if: ${{ inputs.enable-snyk-scan && !cancelled() && !failure() }}
-        run: npm install -g snyk snyk-to-html @wcj/html-to-markdown-cli
+        run: |
+          npm ci --prefix .github/workflows/support/snyk-tools
```
ddeskov-limechain, line 152: Small thing: `snyk test --all-projects` in the next step may now also pick up this new manifest. If you do not want the Snyk tools themselves scanned, `--exclude=snyk-tools` on the scan command keeps it out.

ddeskov-limechain (dismissed):
Looks good. Checked the new lockfile with `npm audit` (0 vulnerabilities) and confirmed nothing in the GCS job calls a bare `solo` binary, so the removed global install was dead. One small nit inline. Also, #5958 is merged now, so the stacking note in the description is stale.

.github/workflows/zxc-code-analysis.yaml
JeffreyDallas: Good catch — added `--exclude=snyk-tools` to the `snyk test` command in 6b284af1e.

jan-milenkov (approved):

PR #5980, opened 2026-09-10
fix: parse semantic version prerelease metadata

### LINKED ISSUE #5932: SemanticVersion misreads build metadata containing a hyphen as a pre-release
### Describe the bug

`SemanticVersion` extracts the pre-release by matching `/-(.+?)(?:\+|$)/` against the whole version string ([`src/business/utils/semantic-version.ts` line 149](https://github.com/hiero-ledger/solo/blob/a139e56c9642e1ecad715bec5cca7cb90d415ede/src/business/utils/semantic-version.ts#L149)). The match is not limited to the part of the string before the `+`, so on a version that has build metadata containing a hyphen but no pre-release, the tail of the build metadata is picked up as a phantom pre-release:

```
new SemanticVersion('1.2.3+build-1')
// actual:   preRelease = '1',  buildMetadata = 'build-1',  toString() => '1.2.3-1+build-1'
new SemanticVersion('1.2.3+sha-a1b2c3')
// actual:   preRelease = 'a1b2c3',  buildMetadata = 'sha-a1b2c3',  toString() => '1.2.3-a1b2c3+sha-a1b2c3'
```

Because the phantom pre-release participates in precedence, `new SemanticVersion('1.2.3+build-1').compare(new SemanticVersion('1.2.3'))` returns `-1` and `lessThan` returns `true`.

Per the [SemVer 2.0.0 specification](https://semver.org/#spec-item-10), build metadata is everything after the `+`, may itself contain hyphens, and MUST be ignored when determining precedence. So:

* `1.2.3+build-1` should parse to `preRelease = null`, `buildMetadata = 'build-1'` and round-trip through `toString()` unchanged.
* `1.2.3+build-1` should compare equal to `1.2.3` (`compare` → `0`, `lessThan` → `false`).

Versions that have both parts, e.g. `1.2.3-alpha+build-1`, already parse correctly; the bug only affects versions with a hyphen in the build metadata and no pre-release. A likely fix is to match the pre-release only against the substring before the first `+`.

### To Reproduce

```bash
npx tsx -e "
import {SemanticVersion} from './src/business/utils/semantic-version.js';
const v = new SemanticVersion('1.2.3+build-1');
console.log(v.preRelease, v.buildMetadata, v.toString());  // 1 build-1 1.2.3-1+build-1
console.log(v.lessThan(new SemanticVersion('1.2.3')));     // true
"
```

### Additional Context

Found while writing the property-based round-trip tests in #5912: the build-metadata generator in `test/unit/business/utils/semantic-version.property.test.ts` had to exclude hyphens so that the round-trip property holds, and this issue is the tracked record of that workaround. Part of the technical-debt checklist of #5912.

mayanksinha17:
## Description

Fixes parsing of semantic version pre-release metadata.

### Changes
- Correctly parse pre-release identifiers containing hyphens.
- Preserve the complete pre-release identifier.
- Handle build metadata correctly when combined with pre-release versions.
- Added/updated tests covering the affected semantic version cases.

### Testing
- Targeted Vitest tests were verified.
- Existing semantic-version tests pass.

## Related Issues

Closes #5932


src/business/utils/semantic-version.ts
```diff
@@ -146,7 +146,7 @@ export class SemanticVersion<T extends string | number> {
         }
       }
 
-      const preReleaseMatch: RegExpMatchArray = this.originalValue.match(/-(.+?)(?:\+|$)/);
+      const preReleaseMatch: RegExpMatchArray = this.originalValue.match(/^v?\d+(?:\.\d+)*-([^+]+)/);
```
JeffreyDallas, line 153: Nit: worth a short comment here on *why* the regex is anchored to the start of the string (`^v?\d+(?:\.\d+)*-`) rather than searching for the first hyphen anywhere. That anchor is the entire fix for issue #5932 — without a note, a future refactor could loosen it back to something unanchored and silently reintroduce the phantom-pre-release bug this PR closes.

```suggestion
      // Anchored to the start so a hyphen inside build metadata (after '+') is never
      // mistaken for the pre-release separator (see issue #5932).
      const preReleaseMatch: RegExpMatchArray = this.originalValue.match(/^v?\d+(?:\.\d+)*-([^+]+)/);
```

JeffreyDallas (dismissed):

jan-milenkov (dismissed):

mayanksinha17:
Hi! I checked the failing PR Automation workflow. The failure appears to be in the Automate PR Labels and Status step, where actions/github-script reports Input required and not supplied: github-token. This seems to be a workflow configuration/permission issue rather than an issue with the PR changes. Could a maintainer please take a look?

JeffreyDallas (approved):

mayanksinha17:
Hi, I noticed that the PR Automation check is currently failing at the Automate PR Labels and Status step with Input required and not supplied: github-token. This appears in the recent runs as well. The PR itself only changes semantic-version.ts and its unit tests, and the other checks are passing. Just flagging this in case the workflow needs attention. Please let me know if anything is needed from my side.

mayanksinha17:
Hi @JeffreyDallas, just following up on this PR. The changes are approved, and the remaining PR Automation check still appears to be failing due to the github-token input issue mentioned above.

Please let me know if any changes are needed from my side. Thanks!

mayanksinha17:
Hi @JeffreyDallas / @jan-milenkov, the branch has now been restored to the original semantic-version changes (c6fe08d and 0ff674a). The accidental Windows deployment commit was removed and moved to separate PR #6041. Could you please re-review/approve #5980 when convenient? Thank you.

mayanksinha17:
Hi @JeffreyDallas / @jan-milenkov, I’ve updated #5980 with the latest main. The workflows are currently awaiting maintainer approval, and the required checks are pending. Could you please approve the workflows and re-review the PR when convenient? Thank you!

PR #5999, opened 2026-09-15
docs: document and test one-shot falcon deploy without a values file

### LINKED ISSUE #3296: Document and test one-shot falcon deploy with no values file
Document and test that if one-shot falcon deploy runs with no values file, that it just uses the defaults of one-shot single and works

Fantomasa:
## Description

Fixes #3296

Documents and tests that `solo one-shot falcon deploy` runs without `--values-file`, in which case it uses the same defaults as `solo one-shot single deploy`. Per the discussion on the issue, the values file stays optional and no default file is introduced.

### Changes

- **CLI help:** the `one-shot falcon deploy` description now states that the values file is optional and that the one-shot single defaults apply without one.
- **Example docs:** `examples/one-shot-falcon/README.md` gains a "Deploying Without a Values File" section listing the defaults that apply (deployment/namespace/cluster-ref `one-shot`, one small-memory consensus node, mirror node + explorer + relay, block node only with `ONE_SHOT_WITH_BLOCK_NODE=true`, release component versions unless overridden by flags).
- **Example task:** `examples/one-shot-falcon/Taskfile.yml` gains a `deploy-defaults` task that runs `solo one-shot falcon deploy` with no values file, so the defaults path can be exercised with `task deploy-defaults`.
- **Testability refactor:** the values-file loading in the deploy orchestrator's `Initialize` phase is extracted into a private `applyValuesFileOverrides(config)` method. Behavior is unchanged.
- **Unit tests:**
  - `test/unit/commands/one-shot-falcon-deploy.test.ts`: `FALCON_DEPLOY_FLAGS_LIST` has no required flags, `--values-file` is optional with an empty default, and every deploy flag shared with `one-shot single deploy` is accepted.
  - `default-one-shot-deploy-orchestrator.test.ts`: without a values file all per-component configurations stay at the one-shot single defaults and no file is read; a missing file throws `ValuesFileNotFoundSoloError`; a values file only overrides the sections it contains.

### Testing

- `npx mocha` on the two unit test files: 53 passing.
- `npx tsc --noEmit` and `npx eslint .`: no errors.
- Manual: `solo one-shot falcon deploy --help` shows the updated description with `--values-file` listed as optional.
- Manual: in a throwaway kind cluster with a scratch `SOLO_HOME`, `solo one-shot falcon deploy -q --deploy-mirror-node=false --deploy-explorer=false --deploy-relay=false` with **no** `--values-file` completed successfully using the one-shot single defaults: deployment/namespace/cluster-ref `one-shot`, consensus node v0.75.1 running, small-memory profile applied (896Mi root-container limit), predefined accounts created. `solo one-shot falcon destroy -q` then cleaned everything up.
- Manual: from `examples/one-shot-falcon`, `task --list` shows `deploy-defaults` and `task deploy-defaults --dry` resolves to `solo one-shot falcon deploy` with no `--values-file`. The task was not executed end to end here; the equivalent command was run directly in the isolated cluster above. To run it for real: `cd examples/one-shot-falcon && task deploy-defaults`, then `task destroy`.
- Note: the existing `test/e2e/commands/small-memory-load.test.ts` already exercises `one-shot falcon deploy` without a values file in CI.

### Related Issues

- #3296


test/unit/commands/one-shot-falcon-deploy.test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import {expect} from 'chai';
+import {describe, it} from 'mocha';
+import {DefaultOneShotCommand} from '../../../src/commands/one-shot/default-one-shot.js';
+import {Flags} from '../../../src/commands/flags.js';
+import {type CommandFlag} from '../../../src/types/flag-types.js';
+
+// Covers https://github.com/hiero-ledger/solo/issues/3296: `one-shot falcon deploy` must run without
+// a values file, in which case it uses the same defaults as `one-shot single deploy`.
+describe('DefaultOneShotCommand.FALCON_DEPLOY_FLAGS_LIST', (): void => {
```
jan-milenkov, line 9: some of the tests are a bit obsolete

examples/one-shot-falcon/README.md
```diff
@@ -63,6 +63,31 @@ Browse the source code and configuration files for this example in the [GitHub r
      * Destroy the Solo network using `solo one-shot falcon destroy`
      * Delete the Kind cluster
 
+## Deploying Without a Values File
+
+The `--values-file` flag is optional. When it is omitted, `solo one-shot falcon deploy` behaves exactly like `solo one-shot single deploy` and uses its defaults:
+
+* Deployment, namespace, and cluster reference named `one-shot`
```
jan-milenkov: explaining the default one-shot single behavior creates a unwanted relationship, i would suggest omitting it because `behaves exactly like solo one-shot single deploy` is enough of an explanation

